PR #1, opened 2020-04-08
Update UMIPs repo

rcai1:


umip-template.md
```diff
@@ -0,0 +1,35 @@
+This is the suggested template for new UMIPs.
+
+Note that an UMIP number will be assigned by an editor. 
+When opening a pull request to submit your UMIP, please use an abbreviated title in the filename, umip-draft_title_abbrev.md.
+
+The title should be 44 characters or less.
```
rcai1, line 6: @mrice32 @hlambur, I copied this over from the EIP template file (https://github.com/ethereum/EIPs/blob/master/eip-template.md) and want to confirm if we'll use this process as well. I can also delete these lines from the umip-template file.

PR #2, opened 2020-04-14
Draft UMIP-2

allisonlu22:
UMIP to add ETHBTC as a price identifier


UMIPs/umip-2.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Motivation
+The DVM currently does not support the ETHBTC index.
+
+Cost: Pricing for this index is easy to access through open centralized exchange APIs, and cross-exchange price discrepancies are typically <0.5%. 
+
+Opportunity: A synthetic token that tracks ETHBTC could be used as a hedging tool. It can also be used to create synthetic bitcoin or synthetic ETH.
+
+## Technical Specification
+The definition of this identifier should be the median of the ETHBTC trading pair price from 3 exchanges’ public APIs: Coinbase, Binance, and Bitstamp, unless that median value differs from “Broad Market Consensus” as defined by the DVM’s decentralized human judgment. The calculation of median should not include observation of ETHUSD or BTCUSD directly, and instead only look at the ETHBTC trading pair on each exchange. The precision should be to [n] decimals with timestamp granularity to 1 second.
```
allisonlu22, line 19: ENG: please fill in the precision that is available across the three exchanges noted

UMIPs/umip-2.md
```diff
@@ -0,0 +1,34 @@
...
+
+More complex computations using centralized exchange prices (like incorporating additional exchanges, calculating a TWAP or VWAP, or imposing price bands, etc.) have the potential to add a greater level of precision and robustness to the definition of this identifier. However, we felt that the costs in increased complexity and mental load to the stakeholders who need to use it (like sponsors, liquidators, and disputers) outweighed the benefits. 
+
+## Implementation
+[ENG to fill in]
```
allisonlu22, line 31: ENG - what's a good way to describe the implementation for an identiier? Links to the CEX APIs? Pointer to an internal library?

UMIPs/umip-2.md
mrice32: just FYI, you can tag our entire dev team using the following: @UMAprotocol/dev

mrice32: @allisonlu22 I'm not aware of any APIs where you can get historical data at a 1 second granularity. The best I'm aware of is https://cryptowat.ch/, which provides OHLC at a 60 second granularity.

Using that same API, it looks like the min granularity of prices is coinbase-pro, at 5 decimals.

UMIPs/umip-2.md
mrice32: I think the implementation is referring to how this UMIP is *implemented* in our protocol. This would just be the transaction that we would need to send to approve the identifier, right? Or were you hoping to add something more here?

UMIPs/umip-2.md
allisonlu22: oh god I didn't know that granularity was only down to 60 seconds. As a liquidator, that would get quite confusing if I'm observing the exchanges directly. Can you reword the PR here to reflect what you think is the right way to define this price?

UMIPs/umip-2.md
allisonlu22: I was thinking that it would provide both the transaction to approve the identifier, and also a technical implementation of the price feed itself so that there is no ambiguity for future liquidators, disputers, and voters.

UMIPs/umip-2.md
mrice32: Sure. Will do.

PR #6, opened 2020-04-28
Move UMIP-3 to Final

mrice32:


hlambur:
All good on content. For future reference (@rcai1), I think we could make this a lot prettier and more readable. The header can be made into a table (see: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables), and we should make each PR reference link to the actual PR.

rcai1:
> All good on content. For future reference (@rcai1), I think we could make this a lot prettier and more readable. The header can be made into a table (see: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables), and we should make each PR reference link to the actual PR.

While I think the table seems like a reasonable addition, I believe we'd need a programmatic way to get the PR links to be associated with the commits we list. It also seems like something we may not continue to do in future UMIPs.

PR #7, opened 2020-04-30
Create umip-4.md

rcai1:


UMIPs/umip-newcontract.md
```diff
@@ -0,0 +1,140 @@
+# Headers
```
mrice32, line 1: nit: shouldn't this be called `umip-4.md`?

UMIPs/umip-newcontract.md
```diff
@@ -0,0 +1,140 @@
...
+| UMIP #     |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Register new financial contract with DVM                                                                                                 |
```
mrice32, line 4: nit: rather than calling this a "new financial contract," I think we should be specific about which financial contract we're referring to -- the `ExpiringMultiParty` contract template.

UMIPs/umip-newcontract.md
```diff
@@ -0,0 +1,140 @@
...
+| Authors    | Matt Rice (matt@umaproject.org), Prasad Tare (prasad@umaproject.org), Chris Maree (chris@umaproject.org), Nick Pai (nick@umaproject.org) |
+| Status     | Draft                                                                                                                                    |
+| Created    | April 30, 2020                                                                                                                           |
+
+
+# Summary
+
+This UMIP registers a new financial contract template with the DVM.
```
mrice32, line 12:
```suggestion
This UMIP registers the `ExpiringMultiParty` template with the DVM. 
```

UMIPs/umip-newcontract.md
```diff
@@ -0,0 +1,140 @@
...
+
+Additional details can be found [here](https://docs.umaproject.org/uma/synthetic_tokens/explainer.html).
+
+# Rationale
+
+Please see the individual PRs for the rationale behind each PR. 
+
+# Implementation
+
+Please see this [repo](https://github.com/UMAprotocol/protocol/tree/master/core/contracts/financial-templates/implementation) for the implementation of this financial contract template.
```
mrice32, line 135: nits:
1. We should refer to that as a directory, not a repo (since it's a directory within our repo, not the entire repo).
2. I think we should be a little more specific here. [This](https://github.com/UMAprotocol/protocol/blob/master/core/contracts/financial-templates/implementation/ExpiringMultiParty.sol) is the implementation of the contract, itself. [This](https://github.com/UMAprotocol/protocol/blob/master/core/contracts/financial-templates/implementation/ExpiringMultiPartyCreator.sol) is the deployer/factory/template (however you want to verbalize it) contract that will be registered with the to allow users to deploy their own `ExpiringMultiParty` contract.

UMIPs/umip-newcontract.md
```diff
@@ -0,0 +1,140 @@
...
+
+Please see the individual PRs for the rationale behind each PR. 
+
+# Implementation
+
+Please see this [repo](https://github.com/UMAprotocol/protocol/tree/master/core/contracts/financial-templates/implementation) for the implementation of this financial contract template. 
+
+# Security considerations
+
+Please see the individual PRs for details on how each affects the security of the UMA ecosystem. This repo will be audited by OpenZeppelin, and the audit report will be made available [here](https://docs.umaproject.org/uma/index.html).
```
mrice32, line 139: I think we should give a more detailed overview here. In this case, we should probably talk about the security implications of adding a new contract template.

Adding a new contract template will:
1. Allow instances of that contract to send requests to the DVM according to its code.
2. Give the creator/factory contract to register new contracts to use the DVM.

UMA tokenholders should be sure that the EMP contract follows the rules of DVM use as defined [here](https://docs.umaproject.org/uma/oracle/dvm_interfaces.html). They should also be verify that the factory contract only registers the instances of the EMP contract template and not other, arbitrary contracts.

@allisonlu22 do you think we should talk about the security of the contract template, itself, here? Or should we only talk about the security implications for the DVM/Oracle system (since that's what the UMA tokenholders are primarily tasked with protecting).

UMIPs/umip-newcontract.md
```diff
@@ -0,0 +1,140 @@
...
+This calls on the UMA DVM to return the value of the token’s price identifier at the expiration timestamp.
+
+After this value is returned to the contract and the contract is settled, any token holder can redeem the tokens against the contract. 
+Redemption of the tokens returns the token holder collateral equal to the price identifier value returned by the UMA DVM. Sponsors may also withdraw any excess collateral from the contract after settlement.
+
+Additional details can be found [here](https://docs.umaproject.org/uma/synthetic_tokens/explainer.html).
+
+# Rationale
+
+Please see the individual PRs for the rationale behind each PR.
```
mrice32, line 131: We don't really refer to any PRs here, so I'm not sure this is particularly relevant. What's the difference between rationale and motivation? Should we remove this section (at least for this UMIP)?

UMIPs/umip-newcontract.md
```diff
@@ -0,0 +1,140 @@
+# Headers
+| UMIP #     |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Register new financial contract with DVM                                                                                                 |
+| Authors    | Matt Rice (matt@umaproject.org), Prasad Tare (prasad@umaproject.org), Chris Maree (chris@umaproject.org), Nick Pai (nick@umaproject.org) |
```
mrice32, line 5:
```suggestion
| Authors    | Matt Rice (matt@umaproject.org), Prasad Tare (prasad@umaproject.org), Chris Maree (chris@umaproject.org), Nick Pai (nick@umaproject.org), Regina Cai (regina@umaproject.org) |
```

UMIPs/umip-newcontract.md
```diff
@@ -0,0 +1,140 @@
+# Headers
+| UMIP #     |                                                                                                                                          |
```
mrice32, line 2:
```suggestion
| UMIP-4     |                                                                                                                                          |
```

UMIPs/umip-newcontract.md
allisonlu22: in the future, convos about contract template security may occur. But as a bare minimum starting point, I think ti's ok just to talk about security implications for DVM since taht's what $UMA holders are tasked with protecting.

UMIPs/umip-newcontract.md
allisonlu22: The idea here is to have a conversation about different ways to implement each feature. For example, deciding to have liquidators post a bond instead of setting a minimum partial liquidation size. 

Since you guys had most of that conversation in the PRs themselves, I thought it would be relevant to leave it to that section.

UMIPs/umip-newcontract.md
allisonlu22: this section might be a good place to discuss how the contract implements paying fees to the store, since that's a DVM security consideratio.

UMIPs/umip-newcontract.md
rcai1: Per the instructions in https://github.com/UMAprotocol/UMIPs/blob/master/umip-template.md: 

> Note that an UMIP number will be assigned by an editor. When opening a pull request to submit your UMIP, please use an abbreviated title in the filename, umip-draft_title_abbrev.md.

I would expect you or @allisonlu22 to be an editor and assign the UMIP number?

mrice32: Oh, I see. Didn't realize that was how this was supposed to work. Thanks for the heads up :).

UMIPs/umip-newcontract.md
rcai1: Thanks - I see the suggestion you added and will add it to the next commit!

UMIPs/umip-newcontract.md
rcai1: Thanks - I see your suggestion and added it to the commit! Will resolve this.

UMIPs/umip-newcontract.md
rcai1: Thank you - I've reflected these changes in this commit: https://github.com/UMAprotocol/UMIPs/pull/7/commits/c2b41f506f4b6f518ac2e9156ef0238a6c1f322e

UMIPs/umip-newcontract.md
```diff
@@ -0,0 +1,140 @@
...
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Register new financial contract with DVM                                                                                                 |
+| Authors    | Matt Rice (matt@umaproject.org), Prasad Tare (prasad@umaproject.org), Chris Maree (chris@umaproject.org), Nick Pai (nick@umaproject.org) |
+| Status     | Draft                                                                                                                                    |
+| Created    | April 30, 2020                                                                                                                           |
+
+
+# Summary
+
+This UMIP registers the `ExpiringMultiParty` template with the DVM.
```
allisonlu22, line 12: rewrite to: "This UMIP provisionally registers the..."

UMIPs/umip-newcontract.md
```diff
@@ -0,0 +1,140 @@
...
+This is the first “priceless” financial contract template to be registered with the DVM. 
+
+# Technical Specification
+
+“Priceless” synthetic tokens are synthetic tokens that are securely collateralized without an on-chain price feed. These tokens are designed with mechanisms to incentivize token sponsors (those who create synthetic tokens) to properly collateralize their positions. These mechanisms include a liquidation and dispute process that allows token holders to be rewarded for identifying improperly collateralized token sponsor positions. The dispute process relies on an oracle, the UMA DVM, to settle disputes regarding liquidations.
+
+A deployment of a priceless synthetic token is defined by the following parameters. The contract template that is registered with the UMA DVM will hardcode or restrict some of these parameters, as indicated below. All other parameters can be customized in individual deployments of the contract. 
+
+- expirationTimestamp: These are discretized to monthly expires on the 1st day of each month at 12:00:00 UTC. The maximum timestamp is June 30, 2021. 
+- collateralAddress
```
allisonlu22, line 42: here is where you should discuss which specific collateral addresses we are looking to register. for this first one i think it should only be Dai.

allisonlu22: It also means that if we wanted to STOP supporting dai in the future, we would have to de-register this factory.

UMIPs/umip-newcontract.md
allisonlu22: also reminder to add in the proposal to add final fees of 10 dai here

allisonlu22: I copied this from UMIP-2 and seems like it can be relevant here:
_Anyone deploying a new priceless token contract should take care to parameterize the contract appropriately to avoid the loss of funds for synthetic token holders. Additionally, the contract deployer should ensure that there is a network of liquidators and disputers ready to perform the services necessary to keep the contract solvent._ 

I think we should also add in the idea of provisional registration as a security consideration. Something like this:
_If approved, this would be the first contract factory registered with the DVM on mainnet. Although extensive security reviews indicate that the factory contract poses no threat to ecosystem security, registration should only be provisional and usage should be limited. $UMA-holders should evaluate the health and solvency of token contracts deployed from this factory contract and contemplate de-registering this factory if any contract counterparty security holes are identified, or if the aggregate amount of user funds at risk in token contracts deployed from this factory exceeds 1M dai before July 1, 2020. Additional identifiers should not be registered for use with this factory contract or the DVM until a solvent and functioning synthetic token contract is on mainnet for at least 30 days._

mrice32: @allisonlu22 ^I like that a lot. Totally up to you, but it also may make sense to add something in there about $UMA-holders being ready to call for emergency shutdown of any contracts that are being misused.

PR #10, opened 2020-05-05
Update umip-newcontract.md

allisonlu22:
I added a discussion of what we considered in choosing dai as the collateral currency
i added a discussion of final fees, and propose to initialize them to 10 dai
I tweaked the security considerations section to reflect provisional registration
i added line breaks to conform to matt's suggested readability suggestion


UMIPs/umip-newcontract.md
```diff
@@ -22,24 +22,31 @@ The proposed contract design has the following features:
...
+A deployment of a priceless synthetic token is defined by the following parameters. The contract template that is 
+registered with the UMA DVM will hardcode or restrict some of these parameters, as indicated below. All other 
+parameters can be customized in individual deployments of the contract. 
 
-- expirationTimestamp: These are discretized to monthly expires on the 1st day of each month at 12:00:00 UTC. The maximum timestamp is June 30, 2021. 
-- collateralAddress: DAI. To register USDC in the future, a new factory contract that has the USDC address in the collateral address line would need to be deployed. To stop supporting DAI the future, this factory and any other factory with DAI as the collateralAddress would have to be de-registered. 
+- expirationTimestamp: These are discretized to monthly expires on the 1st day of each month at 12:00:01 UTC. The max timestamp is June 30, 2021.
```
mrice32, line 48: Why the change to `12:00:01`?

UMIPs/umip-newcontract.md
```diff
@@ -48,19 +55,36 @@ A deployment of a priceless synthetic token is defined by the following paramete
...
-- timerAddress: Hardcoded to 0x0000000000000000000000000000000000000000.
+- timerAddress: Hardcoded to 0x0000000000000000000000000000000000000000
 - withdrawalLiveness: Hardcoded to 120 min.
 - liquidationLiveness: Hardcoded to 120 min.
 
+Since this is the first token factory that proposes using Dai as the collateralAddress, this proposal also includes 
+initializing the finalFees for Dai to 10 Dai for all contracts that use Dai as the collateral currency. The finalFee 
+parameter is an adjustable DVM parameter controlled by $UMA holders to establish a minimum fee for each price request 
+that a contract makes. It is intended to introduce a cost to prevent malicious users from spamming the DVM and overloading 
+it with price requests, but not be so high as to deter disputers from performing their work.
```
mrice32, line 66: Not super opinionated, but it *might* make sense to note that the final fee is added to the "bond" that liquidators *and* disputers pay -- meaning it only adds slightly more downside to both rather than imposing a fixed cost on either. The wording here makes it sound like disputers just pay the final fee. Totally cool with not mentioning it too.

UMIPs/umip-newcontract.md
```diff
@@ -137,12 +165,20 @@ The directory contains both the [implementation](https://github.com/UMAprotocol/
 
 # Security considerations
 
-Please see the individual PRs for details on how each affects the security of the UMA ecosystem. This repo will be audited by OpenZeppelin, and the audit report will be made available [here](https://docs.umaproject.org/uma/index.html).  
+Please see the individual PRs for details on how each affects the security of the UMA ecosystem. This repo has been audited by OpenZeppelin, and the final audit report will be made available [here](https://docs.umaproject.org/uma/index.html).  
+
+Anyone deploying a new priceless token contract should take care to parameterize the contract appropriately to avoid the 
+loss of funds for users. Additionally, the contract deployer should ensure that there is a network of liquidators and 
+disputers ready to perform the services necessary to keep the contract solvent.
```
mrice32, line 172: nit: I think we should also say that anyone planning to *use* a synthetic token (as a sponsor or tokenholder) should validate that the token is solvent and well-parameterized before using since there are configurations and states that make the proposed contract unsafe.

UMIPs/umip-newcontract.md
allisonlu22: gonna ignore then

UMIPs/umip-newcontract.md
allisonlu22: mistake

UMIPs/umip-newcontract.md
allisonlu22: great idea

PR #23, opened 2020-07-10
Create umip-6

claytonroche:
initial draft for umip 6


UMIPs/umip-6.md
```diff
@@ -0,0 +1,73 @@
+# Headers
+| UMIP-6     |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add ETHUSD and USDETH as price identifiers              |
```
nicholaspai, line 4: Curious why these identifier names don't have "/"'s but they do in #24

PR #25, opened 2020-07-16
Create umip-8

claytonroche:
draft for umip 8.  The preview tab isn't showing the markdown formatted (only the raw) -- Not sure if that's my mistake or a preview issue.


UMIPs/umip-8.md
```diff
@@ -0,0 +1,49 @@
...
+This UMIP will have the effect of introducing a new global whitelist contract that will allow new collateral currencies to be added without deploying any new contracts. Previously a new financial contract template would need to be deployed to use a new collateral currency. 
+
+
+## Motivation
+Up until now, to update the collateral currencies, one would need to deploy a new financial contract template since financial contract templates were expected to have a frozen and limited list of collateral currencies. To make the protocol more scalable, it seems sensible to make adding a collateral currency to all existing financial contract templates as easy as proposing a simple governance vote that would approve two Governor transactions: one to add it to the whitelist and another to add a flat final fee for that currency.
+
+
+## Technical Specification
+To accomplish this upgrade, a few actions will need to be taken:
+- A new AddressWhitelist contract will need to be deployed.
```
chrismaree, line 19: This is a reference to code. do we want to consider using appropriate formatting or not?
```suggestion
- A new `AddressWhitelist` contract will need to be deployed.
```

mrice32:
> High level comment: once collateral currencies are whitelisted, how are we going to update their final fee in the store? Without a final fee set, financial contracts that use those currencies will be susceptible to DDOS attack vectors

@nicholaspai  That's only a problem if the final fee is updated to 0, right? We can easily change it from 5 to 10, for example, without opening DoS vectors, I think. Is that your question or am I misreading what you're getting at?

mrice32:
> LGTM! Short and sweet. there is a fair amount of formatting inconsistencies in this markdown file like extra blank lines/missing blank lines that should be changed for consistency sake. Other than this, the copy is great.

@chrismaree updated. PTAL.

PR #26, opened 2020-07-16
Create umip-9

claytonroche:
draft for umip 9


mrice32:
@claytonroche I think the same issue exists here -- I think the file name should be `umip-9.md` rather than `umip-9`.

UMIPs/umip-9
```diff
@@ -0,0 +1,55 @@
...
+
+
+
+## Security considerations
+As before, anyone deploying a new priceless token contract should take care to parameterize the contract appropriately to avoid the loss of funds for users. Additionally, the contract deployer should ensure that there is a network of liquidators and disputers ready to perform the services necessary to keep the contract solvent. Anyone planning to use a synthetic token (as a sponsor or tokenholder) should validate that the token is solvent and well-parameterized before using since there are configurations and states that make the proposed contract unsafe.
```
allisonlu22: can you add the comment you wrote here about how emergencyshutdown would work? 

I think you said that setting final fees to 0 basically prevents contracts from using the collateral type again. However, if there already existed a token contract using the ccollateral type, then that existing contract just gets "free" final fees. do we have any way of also emergency shutdowning existing contracts? we should make a note of that here.

UMIPs/umip-9.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+## Technical Specification
+To accomplish this upgrade, a new financial contract template must be deployed:
+- A modified ExpiringMultiPartyCreator contract will need to be deployed. The notable changes include:
+	- On each attempted deployment, this contract should query the “CollateralWhitelist” address using the Finder contract to find the whitelist address. Then it should verify that the provided collateral currency is on said whitelist.
+	- On each attempted deployment, this contract should allow any timestamp to be provided as the expiry. Note: the ExpiringMultiParty contract has a check that requires the timestamp to be in the future at deployment time.
+	- On each attempted deployment, this contract should allow the deployer to pass in the withdrawal and liquidation liveness, and it should check that they are both > 0.
+
+After deployment, this new ExpiringMutliPartyCreator contract should be approved as a ContractCreator in the Registry.
```
nicholaspai, line 29: Typo: ExpiringMultiPartyCreator

UMIPs/umip-9
nicholaspai: Wait, I don't think setting final fees to 0 prevents the collateral type from being used again. If a contract attempts to pay 0 final fees, then it will be blocked from doing so, but I think its allowed technically for a final fee to be 0. They are all 0 by default I believe.

UMIPs/umip-9.md
```diff
@@ -0,0 +1,55 @@
...
+
+
+## Technical Specification
+To accomplish this upgrade, a new financial contract template must be deployed:
+- A modified ExpiringMultiPartyCreator contract will need to be deployed. The notable changes include:
+	- On each attempted deployment, this contract should query the “CollateralWhitelist” address using the Finder contract to find the whitelist address. Then it should verify that the provided collateral currency is on said whitelist.
+	- On each attempted deployment, this contract should allow any timestamp to be provided as the expiry. Note: the ExpiringMultiParty contract has a check that requires the timestamp to be in the future at deployment time.
+	- On each attempted deployment, this contract should allow the deployer to pass in the withdrawal and liquidation liveness, and it should check that they are both > 0.
```
chrismaree, line 27: I feel that this is not really a sufficient condition. if I set the withdrawal and liquidation liveness to 1 second the contracts will _work_ but all economic assumptions and incentives break down. Do we not want to consider some kind of bound on these? it would be pretty arbitrary but could help protect the contracts from being poorly parameterized. 

I'm fine to leave it like this but just wanted to hear more opinions on this and the security considerations.

UMIPs/umip-9
mrice32: @allisonlu22 Yeah, @nicholaspai is right. Final fees are completely unaffected and uninvolved with this UMIP. New collateral currencies _should_ have a final fee _and_ be whitelisted. However, the whitelist is 100% divorced from the final fees. This means that if you wanted to deprecate a collateral currency, you could remove it from the whitelist but leave the final fee set as a way to ensure that existing contracts that use it continue to pay nonzero fees on DVM requests.

Does that make sense?

UMIPs/umip-9.md
mrice32: Fixed!

UMIPs/umip-9.md
mrice32: I think we've discussed this on the `protocol` repo, but my opinion is that there are particularly bad consequences for setting this to 0 (flash loans, immediate withdrawals, etc).

Outside of those extreme circumstances, I think it's best to leave this parameter to the judgement of the deployer simply because:
1. I think we should generally err on the side of _allowing_ deployers to do what they want, even if we can't see the use-case ahead of time. Deployers know better than we do.
2. It's difficult to determine what's _safe_, especially in a fast-changing Ethereum environment where block times, gas prices, and chain congestion are always moving.

But definitely open to other opinions here! I totally get where you're coming from @chrismaree.

UMIPs/umip-9
allisonlu22: got it, thanks.

UMIPs/umip-9.md
nicholaspai: >I think we should generally err on the side of allowing deployers to do what they want, even if we can't see the use-case ahead of time. Deployers know better than we do.

I think this is the key point.

However, I agree with @chrismaree 's caution. If any financial contract were to use a withdrawal/liquidation liveness of say < 5 minutes, then it would make it very dangerous for any token holders or sponsors respectively.

chrismaree: I agree with your points in that keeping the contracts general and open to being used in different and unique ways is pretty useful. 

My point really was is that if it was set to say `1`, the contract factory would still accept the params and deploy the contract. However, the contract would not function as intended as the economic assumptions breakdown. How we intend to preform liquidations and disputes would no longer function.

With that said, letting the developer choose is properly fine as actually deploying one of these contracts has a pretty high barrier to entry.

mrice32: Totally agreed. I kinda think it's the deployer's responsibility to answer the question: "How we intend to preform liquidations and disputes?". I think that's a somewhat freeing idea from a governance perspective.

PR #38, opened 2020-08-11
Create UMIP-13.md

auryn-macmillan:
Start UMIP-11 draft


adrianmcli:
@auryn-macmillan much apologies for the delay. I think this is a fascinating idea, but it would be great to see a more detailed draft before we consider merging this in. A lot of UMIP discussion happens more on the Discord than Github (for better or worse) so maybe it'd be better to start there?

PR #43, opened 2020-08-31
Create upp-2.md

mrice32:
UPP 2 proposes an increase of the final fees for WETH and DAI.

### DIFF
--- a/UPPs/upp-2.md
+++ b/UPPs/upp-2.md
@@ -4,7 +4,7 @@ Draft
 
 ## Summary
 
-This UPP will increase the final fees for DAI and WETH to be around $200. This will increase the cost of initiating disputes for the DVM.
+This UPP will increase the final fees for DAI and WETH to be around $200. This will increase the cost of initiating disputes for the DVM. The renBTC fee is already 0.018, which is close to the target of $200, so it should not need to be changed.
 
 ## Rationale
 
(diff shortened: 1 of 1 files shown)


UPPs/upp-2.md
```diff
@@ -0,0 +1,19 @@
...
+
+It is extremely expensive, in terms of gas, for the DVM voters to vote on a price request. The DVM also pays out a relatively large number of rewards on each vote.
+To ensure that price requests remain rare, they need to be more expensive to discourage any sort of purposeful triggering.
+
+While this does make the capital requirements for liquidators and disputers higher, the increased size of the final fee lockup is unlikely to deter honest participants.
+
+## Specifics
+
+The WETH final fee will be increased to 0.5 WETH.
+The DAI final fee will be increased to 200 DAI.
```
chrismaree, line 19: Given the recent price increase in $UMA, I think these should be at least 2x this. I propose a target of 400USD so 1 WETH and 400 DAI.

Also, this specification does not mention renBTC which was added in UMIP-11

mrice32: Good point, I'll add it.

mrice32: @hlambur @nicholaspai thoughts?

hlambur: The increase makes sense to me.

mrice32: @chrismaree I have updated this to reflect a $400 final fee. PTAL.

chrismaree (approved):

PR #46, opened 2020-09-04
umip-xx draft

aliatiia:
initial commit


UMIPs/umip-xx.md
```diff
@@ -0,0 +1,66 @@
...
+
+A straight-forward yet sufficiently flexible aggregatory statistics that can serve as a proxy for the gas price in a given block is the $i$th percentile. But there is a block each 12-15 seconds in the Ethereum blockchain, and spikes in gas prices are routinely observed. Hence, a second-order statistic is also needed to smooth-out outliers. It may therefore be desireable to request from the DVM an aggregatory statistic over a range of blocks. Hence, the median of the $i$th percentiles over a range of blocks can be reported.
+
+
+## Implementation
+
+Data point requestors specify three positive integers: `B`, `R`, `i` where `B` is an Ethereum block number, `R>=B` is the maximum block implying the desired range is [`B`,`B+1`, ..,  `B`+`R`], and `P` is the desired percentile and must be in the set `{0, 1, 2, .., 100}`. For `i=0` and `i=100`, the requester is effectively asking for the median of all the `min`'s and `max`'s of gas prices, respectively. For all other values, the requester is asking for `i`th percentile.
```
nicholaspai, line 44: First of all thanks for this awesome entry, I think gas price synthetics are fascinating and useful. One thing to note is that UMA price requests takes in one parameter, `time`, and returns a `price` for a given identifier. In this case, I think the `time` would map 1-to-1 with `B` (block #), and the `price` would be the gas price at that time.

In this UMIP we could keep `R` as a constant to allow for some buffer or one-off spikes in gas.

aliatiia: In a way specifying a block number is a more deterministic way of specifying `time`, since otherwise the DVM would have to deal with different timestamps in different Ethereum nodes which is less reliable.

There is no universal gas `price` since there is no orderbook exchanges where it is bought and sold. So it has to be derived from all the `price`'s that each tx paid in a block. 

> In this UMIP we could keep R as a constant to allow for some buffer or one-off spikes in gas.

We could indeed impose a cap on `R` to make the statistic calculation less intensive for reporters. A `MAX_R=7000` would imply ~1-day worth of blocks to aggregate and report, which is reasonable I think.

nicholaspai: > In a way specifying a block number is a more deterministic way of specifying time, since otherwise the DVM would have to deal with different timestamps in different Ethereum nodes which is less reliable.

I agree with you, but I'm not sure how a price request would work for a specific block height (or any other arbitrary identifier) without having to register a new price identifier for each block #. @mrice32 what do you think here?

> We could indeed impose a cap on R to make the statistic calculation less intensive for reporters. A MAX_R=7000 would imply ~1-day worth of blocks to aggregate and report, which is reasonable I think.
This is totally reasonable and we can specify this in the UMIP so that voters know exactly what they should be calculating.

UMIPs/umip-xx.md
```diff
@@ -0,0 +1,67 @@
...
+    Median = None
+    Percentiles = []
+    for b in range(B, B+R, 1):      
+        transactions = provider.getTransactions(B)
+        gas_prices = []
+        for tx in transactions:
+            gas_prices.append(tx.gasPrice)
+        Percentiles.append(math.percentile(gas_prices, i)) # ensure determinism here
+    return math.median(Percentiles) # ensure determinism here
+```
```
mrice32, line 60: This is interesting. It looks as if you're looking at the median _transaction_ gas price within this block range. I'm curious why you chose that aggregation method. Some other options I'm thinking about:
- Taking the average gas price _per unit gas_ over the period.
- Taking the average unit gas price _in each block_ and taking the median of the block gas prices.

I only ask because I'm not sure why we would weight small transactions the same as large transactions since a large transaction (should?) have a bigger impact than a small one on the gas price that the average user pays since it takes up more block space.

Thoughts?

mrice32 (commented):
I think this is awesome! Thanks for the proposal @aliatiia. Would love to see what sort of products could be built on an identifier like this!

One high-level question for you: do you think users might want the price _in dollars_ or _in ETH_ (as you currently have it). I ask because:
- Some users may be denominating their costs in dollars rather than ETH. For them, it may be useful to hedge the dollar value of gas prices.
- There does seem to be a rough correlation between ETH price and gas prices, so this identifier price stream may not be entirely reflective of the _value_ users are paying to miners.

However it could be that it makes sense to somehow compose this identifier with an ETHUSD identifier to create a dollar-denominated gas price identifier in the future. Just thinking out loud :).

aliatiia:
>it could be that it makes sense to somehow compose this identifier with an ETHUSD identifier to create a dollar-denominated gas price identifier in the future

This is an excellent point. It makes sense that most consumers would want the result USD-denominated. So I could add "multiply by `ETHUSD(timestamp)`" to the "Output processing" section of the specs? So in a way a `GASETH` price requested automatically triggers an `ETHUSD(timestamp)` and the final result is basically the multiplication of the two.

I guess if a consumer does in fact want the result ETH-denominated, they can request `USDETH(timestamp)` (=`1/ETHUSD(timestamp)`) separately and multiply the result by it to get back the `GASETH(timestamp)`

>the block number and timestamp are trivially convertible (since the timestamp is part of the block).

Personally I think providing a `blocknumber` and mapping it to the `timestamp` is safer in terms of determinism (oddities in leap years etc?). This also makes calculating the range safely determnistic as well: finding the block number at the edge of range is simply the `start_block`(=`B` in current spec) plus another integer `R` (which we're now gonna hardcode to ~300 or 7000, see below). If a DVM reporter is determining the range based on Unix timestampes, we must introduce `ceil`/`floor` mechanisms to ensure all reporters will always land on the same range. But maybe Im just too worried about determinism here. 

So TLDR of this point, two ways: (1) supply a block number and we specify a timestamp deduction mechanism in the "Input processing" section, or (2) supply a timestamp and we specify a determnistic timestamp->blocknumber mapping in the implementation. 

**Let me know which you prefer and I'll spec it out.**

>For the rest of these parameters, I might suggest setting them as fixed for this price identifier (with a plan to either offer new price identifiers with slightly different parameters in the future or change the way the DVM receives requests to include params).

>To that end, I might suggest a specification like the following:

>>The value of this price identifier is defined as the median block gas price of the most
>>recent `7000` blocks whose timestamp is <= the request timestamp.

Sure I think this should work for now, but I suggest `R=300` (which would correspond to ~1 hour woth of blocks) instead of `7000`. Price feed consumers can submit 24 separate requests and use that to deduce the same result had `R` been `R=7000` (one day worth of blocks). **The inverse however would be difficult**, i.e. if consumers are provided the median over `7000` it would be difficult to use it to reliably extrapolate a specific hour within that range  without triggering more disputes.

And for percentiles, by specifying "median" we're effectively hardcoding the 50th percentile for now. I think that should be good too for now.

UMIPs/umip-xx.md
aliatiia: I think averaging is problematic considering e.g. that many tx's are from miners to their pool participants to distribute their rewards. Miners tend to specify 1 Wei `gasPrice` in those tx (it goes back to themselves anyway). So I choose median to avoid such outliers affecting skewing the result. 

>  as large transactions since a large transaction (should?) have a bigger impact than a small one on the gas price

Hmm, this is a good point, but I feel the effect of these mega transactions (and high-throughput gas consumers generally, such as ["gas miners"](https://twitter.com/aliatiia_/status/1301247233065717761?s=20)) is already factored in in the `gasPrice` distribution, because they do indeed push the gas prices up and small tx adapt their bids accordingly. So we would be double-counting their impact sort of were we to weight a consumed gas unit by the size of tx it was consumed in.

UMIPs/umip-xx.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+## Implementation
+
+Data point requestors specify three positive integers: `B`, `R`, `i` where `B` is an Ethereum block number, `7000>=R>=B` is the maximum block implying the desired range is [`B`,`B+1`, ..,  `B`+`R`], and `P` is the desired percentile and must be in the set `{0, 1, 2, .., 100}`. Restricting `R` to 7000 implies a maximum range of ~1 day worth of Ethereum blocks which imposes a cap on how much data processing the DVM reporters must do. For `i=0` and `i=100`, the requester is effectively asking for the median of all the `min`'s and `max`'s of gas prices, respectively. For all other values, the requester is asking for `i`th percentile.
```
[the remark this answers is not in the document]
aliatiia: >it could be that it makes sense to somehow compose this identifier with an ETHUSD identifier to create a dollar-denominated gas price identifier in the future

This is an excellent point. It makes sense that most consumers would want the result USD-denominated. So I could add "multiply by `ETHUSD(timestamp)`" to the "Output processing" section of the specs? So in a way a `GASETH` price requested automatically triggers an `ETHUSD(timestamp)` and the final result is basically the multiplication of the two.

I guess if a consumer does in fact want the result ETH-denominated, they can request `USDETH(timestamp)` (=`1/ETHUSD(timestamp)`) separately and multiply the result by it to get back the `GASETH(timestamp)`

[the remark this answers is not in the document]
mrice32: > > it could be that it makes sense to somehow compose this identifier with an ETHUSD identifier to create a dollar-denominated gas price identifier in the future
> 
> This is an excellent point. It makes sense that most consumers would want the result USD-denominated. So I could add "multiply by `ETHUSD(timestamp)`" to the "Output processing" section of the specs? So in a way a `GASETH` price requested automatically triggers an `ETHUSD(timestamp)` and the final result is basically the multiplication of the two.
> 
> I guess if a consumer who does in fact want the result ETH-denominated, they can request `USDETH(timestamp)` (=`1/ETHUSD(timestamp)`) separately and multiply the result by it to get back the `GASETH(timestamp)`

Yep, but I think for simplicity, that composability is probably easier done in the DVM than forcing the consumer to worry about it. I could imagine a GASUSD and a GASETH identifier that are formulated slightly differently under the hood, but are transparent to the consumer. This UMIP could propose supporting both. One would just require the voters to compute an ETHUSD price using the formulation given in the pre-existing ETHUSD identifier (it could just be referenced in this proposal).

Thoughts?

> > the block number and timestamp are trivially convertible (since the timestamp is part of the block).
> 
> Personally I think providing a `blocknumber` and mapping it to the `timestamp` is safer in terms of determinism (oddities in leap years etc?). This also makes calculating the range safely determnistic as well: finding the block number at the edge of range is simply the `start_block`(=`B` in current spec) plus another integer `R` (which we're now gonna hardcode to ~300 or 7000, see below). If a DVM reporter is determining the range based on Unix timestampes, we must introduce `ceil`/`floor` mechanisms to ensure all reporters will always land on the same range. But maybe Im just too worried about determinism here.
> 
> So TLDR of this point, two ways: (1) supply a block number and we specify a timestamp deduction mechanism in the "Input processing" section, or (2) supply a timestamp and we specify a determnistic timestamp->blocknumber mapping in the implementation.
> 
> **Let me know which you prefer and I'll spec it out.**

Just because the DVM's interface at the moment is set up to _only_ take timestamps, I think having the requester specify a timestamp and constructing some deterministic mapping to a range of blocks is the most expedient approach for now. I totally hear your reasoning, though. And I want the DVM to become more customizable down the line.

> > For the rest of these parameters, I might suggest setting them as fixed for this price identifier (with a plan to either offer new price identifiers with slightly different parameters in the future or change the way the DVM receives requests to include params).
> 
> > To that end, I might suggest a specification like the following:
> 
> > > The value of this price identifier is defined as the median block gas price of the most
> > > recent `7000` blocks whose timestamp is <= the request timestamp.
> 
> Sure I think this should work for now, but I suggest `R=300` (which would correspond to ~1 hour woth of blocks) instead of `7000`. Price feed consumers can submit 24 separate requests and use that to deduce the same result had `R` been `R=7000` (one day worth of blocks). **The inverse however would be difficult**, i.e. if consumers are provided the median over `7000` it would be difficult to use it to reliably extrapolate a specific hour within that range without triggering more disputes.
> 
> And for percentiles, by specifying "median" we're effectively hardcoding the 50th percentile for now. I think that should be good too for now.

Agreed on the 50th percentile.

I agree with your suggestion that the 1 day median is lossier, but if this identifier were being used to develop a future, wouldn't the user feel more comfortable (and less manipulatable) if the settlement price of that future were based on a 1-day median rather than a 1 hour median? It's true that the financial contract could make many DVM requests and do post processing on the results to compute something _close_ to the 1 day median, although I'm not sure it could be done perfectly. But more practically, I think constructing it that way would require the contract to make many DVM requests, which would be expensive because each request comes with a fee. The financial contract would also have to do post-processing to combine the medians, which would add complexity to the implementation. Note: this sort of post processing isn't possible in the ExpiringMultiParty contract, which backs the yUSD and uUSD contracts.

Thoughts on setting this to a 1 day median so most financial contracts could just use the value outright?

UMIPs/umip-xx.md
mrice32: These are both good points. Both of these types of transactions are quite problematic, but are difficult to weed out in a systematic way.

Maybe one way to keep using a median (to remove outliers) but continue to weight by units of gas used rather than number of transactions sent would be a weighted median computed like the following:
- Sort all transactions from the block range by gas price.
- Sum up the gas used across all of these txns.
- Keep a rolling sum of gas used starting from the lowest transaction until the `rolling_sum > total_gas / 2`.
- Take the gas price of the last txn you processed.

This would effectively give you the median txn gas price _weighted by gas used_.

Note: I think using a median also makes this value much harder to manipulate.

Thoughts?

aliatiia: > This would effectively give you the median txn gas price weighted by gas used.

I like this. It's also straightforward to compute so no non-determinism concerns. Ok I will amend the UMIP accordingly.

UMIPs/umip-xx.md
```diff
@@ -0,0 +1,78 @@
...
+The definition of this identifier should be:
+- Identifier name: GASETH-1HR GASETH-3HR GASETH-24HR
+- Base Currency: ETH
+- Quote Currency: GAS
+- Sources: any Ethereum full node
+- Result Processing: Exact
+- Input Processing: see Implementation section
+- Price Steps: 1 Wei
+- Rounding: Closest: N/A becauses Wei's are integers
+- Pricing Interval: 300 (GASETH-1HR), 900 (GASETH-3HR), or 7200 (GASETH-24HR) blocks
```
mrice32, line 30: nit: I think this means the time-granularity of the price. For instance, if this were 1 minute, this would mean that the price would only "update" every minute. This is typically used for price feeds that are queried via apis that have a time-granularity. The minimum granularity is 1 second (since the DVM takes timestamps in terms of seconds), so my guess is that you want that.

UMIPs/umip-xx.md
```diff
@@ -0,0 +1,78 @@
...
+
+The reported price has to take two considerations in mind: (1) there is a block each 12-15 seconds in the Ethereum blockchain, and spikes in gas prices are routinely observed, and (2) miners can easily manipulate gas prices in a given block (especially in absence of a fee burn). Therefore, an aggregatory statistic needs to be computed over a sufficiently long range of blocks to eliminate these two vectors. We propose the median gas price over 300, 900, or 7200 blocks _weighted by_ the gas used in a transaction. These ranges corresponding to 1HR, 3HR, or 24H (1D) worth of blocks. 
+
+## Implementation
+
+The total gas used over the last 300 (GASETH-1HR), 900 (GASETH-3HR) or 7200 (GASETH-24HR) blocks that were mined at or before the `timestamp` provided to DVM is computed. Transactions in this range are also sorted by `total_gas_consumed`.
```
mrice32, line 44: nit: this seems to assume a block time of 12 seconds. If that deviates, will these identifiers be slightly incorrect? It might be more consistent to _only_ use timestamps to define inclusion into the median (since we're already using the timestamps to determine which block to _end_ the median on). This would mean that the `24HR` one, for instance, would include all blocks whose block timestamps satisfy the following inequality: 
```
price_request_timestamp - 86400 < block_timestamp < price_request_timestamp
```

UMIPs/umip-xx.md
```diff
@@ -0,0 +1,78 @@
...
+## Technical Specification
+
+The definition of this identifier should be:
+- Identifier name: GASETH-1HR GASETH-3HR GASETH-24HR
+- Base Currency: ETH
+- Quote Currency: GAS
+- Sources: any Ethereum full node
+- Result Processing: Exact
+- Input Processing: see Implementation section
+- Price Steps: 1 Wei
```
mrice32, line 28:
```suggestion
- Price Steps: 1 Wei (1e-18)
```

UMIPs/umip-xx.md
```diff
@@ -0,0 +1,78 @@
...
+- Rounding: Closest: N/A becauses Wei's are integers
```
mrice32, line 29:
```suggestion
- Rounding: Closest: N/A because the median algorithm as described below cannot produce numbers with higher precision than 1 Wei (1e-18).
```

mrice32 (commented):
This looks great! Just a few minor comments and suggestions.

UMIPs/umip-xx.md
aliatiia: noted, so i'll do:
+ Price Interval: 1 second

UMIPs/umip-xx.md
aliatiia: the issue though with using timestamp to define a range is that the network could conceivably become very slow at finalizing blocks, so in the extreme case the medanization could be over 1 block (if it took >=1HR to finalize a block).
So I think it is safer for the feed to define the range in terms of blocks.

mrice32: While I definitely agree this is _possible_. I think this is a very rare edge case (I'm assuming this has never happened). I don't feel as if guarding against this edge case is worth making the definition of the identifier conflict with its naming (uses time in the name, but blocks in the definition).

One solution is to define the identifier in terms of timestamp, but the definition also says that in extreme cases where the block sample size is very small, the DVM voters can extend the range of blocks considered. Similar to most identifiers, this one can rely on the DVM voters to exercise some subjective judgement to handle extreme cases.

Thoughts?

mrice32 (commented):
I think this looks great! I just have one lingering comment above!

smb2796 (approved):
Renaming to umip-16 and approving as a draft

UMIPs/umip-xx.md
aliatiia: > One solution is to define the identifier in terms of timestamp, but the definition also says that in extreme cases where the block sample size is very small, the DVM voters can extend the range of blocks considered. 

This is reasonable; we would have to make an assumption about what that sanity check would be in number of blocks. For example, we could assume a minimum of 200 blocks per hour (1-block/18s)? that should be ok I think.

PR #58, opened 2020-09-26
Add rDAI as collateral

EndymionJkb:


UMIPs/umip-add rDAI as a collateral currency.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Rationale
+This change encourages wider adoption of the UMA protocol through strengthening ties with traditional financial markets, and promoting a new use case: securing synthetic positions with interest-bearing collateral that benefits charities related to the underlying assets.
+
+It also provide general utility by enabling the use of "compounding collateral," which serves to lower risk, offset losses, and monetize "HODLing" of inactive synthetic positions (as the first such tokens are likely to be), or those with far future expiration dates.
+
+## Implementation
+This change has no implementation other than adding the rDAI address to the collateral currency AddressWhitelist.
+
+## Security considerations
```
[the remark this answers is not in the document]
EndymionJkb: Good idea!

rDAI collateral providers should do their own due diligence on the beneficiaries of any given synthetic before opening a position. This is easy to do on a technical level, since all the "hat"-related information, including the literal Ethereum addresses of recipients, is available on-chain through the token contract interface and corresponding subgraph. (Not to mention the graphical rDAI explorer.) rDAI held in positions would be tamper-proof, as it is owned by the synthetic contract. These addresses (ideally linked to ENS domains) could then be verified through the official web sites of the charitable organizations.

UMIPs/umip-add rDAI as a collateral currency.md
```diff
@@ -0,0 +1,48 @@
...
+
+Of course, this idea can be generalized beyond ESG, to any pairing of a charitable cause to a related ETF or index fund. And since new charities are partnering with rDAI all the time - in fact, through the rDAI protocol anyone can create an arbitrary "charity" simply by providing a list of ETH addresses and weights - this is a growth industry.
+
+A local charitable organization could create a monthly "rolling" synthetic that directed funds according to the organization's priorities that month, either by altering weights, or by creating a new list of beneficiaries.
+
+Alternatively, the protocol also allows rDAI to be used without specifying charitable beneficiaries at all. Undesignated ("self-hat") rDAI could be used simply to offset potential losses and de-risk positions, as the collateral would naturally increase over time. Hybrid positions are also possible; e.g., 50% going to charity, and 50% compounding within the synthetic.
```
mrice32, line 23: How does rDAI accrue interest if no "hat" is specified? Would the contract just effectively be "airdropped" rDAI (kind of like how aDAI works)?

UMIPs/umip-add rDAI as a collateral currency.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Motivation
+One of the most basic use cases for synthetic tokens is derivitives tied to underlying real-world assets, including equities - and a popular way to gain exposure to equities is through ETFs or index funds: baskets of securities chosen according to a published set of criteria.
+
+Increasingly, and particularly among millennials, those criteria involve moral judgments and values. "Value driven investment" funds (e.g., the ETHO Climate Leadership Index) represent a rapidly rising trend in the industry, and often outperform the broader market.
+
+An UMA synthetic tied to an "ESG" index would be a great way to "long the environment." But an even greater way would be to use rDAI as collateral for the index position, designating an environmental charity (e.g., Offsetra or rTrees) as the beneficiary. That way, the charity always benefits, even if the index underperforms. (Perhaps especially then, if position holder needs to deposit more rDAI to avoid liquidation.)
```
mrice32, line 17: How does a contract change its "hat" value to delegate its interest? Would this require a custom financial contract that knows how to delegate its interest?

mrice32 (commented):
I think this is a really cool idea! I think it might be nice to go into a bit more detail about how a rDAI accrues in the self-hat and non self-hat cases just so it's clear what changes, if any, would need to be made to the financial contracts to handle rDAI! Happy to discuss the details on discord if anything is unclear!

UMIPs/umip-add rDAI as a collateral currency.md
EndymionJkb: Yes; if you called payInterest(), the rDAI would be sent to the contract address

UMIPs/umip-add rDAI as a collateral currency.md
EndymionJkb: If you wanted to change the hat after creation, yes, the synthetic contract would need to have special logic. In the use cases I'm contemplating though, you wouldn't do that. The synthetic would track an index fund related to a thing, and the charity would also benefit that thing, so it wouldn't make sense to change it.

There is the issue of initially establishing the hat, though. (Once established, new positions would transfer rDAI in, and since all the rDAI is held by the contract itself, all new positions would inherit it.) So I believe you'd have to atomically create the EMP and open a small position in it, to establish the hat for that synthetic.

UMIPs/umip-add rDAI as a collateral currency.md
```diff
@@ -0,0 +1,48 @@
...
+- A final fee of 400 rDAI needs to be added for rDAI in the Store contract.
+## Rationale
+This change encourages wider adoption of the UMA protocol through strengthening ties with traditional financial markets, and promoting a new use case: securing synthetic positions with interest-bearing collateral that benefits charities related to the underlying assets.
+
+It also provide general utility by enabling the use of "compounding collateral," which serves to lower risk, offset losses, and monetize "HODLing" of inactive synthetic positions (as the first such tokens are likely to be), or those with far future expiration dates.
+400 rDai was chosen because the current final fee for DAI is 400 DAI and the value of rDAI is directly equal to the value of DAI.
+## Implementation
+This change has no implementation other than adding the rDAI address to the collateral currency AddressWhitelist and adding the rDAI final fee to the Store contract.
+
+## Security considerations
```
smb2796, line 38:
```suggestion
## Security considerations

rDAI collateral providers should do their own due diligence on the beneficiaries of any given synthetic before opening a position. This is easy to do on a technical level, since all the "hat"-related information, including the literal Ethereum addresses of recipients, is available on-chain through the token contract interface and corresponding subgraph. (Not to mention the graphical rDAI explorer.) rDAI held in positions would be tamper-proof, as it is owned by the synthetic contract. These addresses (ideally linked to ENS domains) could then be verified through the official web sites of the charitable organizations.
```

UMIPs/umip-add rDAI as a collateral currency.md
chrismaree: In this case the contracts total collateral would go up but no sponsor would be able to claim this extra collateral. The EMP has a function https://github.com/UMAprotocol/protocol/blob/3129fc83480531d31ba64af5bb10ac061f93e6bf/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol#L622 which would enable this extra collateral to be drained by the `excessTokenBeneficiary`. 

If you wanted the interest accrued on the collateral to be "clamable" by the sponsors you'd need to use a cToken model, like compounds interest-bearing tokens or Yearn's yUSD.

mrice32: Also, you could just make the excessTokenBeneficiary the desired charitable organization rather than giving the contract a hat. Either would work.

smb2796:
```suggestion
Alternatively, the protocol also allows rDAI to be used without specifying charitable beneficiaries at all. Undesignated ("self-hat") rDAI could be used simply to offset potential losses and de-risk positions, as the collateral would naturally increase over time. Hybrid positions are also possible; e.g., 50% going to charity, and 50% compounding within the synthetic.

When using undesignated ("self-hat") rDAI as collateral for the EMP contract template, a contract sponsor would need to set `excessTokenBeneficiary` to an address they wish to receive the accrued rDAI interest. The EMP function [trimExcess](https://github.com/UMAprotocol/protocol/blob/3129fc83480531d31ba64af5bb10ac061f93e6bf/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol#L622) would then need to be called to drain the excess collateral to the `excessTokenBeneficiary`.
```
I think it is worth describing this process within the UMIP

UMIPs/umip-add rDAI as a collateral currency.md
mrice32: Does the contract, itself, need to make a call to set this initial hat? Or does it automatically inherit it from the first person who sends in rDAI?

EndymionJkb: It should inherit it from the first deposit. So the EMP creator would mint the rDAI with the appropriate hat, then open a nominal position to make the initial rDAI deposit to the contract address, after which all other rDAI deposits would inherit the same hat. (And since the rDAI is owned by the EMP contract after being transferred in, the minter can't change the hat later.)

I suppose there's a slim chance of front-running because of the sequencing (you'd try to do it all atomically in the same block). Especially if you wanted to use the self-hat/trimExcess method, in which case you'd have to create the EMP first so you have its address, before minting the rDAI and opening a position.

The risk is somebody "hijacking" the EMP by making the first deposit before the creator, and setting the hat to their own wallet. You can query the hat on the EMP address to ensure this didn't happen (and I suppose call emergencyShutdown() if it ever did?)

mrice32: I think this is okay. No need to emergency shutdown. Anyone can create any contract they want at any time. So all you'd need to do would be to abandon that contract in favor of one where the hat is set correctly.

mrice32: Could you add a note about this unique property of rDAI and the sensitivity to sequencing of transactions to the UMIP? Just want to make it clear to someone using this currency that this is something to be aware of.

mrice32 (commented):
As long as we add a note about [this thread](https://github.com/UMAprotocol/UMIPs/pull/58#discussion_r499060760) in the technical details or security considerations, this LGTM!

UMIPs/umip-add rDAI as a collateral currency.md
EndymionJkb: Ok; just pushed a note addressing all this in great detail

smb2796: Looks great to me!

smb2796 (approved):
Approving with draft status

PR #60, opened 2020-10-07
Add additional price ids and renaming price ids to reflect block range.

smb2796:
Adding additional price identifiers for ~1 week and ~4 weeks worth of blocks. Rename price identifiers to reflect number of blocks in the calculation rather than amount of time.


mrice32 (approved):
LGTM!

PR #62, opened 2020-10-08
use clock time with sanity minimums, plus 1W 1M tickers

aliatiia:
- clock time is used to define the range over which the gasUsed-weighted median is reported but the DVM imposes a minimum number of blocks within a given range (e.g. a minimum of 200 blocks to medianize over for the GASETH-1HR ticker)

- added 1W and 1M tickers

### DIFF
--- a/UMIPs/umip-xx.md
+++ b/UMIPs/umip-xx.md
@@ -1,7 +1,7 @@
 # Headers
 | UMIP-13     |                                                                                                                                          |
 |------------|------------------------------------------------------------------------------------------------------------------------------------------|
-| UMIP Title | Add GASETH-1HR GASETH-4HR GASETH-24HR as a price identifiers                                                                                                 |
+| UMIP Title | Add GASETH-1HR GASETH-4HR GASETH-1D GASETH-1W GASETH-1M as a price identifiers                                                                                                 |
 | Authors    | Ali Atiia, @aliatiia | Matt Rice, @mrice32
 | Status     | Draft                                                                                                                                    |
 | Created    | September 4, 2020                                                                                                                           |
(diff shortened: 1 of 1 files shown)


smb2796 (approved):
Looks like an elegant solution to both the use of timestamps and having a minimum number of blocks in the range. In my opinion, other than a few formatting items, this looks good to go into Last Call.

aliatiia:
LGTM!

UMIPs/umip-xx.md
```diff
@@ -33,11 +33,22 @@ The definition of this identifier should be:
...
+
+| Identifier | Minimum number of mined blocks |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| GASETH-1HR | 200 |
+| GASETH-4HR | 800 |
+| GASETH-1D | 4800 |
+| GASETH-1W | 33600 |
```
mrice32, line 47: Is there a reasonable way to compute 1 week or 1 month of gas prices in real time (as a bot or UI nay need to do to get the current price)? It seems difficult to do without a fairly sophisticated indexing/caching system.

mrice32 (approved):
Looks great! I only have one small concern about comparability.

UMIPs/umip-xx.md
aliatiia: I don't think you can do it in real time because you would keep **re-sorting**, better do it once at the end. But what would be the benefit of time it in real time rather than at the end of the week/month?

I recently extracted the `gasPrice` of every tx in every block in Ethereum history. It was bandwidth intensive because there is no granual RPC query to just download `gasPrice`s .. you [must download the whole block](https://community.infura.io/t/the-most-bandwidth-efficient-way-of-obtaining-historical-gasprices-per-tx/1723) and extract what you need on-the-fly . 

Looking at that data, I see that 1-month worth of blocks took about ~3.2 minutes to download, and extracting and storing `gasPrice`s took ~10MB disk space (without compression, just utf-8).

mrice32: I think this is totally fine since the voters can always compute it the way you're saying. The issue I'm referring to is computing this price in real-time -- it doesn't only get computed at the end of the month, it must be computed at given points in time to determine whether positions are correctly collateralized. This computation would be difficult. I discussed with @smb2796, and we don't think this should be a blocker for this UMIP. It's just something a product-deployer needs to keep in mind.

PR #68, opened 2020-10-15
Updating GASETH UMIP with an additional calculation implementation

smb2796:
Using a Google BigQuery query that @mrice32 created,  I have added another reference implementation. This will allow for flexibility in the actual usage of this price identifier. I also attempt to make it explicit that DVM reporters are free to define their own implementations of this calculation later, so long as those implementations agree with the calculation methodology.

### DIFF
--- a/UMIPs/umip-16.md
+++ b/UMIPs/umip-16.md
@@ -101,7 +101,7 @@ DECLARE block_count int64;
 DECLARE max_block int64;
 
 -- Querying for the amount of blocks in the preset time range. This will allow block_count to be compared against a given minimum block amount.
-SET (block_count, max_block) = (SELECT AS STRUCT (MAX(number) - MIN(number)), MAX(number) FROM `bigquery-public-data.crypto_ethereum.blocks` WHERE timestamp BETWEEN TIMESTAMP(t2) AND TIMESTAMP(t1));
+SET (block_count, max_block) = (SELECT AS STRUCT (MAX(number) - MIN(number)), MAX(number) FROM `bigquery-public-data.crypto_ethereum.blocks` WHERE timestamp BETWEEN TIMESTAMP(@t2) AND TIMESTAMP(@t1));
 
 CREATE TEMP TABLE cum_gas (
   gas_price int64,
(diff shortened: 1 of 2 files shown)


UMIPs/umip-16.md
```diff
@@ -91,6 +91,48 @@ def return_median(t, N)
...
+            FROM
+                `bigquery-public-data.crypto_ethereum.transactions`
+            WHERE
+                block_timestamp BETWEEN TIMESTAMP(t2)
+                AND TIMESTAMP(t1)
+            GROUP BY
+                gas_price,
+                block_number
+        )
+);
```
mrice32, line 125: Love the formatting! What tool did you use?

UMIPs/umip-16.md
```diff
@@ -91,6 +91,48 @@ def return_median(t, N)
...
+  
+set halfway = (SELECT DIV(MAX(cum_sum),2) from cum_gas);
+
+set block_amount = (SELECT (MAX(block_number) - MIN(block_number)) FROM cum_gas);
+
+select cum_sum, gas_price from cum_gas where cum_sum > halfway order by gas_price limit 1;
```
mrice32, line 131:
```suggestion
SELECT cum_sum, gas_price FROM cum_gas WHERE cum_sum > halfway ORDER BY gas_price LIMIT 1;
```

mrice32 (approved):
LGTM!

UMIPs/umip-16.md
smb2796: Thanks! Found a cool VSCode extension called SQL Formatter -> https://marketplace.visualstudio.com/items?itemName=adpyke.vscode-sql-formatter

daywiss: theres also one built into cloud ui "format"
image

smb2796: Found that one afterwards. It's much better than SQL Formatter

mrice32 (approved):
LGTM! Btw, I'm assuming this is only marginally more expensive than the existing logic, right? If not, we may want to optimize a bit more.

chrismaree (approved):
This is awesome! love that we can do it in google big query. Also, the fact that you get 300 USD + 1 TB of free quota means that this is accessible for voters and bot runners alike. Good work @smb2796

PR #69, opened 2020-10-19
sanity check on timestamp

aliatiia:
- timestamp should be <= timestamp of the highest block on BQ
- if it's not, it is gracefully set to the timestamp of the highest block on BQ

### DIFF
--- a/UMIPs/umip-16.md
+++ b/UMIPs/umip-16.md
@@ -104,7 +104,7 @@ DECLARE latest_timestamp TIMESTAMP;
 -- the latest block in BigQuery may have a timestamp older than @t2, in which case @t2 is set to that latest timestamp instead
 SET latest_timestamp = (SELECT MAX(timestamp) from `bigquery-public-data.crypto_ethereum.blocks`);
 IF t2 > latest_timestamp THEN
-   t2 = latest_timestamp;
+   SET t2 = latest_timestamp;
 END IF;
   
 -- Querying for the amount of blocks in the preset time range. This will allow block_count to be compared against a given minimum block amount.
--- a/UMIPs/umip-16.md
+++ b/UMIPs/umip-16.md
@@ -99,7 +99,14 @@ This query is parameterized with a UTC timestamp `t1`, a time range defined by t
 DECLARE halfway int64;
 DECLARE block_count int64;
 DECLARE max_block int64;
+DECLARE latest_timestamp TIMESTAMP;
 
+-- the latest block in BigQuery may have a timestamp older than @t2, in which case @t2 is set to that latest timestamp instead
+SET latest_timestamp = (SELECT MAX(timestamp) from `bigquery-public-data.crypto_ethereum.blocks`);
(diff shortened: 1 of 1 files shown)


UMIPs/umip-16.md
```diff
@@ -99,7 +99,14 @@ This query is parameterized with a UTC timestamp `t1`, a time range defined by t
...
 DECLARE block_count int64;
 DECLARE max_block int64;
+DECLARE latest_timestamp TIMESTAMP;
 
+-- the latest block in BigQuery may have a timestamp older than @t2, in which case @t2 is set to that latest timestamp instead
+SET latest_timestamp = (SELECT MAX(timestamp) from `bigquery-public-data.crypto_ethereum.blocks`);
+IF t2 > latest_timestamp THEN
+   t2 = latest_timestamp;
+END IF;
+
```
mrice32, line 109: Can you explain the motivation for this change? Will this change the results of any of the queries below? Is the concern about atomicity of the queries?

aliatiia: To handle the case where the user-supplied `t2` happens to be later than the time stamp of the highest block available on BQ. But I see now this may already be handled implicitly in the `sql WHERE timestamp BETWEEN TIMESTAMP(@t2) AND TIMESTAMP(@t1))` clause in: 
```sql
-- Querying for the amount of blocks in the preset time range. This will allow block_count to be compared against a given minimum block amount.
SET (block_count, max_block) = (SELECT AS STRUCT (MAX(number) - MIN(number)), MAX(number) FROM `bigquery-public-data.crypto_ethereum.blocks` WHERE timestamp BETWEEN TIMESTAMP(@t2) AND TIMESTAMP(@t1));

```

So the check I made is I think redundant?

smb2796: Yes, that clause implicitly handles that scenario when the minimum threshold of blocks is not met. The `WHERE block_timestamp BETWEEN TIMESTAMP(@t2) AND TIMESTAMP(@t1)` clause below implicitly handles the same scenario when the minimum threshold of blocks **is** met.

`-- If the minimum threshold of blocks is met, query on a time range
IF block_count >= @B THEN
INSERT INTO cum_gas (
  SELECT
    gas_price,
    SUM(gas_used) OVER (ORDER BY gas_price) AS cum_sum
  FROM (
    SELECT
      gas_price,
      SUM(receipt_gas_used) AS gas_used
    FROM
      `bigquery-public-data.crypto_ethereum.transactions`
    WHERE block_timestamp 
    BETWEEN TIMESTAMP(@t2)
    AND TIMESTAMP(@t1)  
    GROUP BY
      gas_price));`

Also, for clarity's sake, the query is currently written with `t1` as the user supplied timestamp, not `t2`.

> This query is parameterized with a UTC timestamp t1, a time range defined by the price identifier being used (i.e. 24HR) N, a lower bound of time t2 defined by t1 - N * 3600 and a minimum block amount B (explained in the Rationale section).

PR #71, opened 2020-10-26
Add USDC and USDT as whitelisted collateral currencies

smb2796:


mrice32 (dismissed):

chrismaree (dismissed):
Looks good to me! 

One small detail that you could consider adding to the Security considerations section is that USDT is in fact a non-standard ERC20 token. USDT `approve` doesn’t comply with the ERC20 standard. I don't think this matters as we've wrapped all downstream logic in `SafeERC20` but it _could_ have some implementation consequences for some financial contracts and so it should be mentioned.

chrismaree (approved):
Thanks for updating! LGTM!

smb2796:
No problem, thanks for the great comment!

PR #72, opened 2020-10-29
ARSUSD Price Identifier UMIP

smb2796:


UMIPs/umip-19.md
```diff
@@ -0,0 +1,69 @@
...
+There is little cost associated with adding this price identifier, as there are multiple free and easily accessible data sources available.
+
+## Technical Specification
+The definition of this identifier should be:
+
+- Identifier name: ARSUSD
+- Base Currency: ARS
+- Quote Currency: USD
+- Data Sources: Ripio, SatoshiTango, CoinMonitor
+- Result Processing: Mean. BTC/ARS to ARS/USD conversion defined in the implementation section.
```
mrice32, line 29: Why not median?

UMIPs/umip-19.md
```diff
@@ -0,0 +1,69 @@
...
+
+ARSUSD uses Ripio, SatoshiTango and CoinMonitor for BTCARS price information. Ripio and SatoshiTango are Argentinian cryptocurrency exchanges and were chosen because they have some of the highest genuine volumes of ARS/BTC trading. CoinMonitor is a price aggregator that determines its ARS/BTC price as a result of taking the median of a group of buy and a group of sell prices from various cryptocurrency exchanges and then averaging the two. CoinMonitor was chosen as a data source because its price calculation methodology creates a good representation of what the true market rate for BTC/ARS is. All three sources offer free and publicly accessible historical BTC/ARS endpoints.
+
+The decision, to query for BTC/ARS and ARS/USD instead of directly querying for ARS/USD, was made because the ARS/USD pair has a large variance in price reporting, caused by artificial factors. Products built with the ARSUSD price identifier are most likely to be used in Argentinian cryptocurrency exchanges, so the price identifier should use the price reflected by the highest volume crypto-native ARS trading pair.
```
mrice32, line 41: Hmmm, I see the reasoning that BTC/ARS may be easier to query than ARS/USD. However, are you sure doing the multiplication as you have done give a fair/realistic result for ARS/USD? Have you compared it against ARS/USD feeds (like even what google returns, for instance). The only reason I ask is that sometimes BTC and other crypto trades at a strong premium or discount in certain national markets where the exchanges are geofenced and it's difficult to get liquidity in and out from outside the country. The most notable example of this is [SK](https://www.investopedia.com/terms/k/kimchi-premium.asp#:~:text=The%20kimchi%20premium%20is%20the,a%20staple%20in%20Korean%20cuisine.).

mrice32 (commented):
Looks great! A few minor comments.

UMIPs/umip-19.md
smb2796: Should be median. The methodology previously calculated mean, so this is just a missed update

UMIPs/umip-19.md
smb2796: This is a good thing to bring up. There is a difference between the rates that BTC/ARS and ARS/USD return.

A big problem with using ARS/USD directly, other than accessibility, is the noted large variance in price. This is an Argentinian website that reports all of the different exchange rates that are used for ARS/USD.
image

As you can see, there is a large difference between official rates and other exchange methods that are also widely used. This leads to a big question about the method to track the actually used ARS/USD rate.

Since synthetic ARS by nature would be used in cryptocurrency transactions, the argument is that the fairest approach would be to have the value of synthetic ARS track the rate that a crypto user would get by exchanging ARS into BTC and then BTC into USDC (or an equivalent).

manuelgvirtz: The BTC/ARS choice has to do with the fact that Argentina is currently under a capital restrictions scheme, where retail users and companies cannot acces the official exchange rate to buy goods or services. Because of this situation, there are several USD/ARS currently within a ~100% spread between the official rate and the "blue" rate. We consider the fair USD/ARS exchange rate in this case is the one through you can access to purchase cryptocurrencies. BTC/ARS is the pair with the majority of the trading in the country, with an estimate of 80% of the total volume, with more than 10 different brokers operating daily in the country. 
Correctly assesing the USD/ARS exchange rate is a complex task and probably will be even more complex in the near future. However, over the years and different capital restrictions models, BTC has shown to be a robust indicator of fair exchange rate for USD.

mrice32: SGTM!

mrice32 (approved):

PR #73, opened 2020-10-29
Adding UMIP for scaled gas identifiers

smb2796:


UMIPs/umip-20.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Summary (2-5 sentences)
+The DVM should support requests for aggregatory gas prices on the Ethereum blockchain in multiples of a million. This will reflect the price of a million units of gas.
+
+## Motivation
+As specified in price identifiers approved with [UMIP-16](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-16.md), the DVM currently supports reporting aggregatory gas prices of finalized blocks on the Ethereum blockchain. A gas price is the amount of Ether that is paid per unit of gas. 
+
+For the creation of a tokenized gas price futures contract, it is desired that the DVM return the aggregatory gas price for 1 million units of gas. Using the gas price for a million units of gas is more suitable for a tokenized futures contract because tokens will actually represent a non-negligible amount of value. If a token was built with the identifiers defined in UMIP-16, participants would need to transact in millions/billions of tokens to capture any substantial value, as the price of each token would be somewhere in the range of 10-150 Gwei. This is thought to not be an optimal user experience and is the motivation for this UMIP.
```
daywiss, line 15: I think just for clarity on the reasoning, to show a direct comparison to what the token price looks like without this directly followed by what it looks like after this. So i guess 10-150 gwei vs .01 to .15 eth or whatever is price per unit.

UMIPs/umip-20.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Technical Specification
+
+The definition of this identifier should be:
+- Identifier name: GASETH-1HR-1M GASETH-4HR-1M GASETH-1D-1M GASETH-1W-1M GASETH-1M-1M
```
daywiss, line 20: is it worth considering adding longer than 1 month identifier?

UMIPs/umip-20.md
daywiss: I see further down you do this, so you can disregard!

UMIPs/umip-20.md
smb2796: It is definitely worth considering, but for the sake of keeping some sort of scope on the amount of identifiers being added, we scoped down to these. We had this discussion for UMIP-16, and, for the most part, the longest timeframe that was desired for creating a specific product was one month. 

An additional complexity with adding even longer timeframes would be the increased expense and complexity of running the computation. A one month identifier is already relatively expensive and time-consuming to run.

My view is - if eventually somebody came along and wanted to create something like a year long futures contract, they could add the identifier at that time. But for the sake of reducing scope, we should keep this focused on identifiers that have already generated interest.

Thoughts?

daywiss (approved):

claytonroche (approved):

PR #80, opened 2020-11-14
UMIP - Add BTCDOM and ALTDOM as price identifiers

jaxrtech:
Hey UMA! Here is our draft UMIP for adding BTCDOM and ALTDOM as price identifiers for coin market dominance. Let us know your thoughts and feedback on the draft. Thanks!

### DIFF
--- a/UMIPs/umip-draft_btcdom_altdom.md
+++ b/UMIPs/umip-draft_btcdom_altdom.md
@@ -1,9 +1,10 @@
 ## Headers
  - UMIP <#>
  - Title: Add BTCDOM and ALTDOM as price identifiers
- - Author: Domination Finance
+ - Author: Domination Finance Team
  - Status: Pending approval
  - Created: November 14, 2020
+
 
 ## Summary (2-5 sentences)
 The DVM should support price requests for a Bitcoin and altcoin dominance index. Upon redemption, BTCDOM + ALTDOM = 100.00.
--- a/UMIPs/umip-draft_btcdom_altdom.md
+++ b/UMIPs/umip-draft_btcdom_altdom.md
@@ -1,0 +1,166 @@
+## Headers
+ - UMIP <#>
+ - Title: Add BTCDOM and ALTDOM as price identifiers
+ - Author: Domination Finance
+ - Status: Pending approval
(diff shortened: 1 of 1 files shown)


UMIPs/umip-draft_btcdom_altdom.md
```diff
@@ -0,0 +1,167 @@
...
+ - “name” - giving the colloquial ticker symbol,
+ - “id” - CoinGecko’s unique identifier for the coin,
+ - “market_cap_usd” - the total market capitalization in USD,
+ - “dominance_percentage” - the market dominance percentage as a decimal number from 0 to 100
+ - “timestamp” - the Unix timestamp when this data was generated from CoinGecko
+ 
+### Historical Data API
+Dispute bots and voters will need to query the cached historical data, dating back 96 hours. CoinGecko solely provides live data in (1-3) minute intervals, hence a custom solution was needed to ensure accurate historical data for a sufficient enough time frame. The reference implementation of the custom historical data caching solution will be made open-source to allow any stakeholder to host their own copy and provide a base for stakeholder who may choose to implement their own.
+
+The historical data custom solution is available at the following URL:
```
nicholaspai, line 108: Hey team, this an awesome submission! Some thoughts:

I think this historical data source should be a **suggested**, not prescribed, since the community is free to choose which historical data source they want to use. As an example look how the [UMIP for BTCUSD](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-7.md#implementation) is described:

> Here is a reference implementation for an offchain price feed based on the CryptoWatch API. **This feed should be used as a convenient way to query the price in realtime, but should not be used as a canonical source of truth for voters.** Users are encouraged to build their own offchain price feeds that depend on other sources.

Additionally I think it will be viewed favorably by the community if you empower others to access the source code (and verify that the api.dominiation.finance server is valid). Are you able to link the source code URL here?

@mrice32 @smb2796 @claytonroche

smb2796:
This lgtm! I think we can link the source code at a later date, since this is all based upon available live data and voters would be able to build their own solutions (even if they didn't immediately have a reference).

smb2796 (approved):

PR #81, opened 2020-11-15
Adding UMIP draft for GASETH twap price id

smb2796:


UMIPs/umip-22.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Technical Specification
+
+The definition of this identifier should be:
+- Identifier name: GASETH-TWAP-1Mx1M
+- Base Currency: ETH
+- Quote Currency: GAS
+- Sources: any Ethereum full node or data set of Ethereum node data. On-chain Ether/uGAS-JAN21 Uniswap pool. 
+- Result Processing: multiply by a million when calculating aggregatory gas prices.
+- Input Processing: See the UMIP-16 Implementation Section. Additionally, if the contract using this price identifier is an expiring contract, inputs will change depending on the price request timestamp in comparison to the expiry timestamp.
```
mrice32, line 31: Don't we need to define the TWAP and how it's computed somewhere in here?

UMIPs/umip-22.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Security considerations
+
+Security considerations pertaining to calculating an aggregate gas price are covered in the Security Considerations section of [UMIP-16](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-16.md).
+
+There are two general security considerations for using a self-referential token price for collateralization.
+
+1. **Token price manipulation** - Under illiquid market conditions, an attacker could attempt to drive prices down to withdraw more collateral than normally allowed or drive prices up to trigger liquidations.  However, it is important to note that almost all attacks that have been performed on DeFi projects are executed with flash loans, which allows the attacker to obtain leverage and instantaneously manipulate a price and extract collateral.  Collateralization based off of a TWAP should make these attacks ineffective and would require attackers to use significantly more capital and take more risk to exploit any vulnerabilities.
```
mrice32, line 62: One thing to note is that the TWAP is measured based on the price at the end of each block. This means that flash loans have _no_ effect when used to move a market up and then trade it back!

mrice32 (commented):
Looks great -- two minor comments.

UMIPs/umip-22.md
smb2796: Great point! Added a comment in about this, as I think it is important to note

mrice32 (approved):

PR #84, opened 2020-11-17
Sean/fixing nits on gas price umip

smb2796:


UMIPs/umip-22.md
```diff
@@ -11,14 +11,14 @@ This UMIP will reference a synthetic token to be created with this price identif
 
 The DVM should support requests for a price that resolves to either the median monthly Ethereum gas price or a 2-hour Time-Weighted Average Price (TWAP) on the highest volume Uniswap ETH/uGAS pool. The price resolution method to use will depend on the the timestamp the price request was made at.
 
-For a price request made at or after the Unix timestamp `1612051200` (Jan 31, 2021 00:00:00 UTC), the price will be resolved with the median monthly gas price calculation defined for GASETH-1M-1M in UMIP-20.
+For a price request made at or after the Unix timestamp `1612137600` (Feb 1, 2021 00:00:00 UTC), the price will be resolved with the median monthly gas price calculation defined for GASETH-1M-1M in UMIP-20.
```
mrice32, line 14: Where does this fall in the commit-reveal cycle? Should this be 10PM UTC so we can send the price request before the commit period starts?

smb2796: Feb 1, 00:00:00 UTC is the start of a reveal day. The thought is that the contract would expire on the first second of the reveal day. Does this work, or am I missing something?

mrice32: yeah, I think that's fine. We've used 10PM before the commit period before to help reduce the capital lockup period, but it's totally up to you.

mrice32 (dismissed):
LGTM!

mrice32 (approved):

PR #85, opened 2020-11-17
UMIP-21: Add link to historical api reference implementation 

jaxrtech:
Minor addition to UMIP-21 adding a link to the open-source reference implementation for the CoinGecko API historical caching solution.
Commit history got a bit missed up but only this one paragraph change is of significance.

@smb2796


smb2796 (approved):
LGTM!

PR #86, opened 2020-11-23
Added draft of TVL UMIP 

bryanjcampbell1:
Request for price ids

PR #87, opened 2020-11-26
Draft for addition of BCHNBTC as price identifier

chicfilabae:
Discussed with a few members of the team informally, mostly Clayton and Matt. 

I see that #86 is out and if lands before this, am happy to modify this to UMIP 24, but keeping it for now as 23 since there is no strict dependency between these two UMIPs and not sure what the protocol is for which one lands first in multiple are out simultaneously. Further, this UMIP's price identifier addition is rather straightforward to that of #86's.


chicfilabae:
@claytonroche

UMIPs/umip-23.md
```diff
@@ -0,0 +1,62 @@
...
+The definition of this identifier should be:
+ 
+- Identifier name: BCHNBTC
+- Base Currency: BCHN
+- Quote Currency: BTC
+- Data Sources: Binance, Coinbase Pro, Huobi
+- Result Processing: Median of BCH/BTC
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus. Granted, these are two of the most liquid markets in all of the CeFi crypto-verse.
+- Price Steps: 0.000001 (6 decimals in more general trading format)
+- Rounding: Closest, 0.000001 up
```
claytonroche, line 32: suggest changing "Price Steps" and "Rounding" to:

Decimals: 6
Rounding: Yes

chicfilabae: Done.

UMIPs/umip-23.md
```diff
@@ -0,0 +1,62 @@
...
+BCHBTC uses Binance, Coinbase Pro, and Huobi for BCH/BTC price information. The objective was to construct a portfolio of liquid price feeds that maximized geographical diversity as well as user base with negatively/zero correlated downtimes. Binance and Huobi are widely regarded as the most liquid and highest ADV exchanges in the world, and have been so for quite some time. Further, each of the exchanges has "crypto SOTA" APIs with free REST + WSS functionality with generous rate limits and robust endpoints for prices that any user may hit, both real time and historically speaking. Further, Coinbase Pro has a fairly orthogonal user base to the above exchanges and serves as an additive signal to a notion of "fair value" of BCH/BTC, along with a user friendly API. All three of these exchanges have 24/7 customer support + devops team ensuring highly available programmatic access and have strong redundancy across a variety of geographies. Lastly, because Binance and Huobi also have the most liquid derivatives market for both BCH and BTC pairs, users on their platforms should be able to arbitrage the BCHN uma synths if they so please, thereby increasing popularity of these synthetics.
```
claytonroche, line 40: The title is for BCHN but then are getting the BCH price - I just personally don't follow because I don't know much about BCHN.

chicfilabae: Great question, so we wanted to differentiate between the two forks, which are known widely as BCHN and BCHA. Most exchanges, namely the three we listed there for constructing the price, only view BCHN as the "real" fork, so their convention of BCH is just referring to BCHN.

chicfilabae: Do let us known if there's a better way we can disambiguate amongst them.

claytonroche: Great!  Thanks for your reply.  What about including that very text you just provided in your explanation?

I also wonder if most users just assume "BCH = BCHN" and thus maybe it's more clear to make the title of the price ID "BCHBTC" and then just specify in the umip that BCH = BCHN in this case?

claytonroche: On the other hand, the product you launch doesn't need to mention BCHN in the title, so perhaps leaving it as-is but with a comment explaining like you did to me would be best :)

UMIPs/umip-23.md
```diff
@@ -0,0 +1,62 @@
...
+ 
+## Implementation
+ 
+The value of BCHBTC for a given timestamp can be determined with the following process.
+ 
+1. BCH/BTC should be queried for from Binance, Huobi, and Coinbase Pro for the given timestamp rounded to the nearest second. The results of these queries should be kept at the level of precision they are returned at.
+2. The median of the BCH/BTC results should then be taken.
+3. This result should be rounded to six decimal places.
```
smb2796, line 50: Same as decimal amount comment above.

UMIPs/umip-23.md
```diff
@@ -0,0 +1,62 @@
...
...regimes of crypto. Given this, users are able to get by with thinner collateral ratio requirements given that they both tend to trend to together. In other words, more formally, the relative volatility of BCH/USD exceeds that of BCH/BTC by a stat sig amount given the positive correlation between them. Lastly, the use of BCH/USD with renBTC as collateral would have resulted in a quanto effect where there would be embedded convexity (correlation conditional gamma) that users would have to either hedge out or be forced to take a view on. Given the only liquid place to hedge that is on Bitmex, and due to their recent legal troubles, we decided it would be inconvenient to impose that on users. Further, while we are no strangers to quanto derivatives, the core thesis here isn't to productize BCH-BTC correlation. Lastly, the choice of renBTC over WBTC was due to the incremental decentralization that renBTC offers compared to WBTC due to the centralized intermediation involved. renBTC is a liquid market in the ethereum ecosystem and is prevalent across the major DeFi dapps. Further, in the time that it has been live, there have been no major security breaches or exploits of the core renVM.
```
smb2796, line 42: It's perfectly fine to specify that you are intending to use this price identifier with renBTC, but the price identifier itself can't really be restricted to any specific collateral unless it is explicit in the price id definition. Unless the actual quote currency is renBTC instead of BTC, it's my opinion that the language on making a decision on using renBTC over WBTC is slightly confusing. Thoughts on generalizing this?

smb2796 (commented):
Hey guys. Left a few nitpicky comments, but overall this looks great!

UMIPs/umip-23.md
smb2796, line 42: It's perfectly fine to specify that you are intending to use this price identifier with renBTC, but the price identifier itself can't really be restricted to any specific collateral unless it is explicit in the price id definition. Unless the actual quote currency is renBTC instead of BTC, it's my opinion that the language on making a decision on using renBTC over WBTC is slightly confusing for price reporters. Thoughts on generalizing this?

UMIPs/umip-23.md
```diff
@@ -0,0 +1,62 @@
...
+## Technical Specification
+The definition of this identifier should be:
+ 
+- Identifier name: BCHNBTC
+- Base Currency: BCHN
+- Quote Currency: BTC
+- Data Sources: Binance, Coinbase Pro, Huobi
+- Result Processing: Median of BCH/BTC
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus. Granted, these are two of the most liquid markets in all of the CeFi crypto-verse.
+- Decimals: 6
```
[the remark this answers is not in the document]
nicholaspai: Agreed. @chicfilabae I'd recommend following the format for the [USDBTC identifier](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-7.md) which specifies to what precision voters should input their prices with.

UMIPs/umip-23.md
```diff
@@ -0,0 +1,62 @@
...
+ 
+## Technical Specification
+The definition of this identifier should be:
+ 
+- Identifier name: BCHNBTC
+- Base Currency: BCHN
+- Quote Currency: BTC
+- Data Sources: Binance, Coinbase Pro, Huobi
+- Result Processing: Median of BCH/BTC
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus. Granted, these are two of the most liquid markets in all of the CeFi crypto-verse.
```
nicholaspai, line 30: What is the `Input Processing` section intended for? Have we used this in prior UMIPS? @claytonroche @smb2796 . 

```suggestion
- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
```

UMIPs/umip-23.md
```diff
@@ -0,0 +1,62 @@
...
+There is little cost associated with adding this price identifier, as there are multiple free and easily accessible data sources available.
+ 
+## Technical Specification
+The definition of this identifier should be:
+ 
+- Identifier name: BCHNBTC
+- Base Currency: BCHN
+- Quote Currency: BTC
+- Data Sources: Binance, Coinbase Pro, Huobi
+- Result Processing: Median of BCH/BTC
```
nicholaspai, line 29: @chicfilabae Can you please explain how BCHN-BTC would differ from BCH-BTC? Without knowing anything about the recent hard fork, it seems that BCH and BCHN are now two separate products. 

It's a bit strange to me that this is a BCHN-BTC identifier but it tracks the BCH token.

UMIPs/umip-23.md
nicholaspai: > Most exchanges, namely the three we listed there for constructing the price, only view BCHN as the "real" fork, so their convention of BCH is just referring to BCHN.

It would be nice to add proof that Coinbase, Binance, and Huobi take this official stance somewhere in the text of the UMIP. Perhaps a link to their official stance or a copy of their text?

UMIPs/umip-23.md
nicholaspai: Ah I see you have discussed this already [here](https://github.com/UMAprotocol/UMIPs/pull/87/files#r532316074)

UMIPs/umip-23.md
nicholaspai, line 42: > In other words, more formally, the relative volatility of BCH/USD exceeds that of BCH/BTC by a stat sig amount given the positive correlation between them

Can you expand on this statement a bit? What has been the historical correlation of BCH to BTC?

UMIPs/umip-23.md
smb2796: It is intended to be used to specify why voters would change inputs for the price identifier calculation - an example would be the GASETH price id, where the gas price could be queried over a range of time OR a range of blocks. Usually this is left with only the "Human Intervention" clause though.

UMIPs/umip-23.md
```diff
@@ -0,0 +1,62 @@
...
+ 
+## Implementation
+ 
+The value of BCHBTC for a given timestamp can be determined with the following process.
+ 
+1. BCH/BTC should be queried for from Binance, Huobi, and Coinbase Pro for the given timestamp rounded to the nearest second. The results of these queries should be kept at the level of precision they are returned at.
+2. The median of the BCH/BTC results should then be taken.
+3. This result should be rounded to six decimal places.
+ 
+Additionally, [CryptoWatch API](https://docs.cryptowat.ch/rest-api/) is a useful reference. This feed should be used as a convenient way to query the price in realtime, but should not be used as a canonical source of truth for voters. Users are encouraged to build their own offchain price feeds that depend on other sources.
```
daywiss, line 52: I dont understand this part:

> Users are encouraged to build their own offchain price feeds that depend on other sources.

Why is this encouraged if the canonical price feeds are already established?

daywiss (commented):
Looks good to me, and the background info is interesting.

UMIPs/umip-23.md
```diff
@@ -0,0 +1,62 @@
...
+The definition of this identifier should be:
+ 
+- Identifier name: BCHNBTC
+- Base Currency: BCHN
+- Quote Currency: BTC
+- Data Sources: Binance, Coinbase Pro, Huobi
+- Result Processing: Median of BCH/BTC
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus. Granted, these are two of the most liquid markets in all of the CeFi crypto-verse.
+- Decimals: 6
+- Rounding: Yes
```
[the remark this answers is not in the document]
nicholaspai: Great catch @daywiss . The USDBTC identifier specifies [how rounding works](Rounding: Closest, 0.5 up) as "Closest, 0.5 up" and I see no reason why this UMIP shouldn't follow the same.

UMIPs/umip-23.md
nicholaspai: For redundancy, this is language used in [other UMIP's](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-7.md#implementation). I think specifically the source that voters should develop redundancies for is CryptoWatch; one alternative would be querying Coinbase, Huobi and Binance API's directly.

UMIPs/umip-23.md
chicfilabae: All good points, @smb2796 yes that was the intent behind keeping it as BCHN/BTC. @nicholaspai added in some links to exchanges detailing their view on the BCH fork. @claytonroche added in the explanation.

UMIPs/umip-23.md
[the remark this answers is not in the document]
chicfilabae: Modified to 8.

UMIPs/umip-23.md
chicfilabae: Modified to 8.

UMIPs/umip-23.md
chicfilabae: Yes, good point. No need to introduce the renBTC language at this layer- removed it.

UMIPs/umip-23.md
chicfilabae: Ah, I see.

UMIPs/umip-23.md
chicfilabae: Yes, the 1 year historical pearson correlation of BCH to BTC has been approximately 0.75.

chicfilabae: Included this in for voter's reference as well

UMIPs/umip-23.md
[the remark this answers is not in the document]
chicfilabae: image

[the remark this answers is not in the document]
smb2796: @claytonroche and I were having some discussions on how the rounding worked and were trying something out. I think you guys are right though that being more explicit is better. @chicfilabae apologies for the back and forth here. I think we just go back to this convention.

```suggestion
- Rounding: Closest, 0.5 up
```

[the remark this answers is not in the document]
chicfilabae: I originally had it as more specific, but I believe Clayton suggested to modify it. Do you guys think it makes sense to revert? @nicholaspai @daywiss 

I personally prefer the more specific language for rounding rather than "Yes"

smb2796 (approved):
@chicfilabae thanks for responding to those comments! I changed the rounding specification back again, and the UMIP looks great to me!

PR #92, opened 2020-12-11
Draft for addition of STABLESPREAD as price identifier

chicfilabae:
Draft for addition of STABLESPREAD as price identifier. Briefly discussed with Matt offline.


UMIPs/umip-24.md
```diff
@@ -0,0 +1,64 @@
...
+Further, we also believe STABLESPREAD can act as a defi credit spread primitive where folks can build other products on top, such as an insurance based product where the premium of the insurance is tied to the divergence between the two baskets of stablecoins. Additionaly, one can view the probability of default of a stablecoin as `1 - min(price, 1)`. Given that each basket of stablecoins is a convex combination, then it is also a convex combination of default probabilities. STABLESPREAD is then a difference of convex combination of default probabilities. We also believe that a CDS (credit default swap) can be built on top of this since they are effectively a function of default probabilities. 
+ 
+There is little cost associated with adding this price identifier, as there are multiple free and easily accessible data sources available.
+ 
+## Technical Specification
+The definition of this identifier should be:
+ 
+- Identifier name: STABLESPREAD
+- Base Currency: STABLESPREAD
+- Quote Currency: USDC
```
claytonroche, line 28: @smb2796 I'd like your feedback on this:

I'm not really sure how useful a quote currency is to state here when it's not a normal index -- I think the Base/Quote thing makes sense for an index like ETH/BTC but less so for a 'counter' or 'ratio' type product.  I think when you say "quote currency" you are meaning the collateral probably, but that's not determined in a price ID umip.

UMIPs/umip-24.md
```diff
@@ -0,0 +1,64 @@
...
+
+The objective for STABLESPREAD was to construct a portfolio of the most liquid and active non-Ethereum based stablecoins, and rather than using some complicated weighting scheme, we are initially imposing uniform weighting. If it turns out there is some user feedback around doing a market-cap weighting, ADV-weighting, etc. that can be future work for us. On the MUSD side, we were deciding between that and DUSD as both are convex combinations of liquid stablecoins but the latter is not liquid or active compared to the former. We also experimented with a variety of functional forms of computing a divergence of these two baskets to represent a credit spread. Computing the ratio of them had some promising properties, namely non-negativity constraints, but the nonlinearity of it had some subpar properties that might cause unfriendly UX as liquidations might cascade if the denominator went down X% as opposed to the numerator. So, we settled on an affine transformation, where if both baskets are indeed the same then the result will be 1.
```
claytonroche, line 42: Do you mean to say "went down X%"?  or is "x" a placeholder for a percentage you intend to add?

UMIPs/umip-24.md
chicfilabae: Yes, you are correct. I thought it was odd to add there too since it's not a canonical pair. I'm fine with removing

UMIPs/umip-24.md
chicfilabae: Ah yes, I was just saying X% as an arbitrary % to make the point that (any) percentage decrease in the denominator vs. numerator affects the ratio in an asymmetric way.

UMIPs/umip-24.md
smb2796: I agree. For the same reason I don't think it's necessary to state base currency either - but not really a negative either.

smb2796 (commented):
Awesome proposal - left a couple of comments, but overall just really impressed with the level of thought put into this UMIP.

Compared to most of our existing identifiers, the use of STABLESPREAD is relatively complex. I'm personally finding it somewhat difficult to wrap my head around what the price would look like in different scenarios. I think you do a great job at explaining this in the motivation section, but it perhaps it might be useful to include a couple of scenarios w/ example calculations that represent what would happen in the case of basket divergence.

Also - even though this data is easily accessible, the price feed to calculate this is not super straightforward. Because contract launch is dependent on having a working price feed and the process to build this one is not immediately obvious, my instinct is that it would be useful to include a reference price feed for bot operators before actually voting on this. Thoughts on this?

UMIPs/umip-24.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Implementation
+ 
+The value of STABLESPREAD for a given timestamp can be determined with the following process.
+ 
+1. UST, CUSD, BUSD, MUSD should be queried for from the exchanges listed in the "Technical Specification" section for the given timestamp rounded to the nearest second. The results of these queries should be kept at the level of precision they are returned at.
+2. For each one, the average of the prices should be calculated.
+3. Then, calculate 1/3 * UST + 1/3 * CUSD + 1/3 * BUSD, denote this as `A`
+4. Perform A - MUSD + 1
+5. This result should be rounded to eight decimal places.
```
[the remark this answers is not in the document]
chicfilabae: Good catch- I added in the `max` afterwards to guarantee non-negative prices and forgot to update this section

UMIPs/umip-24.md
```diff
@@ -0,0 +1,63 @@
...
+Further, we also believe STABLESPREAD can act as a defi credit spread primitive where folks can build other products on top, such as an insurance based product where the premium of the insurance is tied to the divergence between the two baskets of stablecoins. Additionaly, one can view the probability of default of a stablecoin as `1 - min(price, 1)`. Given that each basket of stablecoins is a convex combination, then it is also a convex combination of default probabilities. STABLESPREAD is then a difference of convex combination of default probabilities. We also believe that a CDS (credit default swap) can be built on top of this since they are effectively a function of default probabilities. 
+ 
+There is little cost associated with adding this price identifier, as there are multiple free and easily accessible data sources available.
+ 
+## Technical Specification
+The definition of this identifier should be:
+ 
+- Identifier name: STABLESPREAD
+- Base Currency: STABLESPREAD
+- Data Sources: {Bittrex, Uniswap V2} for UST, {Binance, Uniswap V2} for BUSD, {Bittrex, OKCoin} for Celo Dollar, {Balancer, Uniswap V2} for MUSD
```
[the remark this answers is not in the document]
chicfilabae: Yes, that'd be wise to do so. Will do it now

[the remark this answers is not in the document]
chicfilabae: Done

chicfilabae:
@smb2796 Just added in a few commits to address some of the feedback that was brought up.

UMIPs/umip-24.md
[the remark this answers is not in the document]
smb2796: Looks great, thanks for doing this.

UMIPs/umip-26.md
```diff
@@ -0,0 +1,66 @@
...
+
+Further, we also believe STABLESPREAD can act as a defi credit spread primitive where folks can build other products on top, such as an insurance based product where the premium of the insurance is tied to the divergence between the two baskets of stablecoins. Additionaly, one can view the probability of default of a stablecoin as `1 - min(price, 1)`. Given that each basket of stablecoins is a convex combination, then it is also a convex combination of default probabilities. STABLESPREAD is then a difference of convex combination of default probabilities. We also believe that a CDS (credit default swap) can be built on top of this since they are effectively a function of default probabilities, and once STABLESPREAD gets whitelisted as a price identifier, we will start development on such products.
```
nicholaspai, line 19: Building a CDS on top of this non-ethereum <--> ethereum stablecoin spread is an awesome idea. OOC, is this implying that you believe the Ethereum stablecoin ecosystem to be the most risk-free stablecoin?

UMIPs/umip-26.md
```diff
@@ -0,0 +1,66 @@
...
+| UMIP Title | Add STABLESPREAD as a price identifier              |
+| Authors    | Bae (bae@youmychicfila.com), K (k@youmychicfila.com) |
+| Status     | Draft                                                                                                                                    |
+| Created    | December 12, 2020                                                                                                                           |
+ 
+## Summary (2-5 sentences)
+The DVM should support price requests for the STABLESPREAD price index. STABLESPREAD is defined as: `max(A - B + 1, 0)`, where `A` refers to an equally weighted basket of {UST, CUSD, BUSD}. B is `MUSD`. UST is TerraUSD, the interchain stablecoin connected by the Cosmos IBC. CUSD is Celo Dollar, the stablecoin on the Celo network. BUSD is Binance USD, a stablecoin issued by Paxos in partnership with Binance. `MUSD` is an autonomous and non-custodial stablecoin on mStable, representing a basket of stablecoins on Ethereum, namely USDT, USDC, TUSD, and DAI.
```
nicholaspai, line 10: What kind of additional risk does MUSD add in addition to the risk of its constituent stablecoins?

UMIPs/umip-26.md
```diff
@@ -0,0 +1,66 @@
...
+
+The objective for STABLESPREAD was to construct a portfolio of the most liquid and active non-Ethereum based stablecoins, and rather than using some complicated weighting scheme, we are initially imposing uniform weighting. If it turns out there is some user feedback around doing a market-cap weighting, ADV-weighting, etc. that can be future work for us. On the MUSD side, we were deciding between that and DUSD as both are convex combinations of liquid stablecoins but the latter is not liquid or active compared to the former. We also experimented with a variety of functional forms of computing a divergence of these two baskets to represent a credit spread. Computing the ratio of them had some promising properties, namely non-negativity constraints, but the nonlinearity of it had some subpar properties that might cause unfriendly UX as liquidations might cascade if the denominator went down X% as opposed to the numerator. So, we settled on an affine transformation, where if both baskets are indeed the same then the result will be 1.
```
nicholaspai, line 43: As an alternative to MUSD, why not do a uniform weighted basket of USDC, USDT, TUSD, and DAI?

nicholaspai, line 43: I agree we should build in flexibility to change the weightings, but this can be done via a future UMIP

UMIPs/umip-26.md
```diff
@@ -0,0 +1,66 @@
+## Headers
+| UMIP-26     |                                                                                                                                          |
```
nicholaspai, line 2: You're going to want to change this UMIP-28 or whatever the next available number us

nicholaspai (commented):
This identifier has me fired up thinking about the CDS curves we can build with it.

Left some questions/comments but overall looks 95% of the way there

UMIPs/umip-26.md
chicfilabae: Yes good question, we looked at the various tradeoffs between us specifying a basket of stablecoins explicitly as you indicate, but the incremental upside there was much lower. As in, we believe it is not as contentious what the set of stablecoins that best maintain their peg on Ethereum is compared non-Ethereum based stablecoins. Hence, we wanted to explicitly specify the latter and abstract away the former by using a basket-stablecoin product, such as mUSD. We also believed there were better priors for the Ethereum based stablecoin basket than a uniform weighting but didn't want to impose this ourselves. For mUSD, https://docs.mstable.org/mstable-assets/massets/minting-and-redemption/rebalancing, ensures that a wider community is incentivized to rebalance the basket as necessary. We also noticed that mUSD also had wider user adoption, etc. compared to the alternatives (DUSD, etc.). 

Lastly, as you mention below, by using an abstraction of mUSD, this allows for dynamic weights + constituents without us having to re-UMIP down the line.

UMIPs/umip-26.md
chicfilabae: Great question, the biggest set of risks that we isolated was primarily governance: https://docs.mstable.org/mstable-assets/massets/minting-and-redemption/rebalancing as well as mStable contract risk. However, based on what we've seen so far at https://consensys.net/diligence/audits/2020/07/mstable-1.1/, and general progress of the community since inception, these passed our threshold compared to newer projects such as DUSD.

UMIPs/umip-26.md
nicholaspai: Makes sense. mUSD's low circulation (looking at ~$30mn) is a bit concerning but I suppose if mUSD's liquidity breached a lower threshold, we could propose a UMIP to change the "ethereum stablecoin" baseline.

UMIPs/umip-26.md
nicholaspai: The reassuring aspect of how the DVM works is that if we ever deem mUSD to be unsuitable as a baseline, we can always upgrade this UMIP. And if mUSD ever breaks, UMA voters will have the discretion to resolve a reasonable price.

UMIPs/umip-26.md
chicfilabae: From what we've seen so far, yes.

smb2796 (approved):
Thanks for making those updates! Going to merge in as a draft

UMIPs/umip-26.md
chicfilabae: Interesting, I hadn't completely absorbed the implications of not bringing price on-chain preemptive to expiry. Yes, the upgrading of this via an UMIP is quite nice

PR #98, opened 2020-12-19
Adding stablespread ref impl

chicfilabae:
Standalone script with minimal deps for computing the price of a STABLESPREAD index as detailed in UMIP #28


UMIPs/umip-28.md
```diff
@@ -58,6 +58,8 @@ Additionally, [CryptoWatch API](https://docs.cryptowat.ch/rest-api/) is a useful
 Voters are responsible for determining if the result of this process differs from broad market consensus. This is meant to be vague as $UMA tokenholders are responsible for defining broad market consensus.
  
 This is only a reference implementation, how one queries the exchanges should be varied and determined by the voter to ensure that there is no central point of failure.
+
+As an example of how one may go about cosntructing such a feed programatically, here is one that was put together as a standalone script with minimal deps: https://gist.github.com/chicfilabae/3dff8aef233e4b9e5315daa7a835867d. As always, double-check the implementation as well as include additional measures for reliability to ensure reasonable assumptions around when the graph or cryptowatch isn't available.
```
nicholaspai, line 62:
```suggestion
As an example of how one may go about constructing such a feed programmatically, here is one that was put together as a standalone script with minimal dependencies: https://gist.github.com/chicfilabae/3dff8aef233e4b9e5315daa7a835867d. As always, double-check the implementation as well as include additional measures for reliability to ensure reasonable assumptions around when the graph or Cryptowatch isn't available. 
```

nicholaspai (approved):
This is a great addition, left one nit. Another question on the `gist` specifically:

Isn't the script runnable without the CRYPTOWATCH_API_KEY? I recall their API having a free tier which would suffice for this script

nicholaspai (changes requested):
In step (5) in the `implementation` section, do you think its worth making the resulting calculation symmetric and returning the minimum of step (5) currently and 2?

```
5. Perform the maximum of the result of step 4 and 0. This is to ensure non-negativity. 
6. Perform the minimum of the result of step 5 and 2.
```

UMIPs/umip-28.md
chicfilabae: Fixed.

chicfilabae:
Made changes where appropriate. Leaving the implementation as is for the steps, the extra min'ing is modifying the intended behavior

nicholaspai (dismissed):
@chicfilabae Thanks for addressing the changes. Do you think its worth adding the extra min-ing though in a subsequent PR?

chicfilabae:
I completely misunderstood your suggestion- I thought you meant step 5 and and step 2, which didn't make sense to me.

But yes, we should absolutely min by `2` for symmetry given the order of A and B shouldn't matter- good catch!

smb2796 (approved):
Great, thanks for the updates! Going to merge in and mark the UMIP as 'Last Call'.

PR #99, opened 2020-12-21
Add AMPLUSD as price identifier

abg4:


UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,87 @@
...
+- The project has set up liquidity mining (Geyser programs) to incentivize liquidity on decentralized exchanges (Uniswap, Sushiswap, Balancer).
+- One of its centralized exchanges, Kucoin, suffered a hack in September. Since the hack, there have been moments where the Kucoin pairs have reflected a lower market price compared to other exchanges. This creates hesitation in using it as one of the main data sources even though it has a high percentage of reported centralized exchange volume.
+
+The option being proposed is to use the index price used by FTX, Bitfinex, and Gate.io for their perpetual futures contracts. This price is used for funding payments, liquidations, and takes multiple trading pairs into account. Using these sources will make manipulation difficult and all three of these exchanges have free price sources.
+
+- FTX: Referred to as index price. The FTX AMPL_PERP contract uses the average of AMPL prices on FTX USD, FTX USDT, Bitfinex BTC, Bitfinex USD (Index), Kucoin BTC, and Kucoin USDT.
```
nicholaspai, line 52: If FTX includes Bitfinex in its average calculation, then why use Bitfinex as one of the 3 price feeds?

nicholaspai, line 52: Are these the three most voluminous exchanges for trading AMPL? I think it'd be the most secure if we use the three highest volume exchanges.

Additionally, I think its worth including at least one DEX price feed for diversity

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,87 @@
...
+
+The chosen AMPL/USD endpoints are:
+
+| Endpoint                                                       | Field name of price |
+| -------------------------------------------------------------- | ------------------- | ---------------------------------------------- |
+| https://ftx.com/api/futures/AMPL-PERP                          | _index_             |
+| https://api.gateio.ws/api/v4/futures/usdt/contracts/AMPL_USDT  | _mark_price_        | \* mark_type = index                           |
+| https://api-pub.bitfinex.com/v2/status/deriv?keys=tAMPF0:USTF0 | _MARK_PRICE_        | \* MARK_PRICE based on the BFX Composite Index |
+
+This is only a reference implementation, how one queries the exchanges should be varied and determined by the voter to ensure that there is no central point of failure.
```
nicholaspai, line 79: These API calls work and do not need additional authentication, +1

UMIPs/umip-draft-AMPLUSD.md
nicholaspai: Screen Shot 2020-12-21 at 08 33 39

However this tables renders poorly FYI

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,87 @@
...
+
+While it's important for tokenholders to have redundancy in their sources, bots and users that interact with the system in real time need fast sources of price information. In these cases, it can be assumed that the exchange median is accurate enough.
+
+The chosen AMPL/USD endpoints are:
+
+| Endpoint                                                       | Field name of price |
+| -------------------------------------------------------------- | ------------------- | ---------------------------------------------- |
+| https://ftx.com/api/futures/AMPL-PERP                          | _index_             |
+| https://api.gateio.ws/api/v4/futures/usdt/contracts/AMPL_USDT  | _mark_price_        | \* mark_type = index                           |
+| https://api-pub.bitfinex.com/v2/status/deriv?keys=tAMPF0:USTF0 | _MARK_PRICE_        | \* MARK_PRICE based on the BFX Composite Index |
```
smb2796, line 77: I get the following when I hit the bitfinex endpoint. Does this work for others?
image

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,87 @@
...
+
+Other factors that are relevant to the volume and price for AMPL:
+
+- Compared to other ERC-20 tokens, there is a high level of complexity for centralized exchanges to integrate AMPL due to its daily changes in supply potentially limiting adoption.
+- The project has set up liquidity mining (Geyser programs) to incentivize liquidity on decentralized exchanges (Uniswap, Sushiswap, Balancer).
+- One of its centralized exchanges, Kucoin, suffered a hack in September. Since the hack, there have been moments where the Kucoin pairs have reflected a lower market price compared to other exchanges. This creates hesitation in using it as one of the main data sources even though it has a high percentage of reported centralized exchange volume.
+
+The option being proposed is to use the index price used by FTX, Bitfinex, and Gate.io for their perpetual futures contracts. This price is used for funding payments, liquidations, and takes multiple trading pairs into account. Using these sources will make manipulation difficult and all three of these exchanges have free price sources.
```
smb2796, line 50: Interesting to use perpetual futures prices in place of spot. I wonder if this price id should be renamed to account for this and also allow a possible AMPL spot/USD price id later.

smb2796 (commented):
Awesome UMIP! Left a few comments, but overall looks really well thought out.

UMIPs/umip-draft-AMPLUSD.md
abg4: I will switch the price identifiers so that there is no overlap between price feeds. The rationale for using the perpetual futures index price was to make it more difficult to manipulate as there is not a lot of AMPL centralized spot volume compared to DEX/Futures volume. See your comment below regarding implementing the Uniswap AMPL/ETH pair which will increase diversity and make this less of an issue.

I will fix the table with the other updates to ensure it renders in a better format.

abg4: Using coingecko.com / coinmarketcap.com, the three most voluminous exchanges by reported volume are Uniswap, KuCoin, and Balancer. 

1. **AMPL/ETH pair on Uniswap (v2)**

I agree with using the AMPL/ETH pair on Uniswap. The primary method according to their documentation is to use thegraph.com (https://thegraph.com/explorer/subgraph/uniswap/uniswap-v2). 

The below would pull the AMPL/ETH price (token0Price):
`{
 pair(id: "0xc5be99a02c6857f9eac67bbce58df5572498f40c"){
     token0 {
       id
       symbol
       name
       derivedETH
     }
     token1 {
       id
       symbol
       name
       derivedETH
     }
     token0Price
     token1Price
 }
}`

We could then multiply by the ETH price to get the AMPL price in USD:
`{
  bundle(id: "1" ) {
   ethPrice
 }
}`

**2. AMPL/USDT pair on KuCoin**

My rationale for not using the KuCoin pair was due to the hack and thinking the FTX index price would be more difficult to manipulate. I will replace it with the Kucoin AMPL/USDT pair shown below if you feel it is a viable source regardless of the hack:

https://api.kucoin.com/api/v1/market/orderbook/level1?symbol=AMPL-USDT  // price

3. **AMPL/USDC pair on Balancer**
I am having some difficulty finding a good API source for Balancer. The graph is reporting the token price at a much higher price than the actual spot price ($1.06 vs $0.82 at the moment). The Balancer website also shows a price that is significantly different (https://pools.balancer.exchange/#/pool/0x7860e28ebfb8ae052bfe279c07ac5d94c9cd2937/about).

This could be a temporary issue or maybe I am missing something here. Either way, having two decentralized exchanges dependent on the graph may not be a good decision. If there isn’t another source, I think we should go to the next exchange in terms of volume which would likely be either the AMPL/USDT pair on FTX or the Gate.io AMPL/USDT pair. 

If there is no objection, I will use the FTX AMPL/USDT endpoint below making the three sources Uniswap, KuCoin, and FTX unless someone presents a better source for Balancer.

https://ftx.com/api/markets/AMPL/USDT //price

UMIPs/umip-draft-AMPLUSD.md
abg4: Thank you for confirming the endpoints.

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,87 @@
+## Headers
+
+- UMIP <#>
+- UMIP title: Add AMPLUSD as a price identifier
```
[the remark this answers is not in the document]
abg4: Thank you for your comments! I will include a USDAMPL identifier as it wouldn’t be difficult to add and could be implemented in some unique ways.

UMIPs/umip-draft-AMPLUSD.md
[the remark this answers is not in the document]
abg4: Great question. I will look into how much price history is available and get back to you.

abg4: The Bitfinex endpoints were a little difficult to work with. I will take another look. However, we may end up not using Bitfinex if switching price identifiers based on the @nicholaspai comments above.

UMIPs/umip-draft-AMPLUSD.md
abg4: So the thinking behind the perpetual futures index prices was to use the index prices that the exchanges use to calculate their funding rates. From how I understand it, FTX averages theirs based on the following pairs: FTX USD, FTX USDT, Bitfinex BTC, Bitfinex USD*, Kucoin BTC, and Kucoin USDT. (https://help.ftx.com/hc/en-us/articles/360027668812-Index-Calculation)

My thinking was that it would be extremely difficult to manipulate this price feed compared to a single exchange. The issue is that it would make it difficult to then use another source without having any overlap. We would have to do something like the FTX index price, Uniswap, and then the best source that isn’t already covered is probably the Gate.io spot price for the AMPL/USDT pair.

Based on the comments above, I am planning on switching to the AMPL/ETH pair on Uniswap (v2), AMPL/USDT pair on KuCoin, and the AMPL/USDT pair on FTX. Let me know your thoughts.

UMIPs/umip-draft-AMPLUSD.md
chicfilabae: In my opinion, overlapping price feeds by themselves aren't necessarily a subpar construction- to answer ^ ^ one upside is for ensuring higher availability (i.e. if FTX were to go down but Bitfinex is up, etc.).

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,87 @@
...
+
+The DVM should support price requests for the AMPL/USD price index.
+
+## Motivation
+
+The DVM currently does not support the AMPL/USD price index.
+
+Supporting the AMPLUSD price identifier would enable the creation of a synthetic AMPLUSD token, backed by USD. Token minters could obtain short exposure on the AMPL/USD index, while token holders could go long or use synthetic AMPL for functional purposes.
+
+There are multiple practical uses and trading strategies for synthetic AMPL which include avoiding rebases, minting tokens for short exposure, and hedging underlying AMPL exposure. Background on Ampleforth and its rebase mechanism can be found in the project whitepaper: https://www.ampleforth.org/papers/
```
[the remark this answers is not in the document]
abg4: Great question regarding the discount. While AMPL hasn’t spent a lot of its history trading significantly above a dollar, there is no denying that when it does most traders would rather own AMPL for its inflation mechanism. There are a few aspects that may prevent the discount from being massive but I agree when the price is significantly over a dollar there will likely be a discount and under a dollar, there may be a premium.
- The supply does its best to pull the price towards $1. It can be argued how effective that actually has been but looking at funding rates and how people trade AMPL, many traders do believe the token will eventually reach $1 when it gets well above or below it. This could help in causing the synthetic to vary in be consistently trading at a discount or a premium.
- The perpetual futures contracts, especially FTX, are liquid enough that it would be a trading opportunity if synthetic AMPL actually was at a massive discount. Let’s say AMPL was at $2 and synthetic AMPL was at $1.50. I would think there are some traders out there (at least I would), that would buy short exposure of 100 AMPL on FTX at $2 and then buy 100 synthetic AMPL at $1.30 not believing $2 is sustainable and prices will eventually converge. While they wouldn't be receiving the AMPL inflation, some may believe it's a better risk-adjusted trade. This scenario obviously does not take liquidity, funding rates, etc into consideration.
- There is still no denying that when price does get well above a dollar most traders are still going to rather own the underlying AMPL for its inflation mechanism. Another way of combating this may be developer mining incentives. A transparent set of rules could be provided where a reserve is created when AMPL is in a certain price range. The reserve could then be used to incentivize purchasing/holding during times where AMPL is significantly above or maybe even below a dollar. Still thinking through some other options but any feedback or other ideas would be great to hear.

The purpose of an AMPL synthetic outside of avoiding rebase contract risk, which I agree is very little, could be:
- the ability to obtain long exposure to AMPL when it is below $1 without having your purchasing power deteriorate with the negative rebases.
- I just checked FTX and people are currently paying ~2.5% per day for funding. Very difficult to hold a long term position with these types of rates.
- act as a hedge against underlying AMPL exposure or high yield liquidity mining programs the coin offers called geysers. 
- AMPL is a trader's coin that could be used in various trading strategies similar to what’s mentioned above.

UMIPs/umip-draft-AMPLUSD.md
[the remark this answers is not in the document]
abg4: FTX has historical prices as you are able to specify a start time, end time, window length. Gate.io, KuCoin, and Bitfinex are the same way and I didn’t see restrictions for any on how far back a user is able to go, just restrictions on the number of queries and time frame parameters. For the thegraph.com API for Uniswap, I am able to find daily data but haven’t figured out a way to get smaller time frames like the centralized exchanges.

Is this what you were looking for? If not, let me know and I can look into more.

FTX: https://docs.ftx.com/#get-historical-index
Gate.io: https://www.gate.io/docs/apiv4/en/index.html#market-candlesticks
Bitfinex: https://docs.bitfinex.com/reference#rest-public-candles
KuCoin: https://docs.kucoin.com/#get-trade-histories

UMIPs/umip-draft-AMPLUSD.md
abg4: I agree some overlap isn’t necessarily a bad way to go…especially when it comes to an index price that incorporates several price sources. One thing I really like about the FTX index price is the transparency they have towards the sources they use and it is used for calculating the funding rates and liquidations. While I wasn’t able to find that same level of transparency with Bitfinex, I still believe it’s a better source than some of the spot prices that have little volume and order depth. I think incorporating the Uniswap pair is a great suggestion and will add a level of diversity…curious what your opinion is towards the best price sources to use?

smb2796: I think it's fair to exclude KuCoin in favor of Gate.io or Bitfinex for the reasons you stated. In favor of the inclusion of Uniswap though

UMIPs/umip-draft-AMPLUSD.md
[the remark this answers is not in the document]
smb2796: Yes this looks useful. Voters would really need to use these endpoints to determine the price rather than the current price. They should be included in the UMIP with directions on how to use each.

UMIPs/umip-draft-AMPLUSD.md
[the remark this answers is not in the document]
Tommy1231232: One other benefit would be tax advantages! I do not use ample because of the tax implications from rebasing.

UMIPs/umip-draft-AMPLUSD.md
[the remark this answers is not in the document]
nicholaspai: Wait @smb2796 why do we need both an AMPLUSD identifier and a USDAMPL one? Are they not redundant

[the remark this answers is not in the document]
chrismaree: @nicholaspai I think it's to simplify adding products that require the inverse so that we already have the registered identifier, should the product require it. otherwise, how would an EMP/perp be able to request 1/AMPUSD price?

smb2796 (commented):
Discussed some with @nicholaspai and @chrismaree. We think the best approach would be to use one DEX (Uniswap) together with FTX and either Bitfinex or Gate.io. The FTX/Bitfinex/Gateio are all available on Cryptowat.ch and Uniswap prices are easily queried, so they should all be readily accessible.

Additionally, I think the historical endpoints you listed should be included as a reference implementation.

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,87 @@
...
+
+The definition of this identifier should be:
+
+- Identifier name: AMPLUSD
+- Base Currency: AMPL
+- Quote Currency: USD
+- Data Sources: FTX, Bitfinex, Gate.io
+- Result Processing: Median
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+- Price Steps: 0.0001 (4 decimals in more general trading format)
```
smb2796, line 33: Is the intention to use this price identifier with USDC as a collateral? If so, this should be 6 decimals to match the precision of USDC

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,87 @@
...
+- Rounding: Closest, 0.5 up
+- Pricing Interval: 60 seconds
```
smb2796, line 35: This pricing interval should match what is specified in the Implementation section. This should really be one second if prices are queried with a granularity of 1 second.

abg4: updated the pricing interval to 1 second

UMIPs/umip-draft-AMPLUSD.md
abg4: Yes, the intention is to use the price identifier with USDC. Updated to 6 decimals to match the precision of USDC.

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,177 @@
...
+## Technical Specification
+
+The definition of this identifier should be:
+
+- Identifier name: AMPLUSD
+- Base Currency: AMPL
+- Quote Currency: USD
+- Data Sources: FTX, Uniswap, Gate.io
+- Result Processing: Median
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
```
smb2796, line 32: Since these two identifiers are denominated in different currencies and are both non-standard scaling currencies, we should be explicit about how many decimals these price identifiers should be scaled to.

To use USDC/USDT as collateral:

```suggestion
- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
- Collateral Decimals: 6
```

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,177 @@
...
+
+The definition of this identifier should be:
+
+- Identifier name: USDAMPL
+- Base Currency: USD
+- Quote Currency: AMPL
+- Data Sources: FTX, Uniswap, Gate.io
+- Result Processing: Median
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+- Price Steps: 0.000001 (6 decimals in more general trading format)
```
smb2796, line 46: Adding collateral decimal specification for AMPL. As a side note, to ever use this USDAMPL effectively, AMPL would need to be supported as a collateral type first.

```suggestion
- Price Steps: 0.000001 (6 decimals in more general trading format)
- Collateral Decimals: 9
```

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,177 @@
...
+                  derivedETH
+            }
+         }
+      }
+   }
+```
+where “`{hourStartUnix}`” is the requested historical hourly snapshot data as a Unix timestamp. The result needs to then be multiplied by the ETH price at the same time to obtain the historic price. See the implementation below as an example of how it can be calculated.
+
+As an example of how one might go about constructing a feed programmatically, here is a standalone script with minimal dependencies that calculate the current price and historical price:
+- AMPLUSD and USDAMPL historical price calculation: https://gist.github.com/abg4/ef494aeace3c779fc49f1f774da349ff
```
smb2796, line 161: Awesome job building reference implementations. This will also be able to be supported by our existing price feeds for bots.

UMIPs/umip-draft-AMPLUSD.md
```diff
@@ -0,0 +1,177 @@
+## Headers
+
+- UMIP <#>
```
smb2796, line 3: Can you also change the file name to `umip-33`?

```suggestion
- UMIP-33
```

smb2796 (commented):
Looking great - just a few outstanding comments and then I think we can merge into last call status

smb2796 (approved):
Thanks for making those updates. Going to merge in and change status to Last Call

PR #102, opened 2020-12-22
Add EURUSD, CHFUSD and GBPUSD as price identifiers

SayWhatDev:


UMIPs/umip-29.md
```diff
@@ -0,0 +1,69 @@
...
+| Created    | December 22, 2020                                        |
+
+## Summary
+The DVM should support price requests for the EUR/USD, GBP/USD and CHF/USD price index.
+
+## Motivation
+The DVM currently does not support the EURUSD, GBPUSD and CHFUSD index.
+
+### Cost: 
+While Forex data are free through open centralized exchange, brokers, and other sources APIs, the most reliable are paid. We propose to use TraderMade's APIs whose pricing is accessible on their website.
```
smb2796, line 16: Tradermade also has a free tier which should be sufficient for voters

UMIPs/umip-29.md
```diff
@@ -0,0 +1,69 @@
...
+
+Identifier name: CHF/USD-1e18
+Base Currency: CHF
+Quote Currency: USD
+
+Source: https://marketdata.tradermade.com
+Result Processing: Median
+Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+Price Steps: 0.0001 (4 decimals in more general trading format)
+Rounding: Closest, 0.0001 up
```
smb2796, line 43: We typically use a slightly convention. Want to start being more explicit with these since there is generally some confusion associated with rounding specifications. My suggestion is:

```suggestion
Rounding: Closest, 0.5 up (> .00005 rounds up, < .00005 rounds down)
```

UMIPs/umip-29.md
```diff
@@ -0,0 +1,69 @@
...
+
+While the price of EURUSD does not vary much across various price feed during the weekday, it may be more challenging to get a unified price feed on the weekend: first, while the underlying market is not closed, most of the broker and banks feeding the APIs price feed are; then, for the one providing a price feed, they often add a very big spread, which makes the price of forex pairs vary significantly by source; a way to mitigate it is to use the last price known. 
+
+A consequence to this is the existence of gaps between the last quote on Friday to the first one on Sunday; while such gaps rarely exceed +/-2%, which should not put in danger the collateralization ratio, an exceptional event happening the weekend may create a bigger gap and could, in theory, endanger the collateral ratio; yet, it is very unlikely to see +/-20% gap on EURUSD, but it is in theory possible. Nevertheless, due to gaps, a larger number of liquidation and disputes may arise on Sunday 9 pm UTC if the collateral ratio was already getting weaker before the week’s end.
```
smb2796, line 69: This is well thought out and explained. One additional point I would make is that contracts using this price identifier should have a reasonably high collateralization requirement. This is to combat the danger where a positionholder could be close to the collateralization ratio when entering the weekend, and then a large price gap causes them to be undercollateralized under a 100% ratio. This would mean that their position was no longer profitable to liquidate.

UMIPs/umip-29.md
```diff
@@ -0,0 +1,69 @@
+## Headers
+| UMIP-29    |                                                          |
+|------------|----------------------------------------------------------|
+| UMIP Title | Add EURUSD, CHFUSD and GBPUSD as price identifiers       |
```
smb2796, line 4: Any interest in also proposing the inverses of these?

UMIPs/umip-29.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Cost: 
+While Forex data are free through open centralized exchange, brokers, and other sources APIs, the most reliable are paid. We propose to use TraderMade's APIs whose pricing is accessible on their website.
+
+### Opportunity: 
+Synthetic tokens that track Forex pairs suche as EURUSD GBPUSD and CHFUSD could be used both for speculationt and hedging, but we see a bigger opportunity for using them as building blocks for helping other DeFi protocols and dApp addressing the european market.
+
+## Technical Specification
+The definition of this identifier should be:
+Identifier name: EUR/USD-1e18
```
smb2796, line 23: This is to indicate that the price identifier is returned as 18 decimals? I would suggest just naming these the same as you have in the `Motivation` section (EURUSD etc)

UMIPs/umip-29.md
```diff
@@ -0,0 +1,69 @@
...
+Quote Currency: USD
+
+AND
+
+Identifier name: CHF/USD-1e18
+Base Currency: CHF
+Quote Currency: USD
+
+Source: https://marketdata.tradermade.com
+Result Processing: Median
```
smb2796, line 40: Median does not apply as these prices are determined from the results of 1 data source

UMIPs/umip-29.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Rationale
+Apart from the weekend, there is little to no difference in prices on liquid major Forex pairs like EURUSD, so any price feed could be used; however, for convenience, we recommend using the one of TraderMade.
+
+## Implementation
+The value of this identifier for a given timestamp should be determined by querying for the price of EURUSD, GBPUSD and CHFUSD from TraderMade.com for that timestamp. More specifically, users can query “https://marketdata.tradermade.com/api/v1/minute_historical?currency=EURUSD&date_time=2020-11-11-13:01&api_key=apikey” and use the close price as reference. 
+
+Tradermade’s price feed is an aggregated feed from multiple Tier One and Two Banks, Market-Makers and Data Providers. They are popular with Quantitative Traders, Fintech companies and Institutional customers who require a high quality and trusted feed.
+
+Yet, theur REST API does not work when the Forex market is closed, eem to work on Sunday even though the market opens at 10 pm london time (I asked them). The API just returns the friday OHLC
```
smb2796, line 56: Can you clarify this sentence. What time do the forex markets close/open etc

UMIPs/umip-29.md
```diff
@@ -0,0 +1,69 @@
...
+
```
smb2796, line 57: I think some clarification on what tokenholders should do when a price request occurs on a weekend would be good. What I'm hearing is that they should use the Friday ohlc, but would be good to make this more explicit.

UMIPs/umip-29.md
```diff
@@ -0,0 +1,69 @@
...
+As most of the price feed does not provide a price for CHFUSD but USDCHF, the value of this identifier will undergo one additional step: CHFUSD = 1/USDCDF.  Token holders should take care to confirm that the order of the quote and base currency they refer to matches the one being requested by the DVM in the event of a price request.
+Ultimately, how one queries the exchanges should be varied and determined by the voter to ensure that there is no central point of failure.
```
smb2796, line 59:
```suggestion
TraderMade is provided as an accessible source to query for this data, but ultimately how one queries for these rates should be varied and determined by the voter to ensure that there is no central point of failure.
```

UMIPs/umip-29.md
```diff
@@ -0,0 +1,69 @@
...
+While it's important for tokenholders to have redundancy in their sources, bots and users that interact with the system in realtime need fast sources of price information. In these cases, it can be assumed that the exchange median is accurate enough
+
+Liquidation and dispute bots should have their own subscription to price feeds. Our price-feed provider’s API documentation can be found [here](https://marketdata.tradermade.com/documentation).
+
+## Security considerations
```
smb2796, line 64: There are also some security considerations around using a singular data source. Are there alternatives that voters could turn to in the case of a TraderMade outage?

smb2796 (commented):
Left some comments, but overall looks great!

UMIPs/umip-29.md
SayWhatDev: Usually voters can turn to basically any price feed out there, since the differences in price feeds between brokers for Forex pairs are almost insignificant.

UMIPs/umip-29.md
SayWhatDev: What would be the point of inverses in our case? @smb2796

UMIPs/umip-29.md
SayWhatDev: What is meant here by 0.5 @smb2796. Since a fraction on the Forex market is 0.0001 (depending on the currency pair), so if we are rounding at 0.5 it would cause an increase or decrease of 50%.

SayWhatDev:
Added some changes based on the inputs above in my last commit. @smb2796

UMIPs/umip-29.md
smb2796: It means that a five in the decimal place before your chosen amount of decimal places would round up. Or (> .00005 rounds up, < .00005 rounds down)

UMIPs/umip-29.md
nicholaspai: @smb2796 As mentioned in other UMIPs I don't see the point of adding inverse identifiers and unnecessarily adding to smart contract storage.

UMIPs/umip-29.md
```diff
@@ -0,0 +1,77 @@
...
+Quote Currency: USD
+
+AND
+
+Identifier name: CHFUSD
+Base Currency: CHF
+Quote Currency: USD
+
+Source: https://marketdata.tradermade.com
+Result Processing: Mediam (if more than 1 source is utillized), otherwise Exact Price
```
nicholaspai, line 40:
```suggestion
Result Processing: Median (if more than 1 source is utilized), otherwise Exact Price
```

UMIPs/umip-29.md
```diff
@@ -0,0 +1,77 @@
...
+Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+Price Steps: 0.0001 (4 decimals in more general trading format)
+Rounding: Closest, 0.0001 up
+Decimal: 18 (1e18)
+Pricing Interval: 60 seconds
+Dispute timestamp rounding: down
+
+## Rationale
+Apart from the weekend, there is little to no difference in prices on liquid major Forex pairs like EURUSD, so any price feed could be used; however, for convenience, we recommend using the one of TraderMade.
```
nicholaspai, line 49: Using one API source creates a problem for UMA voters if the API source goes down temporarily or does not provide faithful data for a given price request time stamp. I think we should propose at least 2, ideally 3+, data providers. 

It's not about the variance in Fx prices across data providers but about mitigating the chance that a corrupt or incompetent API provider can make resolution more challenging for UMA voters.

UMIPs/umip-29.md
```diff
@@ -0,0 +1,77 @@
...
+
+Tradermade’s price feed is an aggregated feed from multiple Tier One and Two Banks, Market-Makers and Data Providers. They are popular with Quantitative Traders, Fintech companies and Institutional customers who require a high quality and trusted feed.
+
+Over the weekend the REST API does not return any price, but we can request the price of a certain moment before the market close (as ex: the closing price of Friday).
+
+Due to unavailability of price feed for any financial asset over the weekend, tokenholders and users will be using the latest known price, before the market close on Friday, which esentially is the closing price of the asset from Friday. Same goes in a case of a liquidation process - the liquidator should use the last price from Friday in order to match with the price on which a synthetic asset was created, if it was created over the weekend. If not the closing price on Friday for a certain asset should be a navigating point in calculating the collateralization ratio of a position and in the liquidation process.
```
nicholaspai, line 58: So just to be clear, we're establishing a convention that if a request timestamp takes place on a weekend (or any day that the Fx market is closed), then we should use the last traded tick as the price? 

I think this is reasonable, but we should make this more clear in the UMIP itself. Perhaps in the "technical specification" section.

UMIPs/umip-29.md
```diff
@@ -0,0 +1,77 @@
...
+
+Forex markets are usually open throughout bussiness days from 10:00 PM UTC time on Sunday, until 9:00 PM UTC time on Friday. However it is good to note that some brokers might close their price feeds 5-10 minutes prior to the official market closing time, thus the latest price given from the broker will be used as a official price for the asset. We should also note here that during those minutes the price remains flat without any changes, however theoretically it is possible to have a stronger price movement.
```
nicholaspai, line 60:
```suggestion
Forex markets are usually open throughout business days from 10:00 PM UTC time on Sunday, until 9:00 PM UTC time on Friday. However it is good to note that some brokers might close their price feeds 5-10 minutes prior to the official market closing time, thus the latest price given from the broker will be used as a official price for the asset. We should also note here that during those minutes the price remains flat without any changes, however theoretically it is possible to have a stronger price movement.
```

UMIPs/umip-29.md
```diff
@@ -0,0 +1,77 @@
...
+
+As most of the price feed does not provide a price for CHFUSD but USDCHF, the value of this identifier will undergo one additional step: CHFUSD = 1/USDCDF.  Token holders should take care to confirm that the order of the quote and base currency they refer to matches the one being requested by the DVM in the event of a price request.
```
nicholaspai, line 62: Why don't we just add a USDCHF identifier instead of a CHFUSD to reduce steps for voters?

UMIPs/umip-29.md
nicholaspai: We should still probably officially list more than one data providers in this UMIP so that UMa voters don't unintentionally resolve to a corrupted data providers price.

UMIPs/umip-29.md
```diff
@@ -0,0 +1,77 @@
...
+
+In the past the CHF was pegged to the EUR, but in 2015 this pegged was removed by the Swiss Central Bank in a surprise move, which caused a huge volatile movement on the CHF price. Such movement could put the CHF below the 100% collateralization ratio, although it is highly unlikely to happen. This risk could appear again one day if the peg is reinstated. In the forementioned situation the CHF lost 20% within one day. However with a higher required collateralization ratio such events can be mitigated.
```
nicholaspai, line 77: Good color thanks!

nicholaspai (commented):
Really looking forward to adding these identifiers but I think we should consider adding more price resolution sources, and specific instructions in Technical Specifications about how to handle timestamps that fall on weekends.

UMIPs/umip-29.md
claytonroche: I think in the event of a big event (a 20%+ gap on EURUSD) we might see some human judgement, whereby the APIs fail to update but we see liquidators go ahead and liquidate positions.  I suppose that would cause disputes since the API's feeding dispute bots wouldn't know any different, and in that event the tokenholders would be asked to decide on the "market consensus" price -- And I think in the end, we'd arrive at the correct answer.

My tl;dr here is that it _appears_ as if the very very edge case would still have an appropriate (if somewhat messy) outcome.

UMIPs/umip-29.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Cost: 
+While Forex data are free through open centralized exchange, brokers, and other sources APIs, the most reliable are paid. We propose to use TraderMade's APIs whose pricing is accessible on their website. TraderMade has also a Free Tier which can be used by the voters and would be sufficient to provide the price of a certain asset.
+
+### Opportunity: 
+Synthetic tokens that track Forex pairs suche as EURUSD GBPUSD and CHFUSD could be used both for speculationt and hedging, but we see a bigger opportunity for using them as building blocks for helping other DeFi protocols and dApp addressing the european market.
+
+## Technical Specification
+The definition of this identifier should be:
+Identifier name: EURUSD
```
chrismaree, line 23: This formatting does not conform to the standard. Please make these bullet lists like other UMIPs. See https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-7.md#technical-specification for an example.

UMIPs/umip-29.md
chrismaree: Additionally, I think that the API used here should match that used in the `CNYUSD` & `USDCNY` UMIPs: https://github.com/UMAprotocol/UMIPs/pull/97/files#r549197519

Ideally, this will enable re-use of the price feeds with two sets of settings without needing custom implementations for each set.

UMIPs/umip-29.md
```diff
@@ -0,0 +1,77 @@
...
+Rounding: Closest, 0.0001 up
+Decimal: 18 (1e18)
+Pricing Interval: 60 seconds
+Dispute timestamp rounding: down
+
+## Rationale
+Apart from the weekend, there is little to no difference in prices on liquid major Forex pairs like EURUSD, so any price feed could be used; however, for convenience, we recommend using the one of TraderMade.
+
+## Implementation
+The value of this identifier for a given timestamp should be determined by querying for the price of EURUSD, GBPUSD and CHFUSD from TraderMade.com for that timestamp. More specifically, users can query “https://marketdata.tradermade.com/api/v1/minute_historical?currency=EURUSD&date_time=2020-11-11-13:01&api_key=apikey” and use the close price as reference.
```
chrismaree, line 52: Is there an option for polling this API that does not require an API key?

UMIPs/umip-29.md
SayWhatDev: @nicholaspai USDCHF price feed is not supported by price feed providers, since the original display of the currency pair is CHFUSD, thus voters will have to make the switch calculation.

nicholaspai: What do you mean by "As most of the price feed does not provide a price for CHFUSD but USDCHF, the value of this identifier will undergo one additional step: CHFUSD = 1/USDCDF" then?

Let's just use the identifier that requires fewer calculation steps for the voters, so if price feed providers only support CHFUSD then let's do that. Am I understanding correctly?

smb2796 (commented):
Looking good. There are still some outstanding small comments, and @nicholaspai brings up a good point about explicitly stating what to do when prices are unavailable due to Weekends/Holidays.

UMIPs/umip-29.md
Pascaltallarida: Hey, no it was to say that the jEUR token will have 18 dec.

UMIPs/umip-29.md
SayWhatDev: It is not possible to match the API provided in https://github.com/UMAprotocol/UMIPs/pull/97/files#r549197519 since the price feed updates for that API are not in real time. Tradermade however and the other providers I've outlined in my latest commit provide real-time price feed data.

UMIPs/umip-29.md
```diff
@@ -53,20 +58,62 @@ The value of this identifier for a given timestamp should be determined by query
...
 
+### Price feed - liquidations and disputes
+
 Liquidation and dispute bots should have their own subscription to price feeds. Our price-feed provider’s API documentation can be found [here](https://marketdata.tradermade.com/documentation).
 
 In the case of a TraderMade outage voters can turn to any other available price feed API or a broker API, as the price feeds for the forementioned financial assets does not differ much between different providers. There might be some slight differences, however they are quite insignificant and would not affect the liquidation or dispute processes. In that case we'll use a median price from the different price feeds provided by the voters.
 
+## Additional price feed providers
+
+### FXMarketAPI
```
smb2796, line 87: Thanks for adding this. Great alternative for voters

UMIPs/umip-29.md
```diff
@@ -53,20 +58,62 @@ The value of this identifier for a given timestamp should be determined by query
...
 
-Forex markets are usually open throughout bussiness days from 10:00 PM UTC time on Sunday, until 9:00 PM UTC time on Friday. However it is good to note that some brokers might close their price feeds 5-10 minutes prior to the official market closing time, thus the latest price given from the broker will be used as a official price for the asset. We should also note here that during those minutes the price remains flat without any changes, however theoretically it is possible to have a stronger price movement.
+If a request timestamp takes place on a weekend or any other day the Forex market is closed, voters should use the latest tick as the price. For the weekend that would be the closing price of the asset on Friday.
```
smb2796, line 67: Thank you for this clarification. Seems sufficient to me

UMIPs/umip-29.md
smb2796: Thoughts here?

UMIPs/umip-29.md
smb2796: I think Chris's comment was centered around that FX UMIPs should use the same price sources, not that you had to match to the other UMIP.

SayWhatDev:
Since we'll need 18 decimals for all identifiers and we've noticed that the XAUUSD identifier currently registered has less, I've added to this UMIP an identifier for XAUUSD and also noted that all identifiers should have 18 decimal places due to the way our system implements them

UMIPs/umip-29.md
```diff
@@ -0,0 +1,134 @@
...
+- Decimals: 18
+
+-----------------------------------------
+
+Source: https://marketdata.tradermade.com
+Result Processing: Median (if more than 1 source is utilized), otherwise Exact Price
+Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+Price Steps: 0.0001 (4 decimals in more general trading format)
+Rounding: Closest, 0.0001 up
+Decimal: 18 (1e18)
```
smb2796, line 59: To be clear, you want prices to be rounded to 4 decimal places, but still have prices returned with 18 decimal places?

UMIPs/umip-29.md
```diff
@@ -0,0 +1,134 @@
...
+
+Tradermade’s price feed is an aggregated feed from multiple Tier One and Two Banks, Market-Makers and Data Providers. They are popular with Quantitative Traders, Fintech companies and Institutional customers who require a high quality and trusted feed.
+
+### Weekend timestamp
+
+Over the weekend the REST API does not return any price, but we can request the price of a certain moment before the market close (as ex: the closing price of Friday).
+
+Due to unavailability of price feed for any financial asset over the weekend, tokenholders and users will be using the latest known price, before the market close on Friday, which esentially is the closing price of the asset from Friday. Same goes in a case of a liquidation process - the liquidator should use the last price from Friday in order to match with the price on which a synthetic asset was created, if it was created over the weekend. If not the closing price on Friday for a certain asset should be a navigating point in calculating the collateralization ratio of a position and in the liquidation process.
```
smb2796, line 75: Small nitpick, but more specifically a lack of availability for FX here. 

```suggestion
Due to unavailability of price feed for foreign exchange rates over the weekend, tokenholders and users will be using the latest known price, before the market close on Friday, which essentially is the closing price of the asset from Friday. Same goes in a case of a liquidation process - the liquidator should use the last price from Friday in order to match with the price on which a synthetic asset was created, if it was created over the weekend. If not the closing price on Friday for a certain asset should be a navigating point in calculating the collateralization ratio of a position and in the liquidation process.
```

UMIPs/umip-29.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Security considerations
+Adding this new identifier by itself poses a little security risk to the DVM or priceless financial contract users. However, anyone deploying a new priceless token contract referencing this identifier should take care to parameterize the contract appropriately to avoid the loss of funds for synthetic token holders. Additionally, the contract deployer should ensure that there is a network of liquidators and disputers ready to perform the services necessary to keep the contract solvent.
+
+While the price of EURUSD does not vary much across various price feed during the weekday, it may be more challenging to get a unified price feed on the weekend: first, while the underlying market is not closed, most of the broker and banks feeding the APIs price feed are; then, for the one providing a price feed, they often add a very big spread, which makes the price of forex pairs vary significantly by source; a way to mitigate it is to use the last price known.
```
smb2796, line 130: Small nit but doesn't this apply to all of these rates, not just EURUSD?

UMIPs/umip-29.md
```diff
@@ -0,0 +1,134 @@
...
+Rounding: Closest, 0.0001 up
+Decimal: 18 (1e18)
+Pricing Interval: 60 seconds
+Dispute timestamp rounding: down
+
+## Rationale
+Apart from the weekend, there is little to no difference in prices on liquid major Forex pairs like EURUSD, so any price feed could be used; however, for convenience, we recommend using the one of TraderMade.
+
+## Implementation
+The value of this identifier for a given timestamp should be determined by querying for the price of EURUSD, GBPUSD and CHFUSD from TraderMade.com for that timestamp. More specifically, users can query “https://marketdata.tradermade.com/api/v1/minute_historical?currency=EURUSD&date_time=2020-11-11-13:01&api_key=apikey” and use the close price as reference.
```
smb2796, line 67: Need XAU/USD here as well

UMIPs/umip-29.md
SayWhatDev: Yes, that is correct. The idea is that 18 decimals are set when writing on the blockchain otherwise 4 decimals of rounding are used. As ex. EUR/USD price is 1.23456 which rounds to 1.2345, but on the blockchain it is displayed 12345000000000000000000. @smb2796

UMIPs/umip-29.md
SayWhatDev: Thank you for the pointer :)

SayWhatDev: Found a few more places actually. Pushing a commit :) @smb2796

UMIPs/umip-29.md
SayWhatDev: I have added this info

UMIPs/umip-29.md
SayWhatDev: Got your point and changed it, thanks :)

UMIPs/umip-29.md
SayWhatDev: We want 18 decimals on chain, already added that info in the umip

UMIPs/umip-29.md
SayWhatDev: Fixed

UMIPs/umip-29.md
SayWhatDev: Added additional clarifications

UMIPs/umip-29.md
SayWhatDev: Added with a later commit :)

UMIPs/umip-29.md
SayWhatDev: This suggested change is part of a later commit

UMIPs/umip-29.md
SayWhatDev: @nicholaspai Sorry my previous comment was mistaken. However since in the derivative we use CHF/USD it is better to have CHF/USD as price identifier, however as mentioned in the umip voters will have to get USD/CHF price feed and calculate the reversed price.

UMIPs/umip-29.md
SayWhatDev: Fixed it :)

UMIPs/umip-29.md
```diff
@@ -0,0 +1,134 @@
...
+-----------------------------------------
+
+- Identifier name: **XAU/USD**
+- Base Currency: XAU
+- Quote Currency: USD
+- Decimals: 18
+
+-----------------------------------------
+
+Source: https://marketdata.tradermade.com
```
smb2796, line 54: Formatting here is not showing up correctly

SayWhatDev: @smb2796 Can you give me more details as I am not seeing an issue here locally?

smb2796: Looks like this: 
image

UMIPs/umip-29.md
smb2796: As discussed offline, the existing XAUUSD identifier should suffice

UMIPs/umip-29.md
```diff
@@ -0,0 +1,124 @@
+## Headers
+| UMIP-29    |                                                          |
+|------------|----------------------------------------------------------|
+| UMIP Title | Add EURUSD, CHFUSD and GBPUSD                            |
```
smb2796, line 4: Naming convention looks good!

UMIPs/umip-29.md
```diff
@@ -0,0 +1,124 @@
...
+
+-----------------------------------------
+
+- Identifier name: **CHFUSD**
+- Base Currency: CHF
+- Quote Currency: USD
+
+-----------------------------------------
+
+Source: https://marketdata.tradermade.com <br>
```
smb2796, line 44: Thanks for fixing the formatting

UMIPs/umip-29.md
```diff
@@ -0,0 +1,124 @@
...
+Rounding: Closest, 0.5 up (> .000005 rounds up, < .000005 rounds down) <br>
+Decimal: 18 (1e18) <br>
+Pricing Interval: 60 seconds <br>
+Dispute timestamp rounding: down <br>
+
+## Rationale
+Apart from the weekend, there is little to no difference in prices on liquid major Forex pairs like EURUSD, so any price feed could be used; however, for convenience, we recommend using the one of TraderMade.
+
+## Implementation
+The value of this identifier for a given timestamp should be determined by querying for the price of EURUSD, GBPUSD and CHFUSD from TraderMade.com for that timestamp. More specifically, users can query “https://marketdata.tradermade.com/api/v1/minute_historical?currency=EURUSD&date_time=2020-11-11-13:01&api_key=apikey” and use the close price as reference.
```
smb2796, line 57: This should really be the price request timestamp rounded back to the nearest minute, correct?

UMIPs/umip-29.md
```diff
@@ -0,0 +1,124 @@
...
+- Quote Currency: USD
+
+-----------------------------------------
+
+Source: https://marketdata.tradermade.com <br>
+Result Processing: Median (if more than 1 source is utilized), otherwise Exact Price <br>
+Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus. <br>
+Price Steps: 0.00001 (5 decimals in more general trading format) <br>
+Rounding: Closest, 0.5 up (> .000005 rounds up, < .000005 rounds down) <br>
+Decimal: 18 (1e18) <br>
```
smb2796, line 49: This decimal specification should be removed as decimals within the UMIP specify rounding decimals not scaling decimals and the identifier will be treated as 18 decimals of scaling by default

smb2796, line 49: Let's call this Scaling Decimals to differentiate from the price steps

UMIPs/umip-29.md
SayWhatDev: I believe so yes @smb2796

UMIPs/umip-29.md
SayWhatDev: Fixed and pushed :)

UMIPs/umip-29.md
```diff
@@ -0,0 +1,124 @@
...
+TraderMade is provided as an accessible source to query for this data, but ultimately how one queries for these rates should be varied and determined by the voter to ensure that there is no central point of failure.
+While it's important for tokenholders to have redundancy in their sources, bots and users that interact with the system in realtime need fast sources of price information. In these cases, it can be assumed that the exchange median is accurate enough
+
+### Price feed - liquidations and disputes
+
+Liquidation and dispute bots should have their own subscription to price feeds. Our price-feed provider’s API documentation can be found [here](https://marketdata.tradermade.com/documentation).
+
+In the case of a TraderMade outage voters can turn to any other available price feed API or a broker API, as the price feeds for the forementioned financial assets does not differ much between different providers. There might be some slight differences, however they are quite insignificant and would not affect the liquidation or dispute processes. In that case we'll use a median price from the different price feeds provided by the voters.
```
smb2796, line 83:
```suggestion
In the case of a TraderMade outage voters can turn to any other available price feed API or a broker API, as the price feeds for the forementioned financial assets does not differ much between different providers. There might be some slight differences, however they are quite insignificant and would not affect the liquidation or dispute processes. For this case, we provide options for additional price feed providers that voters could utilize.
```

UMIPs/umip-29.md
```diff
@@ -0,0 +1,124 @@
...
+-----------------------------------------
+
+- Identifier name: **CHFUSD**
+- Base Currency: CHF
+- Quote Currency: USD
+
+-----------------------------------------
+
+Source: https://marketdata.tradermade.com <br>
+Result Processing: Median (if more than 1 source is utilized), otherwise Exact Price <br>
```
smb2796, line 45: I think Median is slightly confusing with two sources. Also, this should really be for the recommended way of querying the price. I think providing the alternative sources as options for voters is more than enough. I would recommend removing median.

```suggestion
Result Processing: Exact Price <br>
```

smb2796 (approved):
Looks good, thank you for addressing all of the comments. Going to approve and merge in. 

Spoke to @Pascaltallarida offline, and going to update the two outstanding comment on his behalf after merging.

PR #103, opened 2020-12-22
Draft to add an improved price identifier UMIP template

smb2796:


price-id-template.md
```diff
@@ -0,0 +1,55 @@
+This is the suggested template for new UMIPs that propose supporting new price identifiers.
+
+When opening a pull request to submit your UMIP, please use an abbreviated title in the filename, umip-draft_title_abbrev.md.
+
+The title should be 44 characters or less. 
+
+All italicized text is provided for instructional purposes and should be removed before draft submission if unused in your UMIP.
+
+## Headers
```
chicfilabae, line 9: Just curious, where do folks write their rationale for why they believe this price identifier makes sense to add? Seems that section `Motivation` is missing

smb2796: Very much WIP. Motivation was intentionally left out as there is some thought that this should be optional content. Interested to hear your thoughts on if this makes sense.

chicfilabae: Got it. 

My 2 wei: it seemed that the UMIPs with a stronger motivation led to a quicker process in terms of releasing UIs, spinning up liq+dispute bots, gaining TVL, etc. I would imagine it's the strongest filter from folks tossing up random UMIPs with products that might not have any demand but would trigger extra votes for them to accrue rewards. I'm not naming any UMIPs in particular here, but just my general thoughts on preventing inorganic UMIP bloat.

price-id-template.md
```diff
@@ -0,0 +1,55 @@
...
+- Created: <date created on>
+
+
+## Identifier Specifications
+
+- Identifier Name: *(BTCUSD)*
+- Data Source(s): (*Binance, Kraken, Coinbase)*
+- Result Processing: *(Median)*
+- Input Processing: *(None. Human intervention in extreme circumstances where the result differs from broad market consensus.)*
+- Decimals: *(4, .0001)*
```
chrismaree, line 23: Seeing this is an example this matters less but a BTC denominated UMIP (like [UMIP7](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-7.md#technical-specification) should have 8 decimal points or a lot of precision is lost in rounding.

price-id-template.md
```diff
@@ -0,0 +1,55 @@
...
+*These endpoints should provide more than 72 hours worth of historical data availability. They should also provide prices at a rate more often than once per hour.*
+
+*Please list any additional transformation steps that need to be performed on the returned prices.*
+
+## Rationale for Data Sources
+*(Bitstamp, Binance, and Kraken were chosen because they have the highest volume for these assets.  We choose USDT instead of USD because the USDT/BTC pair is what is more commonly traded and the exchange rate between USD/USDT is not meaningfully variable.)*
+
+
+## Price Feed
+*Include a link to a PR to the UMA protocol repo with an example price feed that inherits this  [PriceFeedInterface](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/PriceFeedInterface.js):*
```
chrismaree, line 47: This seems like a pretty strong requirement. I'm in favor of it but it _will_ slow down UMIPs potentially as this requires a pretty high technical barrier for some implementors. 

Would it be sufficient to have _some_ reference implementation that voters could run & then include this requirement within developer mining?

nicholaspai:
Should we add a section about naming conventions? See [this comment](https://github.com/UMAprotocol/UMIPs/pull/106#discussion_r550192923)

smb2796:
Closing due to replacement by pull #123

PR #107, opened 2020-12-29
fix(umip-28) Clarify that stablespread should be denominated in ethusd 

chicfilabae:
As discussed with Matt and Nick offline.

### DIFF
--- a/UMIPs/umip-28.md
+++ b/UMIPs/umip-28.md
@@ -7,7 +7,7 @@
 | Created    | December 12, 2020                                                                                                                           |
  
 ## Summary (2-5 sentences)
-The DVM should support price requests for the STABLESPREAD price index. STABLESPREAD is defined as: `min(max(A - B + 1, 0), 2)`, where `A` refers to an equally weighted basket of {UST, CUSD, BUSD}. B is `MUSD`. UST is TerraUSD, the interchain stablecoin connected by the Cosmos IBC. CUSD is Celo Dollar, the stablecoin on the Celo network. BUSD is Binance USD, a stablecoin issued by Paxos in partnership with Binance. `MUSD` is an autonomous and non-custodial stablecoin on mStable, representing a basket of stablecoins on Ethereum, namely USDT, USDC, TUSD, and DAI. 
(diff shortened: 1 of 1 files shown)


UMIPs/umip-28.md
```diff
@@ -52,7 +52,8 @@ The value of STABLESPREAD for a given timestamp can be determined with the follo
 4. Perform A - MUSD + 1
 5. Perform the maximum of the result of step 4 and 0. This is to ensure non-negativity. 
 6. Perform the minimum of the result of step 5 and 2. This is to ensure symmetry.
-7. This result should be rounded to eight decimal places.
+7. Take the result of step 6 and divide by the value of ETHUSD
```
nicholaspai, line 55: You should also specify in the `Technical Specification` section how ETHUSD will be retrieved. You could just reference the ETHUSD price identifier from another UMIP or just define 2 exchanges to use as data sources.

nicholaspai (commented):
LGTM, but should we also modify the description to make it clear that we are computing stablespread / eth? (i.e. stablespread, denominated in ETH)

smb2796:
What was the reasoning for removing Okcoin?

chicfilabae:
@nicholaspai requested it, as "OKCoin exchange is not available on cryptowatch, I think its OK just to use bittrex for now since thats where 90% of volume is....Feel free to quote me on if you want to make that change"

I agree with the reasoning as well, so I removed it

UMIPs/umip-28.md
```diff
@@ -27,7 +27,7 @@ The definition of this identifier should be:
  
 - Identifier name: STABLESPREAD
 - Base Currency: STABLESPREAD
```
smb2796, line 29: Similar to my comments on your other UMIPs, since this is quoted in ETH I think we should list a quote currency of ETH.

chicfilabae: Ack.

claytonroche (dismissed):

UMIPs/umip-28.md
```diff
@@ -52,7 +53,8 @@ The value of STABLESPREAD for a given timestamp can be determined with the follo
 4. Perform A - MUSD + 1
 5. Perform the maximum of the result of step 4 and 0. This is to ensure non-negativity. 
 6. Perform the minimum of the result of step 5 and 2. This is to ensure symmetry.
-7. This result should be rounded to eight decimal places.
+7. Take the result of step 6 and divide by the value of ETH/USD
```
chrismaree, line 56:
```suggestion
7. Take the result of step 6 and divide by the value of ETH/USD.
```

smb2796 (approved):
Looks good!

PR #108, opened 2020-12-29
Draft for addition of STABLESPREAD-BTC as a price identifier

chicfilabae:
Similar to UMIP 28.


UMIPs/umip-30.md
```diff
@@ -0,0 +1,70 @@
...
+
+Let's walk through some cases to better understand what happens in various cases for STABLESPREAD-BTC. If both of these baskets were to trade at their theoretical value, then the resulting value of STABLESPREAD-BTC will be 1. If the Ethereum based stablecoin basket loses its peg and appreciates in value relative to the non-Ethereum stablecoin basket, then the value of STABLESPREAD-BTC will be less than 1. If the opposite happens, then it will be more than 1. It is important that this only measures "relative ability to maintain peg". That is, if both of them lose their peg in the same form, the resulting value still may be 1. 
+ 
+There is little cost associated with adding this price identifier, as there are multiple free and easily accessible data sources available.
+ 
+## Technical Specification
+The definition of this identifier should be:
+ 
+- Identifier name: STABLESPREAD-BTC
```
smb2796, line 28: Similar comment as on the STABLESPREAD-USDC proposal. Since this is quoted in BTC, I think the Base currency should be STABLESPREAD and the quote currency should be BTC. Also potentially should change the identifier name to be clearer about this.

UMIPs/umip-30.md
```diff
@@ -0,0 +1,70 @@
...
+- Base Currency: STABLESPREAD-BTC
+- Data Sources: {Bittrex: UST/USDT, Uniswap V2: UST/USDT} for UST, {Binance: BUSD/USDT, Uniswap V2: BUSD/USDT} for BUSD, {Bittrex: CUSD/USDT} for CUSD, {Balancer: MUSD/USDC, Uniswap V2: MUSD/USDC} for MUSD, {Coinbase: BTC/USD, Bitfinex: BTC/USD} for BTC
+- Result Processing: For each constituent asset of the basket, the average of both exchanges
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+- Decimals: 8
```
smb2796, line 33: Decimals look good if the intention is to use with renBTC

UMIPs/umip-30.md
```diff
@@ -0,0 +1,70 @@
...
+- Rounding: Closest, 0.5 up
+- Pricing Interval: 60 seconds
```
smb2796, line 35: Same as STABLESPREAD-USDC comment, this should be adjusted to match the `Implementation` section.

UMIPs/umip-30.md
chicfilabae: Yes

UMIPs/umip-30.md
chicfilabae: Done

UMIPs/umip-30.md
chicfilabae: Done

smb2796:
Seems best if this and STABLESPREAD-USDC are contained within one UMIP. Most of the content can also reference the existing STABLESPREAD UMIP except for the Technical Specifications and Implementation sections. This seems like a cleaner approach as it is easier to understand, removes redundant content and reduces the amount of unnecessary votes.

chicfilabae:
Okay

chicfilabae:
Closing given that the PRs have been combined.

PR #111, opened 2020-12-30
fix(umip-29): Update language for alternative data sources and result processing specification

smb2796:
As discussed with Pascal offline, merging in the suggested outstanding comments for editing the language around alternative data sources as well as changing the result processing to match the `Implementation` section.


SayWhatDev (approved):
All is good with the added changes.

Tommy1231232 (approved):

PR #124, opened 2021-01-12
improve(umip-30): Missing specification for DAI price feed source

nicholaspai:
BAC/DAI is listed as a price feed source, DAI source is missing details


chicfilabae:
Can you remove the language regarding the ETH normalization below 

"Normalizing the ETH result to USDC for FRAX, ESD, and BAC."

It's not needed anymore.

smb2796 (approved):
Looks good to me except for needing to remove the language that chicfilabae mentioned.

PR #125, opened 2021-01-13
Add Dynamic Set Dollar (DSD) as collateral

akshaysrivastav:

PR #128, opened 2021-01-14
Add DSDUSD and USDDSD as price identifiers

akshaysrivastav:


UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
...
+## Summary (2-5 sentences)
+The DVM should support price requests for the DSD/USD and USD/DSD price index.
+
+## Motivation
+The DVM currently does not support the DSDUSD or USDDSD price index.
+Supporting the DSDUSD price identifier would enable the creation of the DSDO stablecoin, backed by DSD as collateral. DSD token holders can utilize this as a hedging tool, and could go long or use DSDO for functional purposes.
+There is little cost associated with adding this price identifier, as there are multiple free and easily accessible data sources available.
+
+More information on the Dynamic Set Dollar can be found on the website: https://dsd.finance/
+
```
Tommy1231232, line 18: There is a price identifier UMIP guide and template below. Can you add in a section for "Markets & Data Sources" per the below guide? 

https://github.com/UMAprotocol/UMIPs/blob/master/price-identifier-guide.md

Tommy1231232: To allow for the creation of bots that can programmatically calculate prices off-chain to liquidate and dispute transactions, you must create a price feed following the UMA Protocol format (outlined below). This price feed is also necessary to calculate developer mining rewards.

https://github.com/UMAprotocol/UMIPs/blob/master/price-identifier-guide.md

UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
...
+## Technical Specification
+The definition of this identifier should be:
+- Identifier name: DSDUSD
+- Base Currency: DSD
+- Quote Currency: USD(C)
```
Tommy1231232, line 23: This can just be USD. No need for the (C)

UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
...
+- Exchanges: Uniswap
```
Tommy1231232, line 24: Are there any other exchanges DSD is traded on? It's best to take the 3 largest markets

UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
...
+- Result Processing: No processing.
```
Tommy1231232, line 25: What type of processing should voters perform to determine the final price of an asset (e.g., take the median or mode of all votes submitted)?

    Note - a result processing of "median" is more resilient to market manipulation versus a result processing of "average".

UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
...
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+- Price Steps: 0.00001 (5 decimals in more general trading format)
```
Tommy1231232, line 27: Price identifiers need to be automatically scaled to reflect the units of collateral that a price represents. Because of this, the amount of decimals that a price is scaled to needs to match the used collateral currency.

Since DSD is the collateral currency I see it has 18 decimals. So would suggest changing this from 5 to 18. 

@smb2796 do you agree?

UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
...
+- Rounding: Closest, 0.5 up
+- Pricing Interval: 60 seconds
+- Dispute timestamp rounding: down
+
+The definition of this identifier should be:
+- Identifier name: USDDSD
+- Base Currency: USD(C)
```
Tommy1231232, line 34: Same comment as above

UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
...
+- Quote Currency: DSD
+- Exchanges: Uniswap
+- Result Processing: 1/DSDUSD.
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+- Price Steps: 0.00001 (5 decimals in more general trading format)
```
Tommy1231232, line 39: Same comment as above

UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
...
+
+Prices are primarily used by Priceless contracts to calculate a synthetic token’s redemptive value in case of liquidation or expiration. Contract counterparties also use the price index to ensure that sponsors are adequately collateralized. More complex computations (like incorporating additional exchanges, calculating a TWAP or VWAP, or imposing price bands, etc.) have the potential to add a greater level of precision and robustness to the definition of this identifier, particularly at settlement of each expiring synthetic token.
+
+The addition of DSDUSD  and USDDSD fits into a larger goal of advancing the adoption of the UMA protocol by allowing DSD to be used as collateral for minting a stable coin among a suite of [OpenDAO](https://opendao.io) stable coins. This furthers adoption of the protocol by encouraging a convergence of capital from different projects and increasing TVL.
+
+Currently the largest exchange with USD or stablecoin markets for DSD is [Uniswap](https://info.uniswap.org/pair/0x66e33d2605c5fb25ebb7cd7528e7997b0afa55e8) (~90% volume). If liquidity is withdrawn too fast there may be a risk in the price peg, and therefore the integrity of the system.
```
Tommy1231232, line 52: Per comment above, it would be good to add the top 3 markets for DSD instead of only one

UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
...
+
+In the current setting, there will need to be a significant event that erodes confidence in DSD and the token, at the same time where Uniswap liquidity is withdrawn too quickly en masse. However this may pose a security issue depending on the depth of DSDUSD liquidity on Uniswap. This may be improved by listing on additional exchanges with genuine volume.
+Over time, as liquidity in the DSD token migrates across platforms, this identifier can be re-defined to add exchanges, remove exchanges, or change the way that the price is calculated. Any re-definition would be done via off-chain social consensus by $UMA-holders, and ultimately reflected in the way that $UMA-holders vote upon the price of DSDUSD and USDDSD when called to do so by disputers, or at settlement.
+
+
+
+## Implementation
+
+The value of this identifier for a given timestamp should be determined by querying the price from DSDUSDC pair https://info.uniswap.org/pair/0x66e33d2605c5fb25ebb7cd7528e7997b0afa55e8 on Uniswap for that timestamp. Most of the DSD trading volume happens on Uniswap, which forms the broad market consensus.
```
Tommy1231232, line 61: Can you document each step a voter should take to query for and return a price at a specific timestamp? The goal here is to make it so easy a caveman can do it ;)

UMIPs/umip-draft_DSD_price_identifier.md
akshaysrivastav: Hey @Tommy1231232, the primary and major listing for DSD is uniswap. There are multiple pairs on Uniswap as well but the pair with USDC has the most volume and liquidity.

UMIPs/umip-draft_DSD_price_identifier.md
akshaysrivastav: Okay, we thought the precision also has to account for pair's currency decimals like USDC which only has 6 decimals. We will make it 18 now.

UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
...
+- Base Currency: DSD
+- Quote Currency: USD(C)
+- Exchanges: Uniswap
+- Result Processing: No processing.
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+- Price Steps: 0.00001 (5 decimals in more general trading format)
+- Rounding: Closest, 0.5 up
+- Pricing Interval: 60 seconds
+- Dispute timestamp rounding: down
+
```
Tommy1231232, line 31: What is the intended collateral currency this price ID would use?

UMIPs/umip-draft_DSD_price_identifier.md
```diff
@@ -0,0 +1,68 @@
+## Headers
+| UMIP-#    |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add DSDUSD, USDDSD as price identifiers              |
```
Tommy1231232, line 4: After further discussion, since there is only one market lets change DSDUSD and USDDSD names to DSDUSDC and USDCDSD (and update below in technical specifications)

UMIPs/umip-draft_DSD_price_identifier.md
Tommy1231232: Do you plan on using other collateral types?

UMIPs/umip-draft_DSD_price_identifier.md
Tommy1231232: Can we note that we are assuming USDC = $1?

Tommy1231232 (commented):
Akshaysrivastav, Great job, if you can address the minor comments within the next 24 hours we can get this up for a vote this week!

UMIPs/umip-draft_DSD_price_identifier.md
akshaysrivastav: Sure, I'll add a note.

UMIPs/umip-draft_DSD_price_identifier.md
akshaysrivastav: We plan to use Dynamic Set Dollar (DSD) as the collateral currency. The UMIP for adding DSD as collateral whitelist is also in process here - https://github.com/UMAprotocol/UMIPs/pull/125

UMIPs/umip-draft_DSD_price_identifier.md
akshaysrivastav: Updated

UMIPs/umip-draft_DSD_price_identifier.md
akshaysrivastav: Added result processing of "median"

UMIPs/umip-draft_DSD_price_identifier.md
akshaysrivastav: Updated

UMIPs/umip-draft_DSD_price_identifier.md
chrismaree: Can you please add to the specification that this is the intended collateral currency for these financial contracts. 

In addition, can you add the number of decimal points for this collateral (DSD is 18) as this is required to inform the scalling the identifier correctly for the EMP.

akshaysrivastav: Sure @chrismaree I already updated the _price steps_ field to 0.000000000000000001. Should I add this value again in another field (like `Collateral currency decimals: 0.000000000000000001`)?

Tommy1231232 (approved):

PR #140, opened 2021-01-19
Draft for addition of basis synthetics as a price identifier

chicfilabae:
Draft for addition of basis synthetics as a price identifier


chicfilabae:
@smb2796 
@claytonroche 
@nicholaspai

UMIPs/umip-35.md
```diff
@@ -0,0 +1,197 @@
...
+
+    FTX: BTC/USDT, BTC-0326, BTC-0625
+    Binance: BTC/USDT, BTCUSD Quarterly 0326, BTCUSD Quarterly 0625
+    Huobi: BTC/USDT, BTC0326Quarterly, BTC0625Bi-Quarterly
+
+3. Provide recommended endpoints to query for real-time prices from each market listed. 
+
+    FTX: [BTC](https://ftx.com/api/markets/BTC/USDT), [BTC-0326](https://ftx.com/api/markets/BTC-0326), [BTC-0625](https://ftx.com/api/markets/BTC-0625)
+
+    Binance: [BTC](https://api.binance.com/api/v3/ticker/price?symbol=BTCUSDT)
```
pemulis, line 54: Can you add links for the Binance quarterly endpoints?

UMIPs/umip-35.md
```diff
@@ -0,0 +1,197 @@
...
+    Binance: BTC/USDT, BTCUSD Quarterly 0326, BTCUSD Quarterly 0625
+    Huobi: BTC/USDT, BTC0326Quarterly, BTC0625Bi-Quarterly
+
+3. Provide recommended endpoints to query for real-time prices from each market listed. 
+
+    FTX: [BTC](https://ftx.com/api/markets/BTC/USDT), [BTC-0326](https://ftx.com/api/markets/BTC-0326), [BTC-0625](https://ftx.com/api/markets/BTC-0625)
+
+    Binance: [BTC](https://api.binance.com/api/v3/ticker/price?symbol=BTCUSDT) 
+
+    Huobi: [BTC](https://api.huobi.pro/market/trade?symbol=btcusdt), [BTC0326Quarterly](https://api.hbdm.com/market/trade?symbol=BTC_CQ), [BTC0625Bi-Quarterly](https://api.hbdm.com/market/trade?symbol=BTC_NQ)
```
pemulis, line 56: Can the quarterly and bi-quarterly tickers be specified in the URL?

UMIPs/umip-35.md
```diff
@@ -0,0 +1,197 @@
...
+
+    
+4. How often is the provided price updated?
+
+   The lower bound on the price update frequency is a second.
+
+5. Provide recommended endpoints to query for historical prices from each market listed. 
+
+   FTX, use this endpoint: https://ftx.com/api/markets/BTC/USDT/candles?resolution=15&limit=1&start_time=1608241140&end_time=1608241150
+   Huobi, use this endpoint: https://api.hbdm.com/market/history/trade?symbol=BTC_CQ&size=100
```
pemulis, line 66: Is there a better way of querying a particular historical timestamp range for Huobi? Would this query provide data going back far enough for a DVM voter to find the specific timestamp they're looking for?

chicfilabae: Not that we have seen unfortunately, would be open to any suggestions that you may have here

UMIPs/umip-35.md
chicfilabae: Don't believe so, the idea is to abstract away the notion of the specific dates to the user where Huobi specifies it as `NQ` and `CQ` where hitting the endpoint handles rolls internally without the user modifying

UMIPs/umip-35.md
```diff
@@ -0,0 +1,197 @@
...
+
+<br>
+
+# Security considerations
+
+**Example questions**
+
+1. Where could manipulation occur?
+
+As discussed in the `What is the potential for the price to be manipulated on the chosen exchanges?`, manipulation could occur at either at the spot layer or the futures layer on each of the each of the aforementioned exchanges. To disrupt BTC or ETH spot of futures on the biggest exchanges globally would require capital at the highest scale, and the dislocation, were it to be caused, would be short-dated due to the efficiency of those markets.
```
rubberbandedman, line 179: seems like it'd be worthwhile to specify "clamps" (just like how the funding rate clamps exist for centralized futures + perpetual swaps exchanges) to mitigate manipulation here, no? even as something as generous as 20% below and above the price...seems most of the time it would be inactive constraints but would help prevent liquidations

UMIPs/umip-35.md
```diff
@@ -0,0 +1,197 @@
...
+
+11.  What would be the cost of sending 15,000 queries?
+
+   In the context of the above answers, doesn't make a ton of sense given there is no explicit price attached to it. 
+
+<br>
+
+# PRICE FEED IMPLEMENTATION
+
+The price feed implementation is heavily-based on @Nick's [PR](https://github.com/UMAprotocol/protocol/pull/2354/files) for a BasketSpreadPriceFeed, we currently put it up at [here](https://gist.github.com/chicfilabae/6e30982a13b23ed97e8e2e3bf54ff9f8). Will create a separate PR for it, but don't have an intention of abstracting out such that both stablespread + basis can use the same script.
```
rubberbandedman, line 96: seems this should be a separate PR as you say, but the gist is helpful by itself

UMIPs/umip-35.md
chicfilabae: Yes, added! Not sure why they weren't added the first time, slipped through

UMIPs/umip-35.md
```diff
@@ -0,0 +1,198 @@
...
+
+<br>
+
+# PRICE FEED IMPLEMENTATION
+
+The price feed implementation is heavily-based on @Nick's [PR](https://github.com/UMAprotocol/protocol/pull/2354/files) for a BasketSpreadPriceFeed, we currently put it up at [here](https://gist.github.com/chicfilabae/6e30982a13b23ed97e8e2e3bf54ff9f8). Will create a separate PR for it, but don't have an intention of abstracting out such that both stablespread + basis can use the same script. 
+
+# TECHNICAL SPECIFICATIONS
+
+**1. Price Identifier Name** - BTC-BASIS-3M/USDC, BTC-BASIS-6M/USDC, ETH-BASIS-3M/USDC, ETH-BASIS-6M/USDC
```
Tommy1231232, line 101: Suggest changing the name ti [BTC/BASIS-3M]/USDC to stick with standards

UMIPs/umip-35.md
```diff
@@ -0,0 +1,198 @@
...
+
+**2. Base Currency** - BTC-BASIS-3M, BTC-BASIS-6M, ETH-BASIS-3M, ETH-BASIS-6M
+
+**3. Quote currency** - USDC
```
Tommy1231232, line 105: We could make this USD to generalize the price id

UMIPs/umip-35.md
chicfilabae: Great idea. Adding it in now.

UMIPs/umip-35.md
chicfilabae: Would rather keep it specific to USDC for now

UMIPs/umip-35.md
```diff
@@ -0,0 +1,198 @@
+## HEADERS
+| UMIP-35     |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add BTC-BASIS-3M, BTC-BASIS-6M, ETH-BASIS-3M, and ETH-BASIS-6M as price identifiers              |
```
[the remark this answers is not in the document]
chicfilabae: Yes, updating the names to be [BTC/BASIS-3M]/USDC now

UMIPs/umip-35.md
chicfilabae: What would be the meaning of the BTC/BASIS-3M part? Wouldn't the `/` lose meaning there?

UMIPs/umip-35.md
chicfilabae: Done!

UMIPs/umip-35.md
chicfilabae: Okay will do!

chicfilabae: https://github.com/UMAprotocol/protocol/pull/2457

chicfilabae: @nicholaspai

UMIPs/umip-35.md
```diff
@@ -0,0 +1,204 @@
...
+| Authors    | Bae (bae@youmychicfila.com), K (k@youmychicfila.com) |
+| Status     | Draft                                                                                                                                    |
+| Created    | January 18, 2020                                                                                                                           |
+                                                                          
+# SUMMARY 
+
+The DVM should support price requests for both BTC-BASIS-3M/USDC, BTC-BASIS-6M/USDC, ETH-BASIS-3M/USDC, ETH-BASIS-6M/USDC price indices. BTC-BASIS-3M/USDC is defined as: min(max(100.0 * [1 + (F-S)/S], 75.0), 125.0), where `F` references a basket of BTC futures that are set to expire in March and `S` represents a basket of BTC spot prices. ETH-BASIS-3M/USDC is defined in the same way, except the `F` and `S` correspond to `ETH` futures and spot. Similarly, the `6M` `BTC and `ETH` price identifiers would be the same formula except with futures that are set to expire in `June` instead of `March`.
```
smb2796, line 11:
```suggestion
The DVM should support price requests for both BTC-BASIS-3M/USDC, BTC-BASIS-6M/USDC, ETH-BASIS-3M/USDC, ETH-BASIS-6M/USDC price indices. BTC-BASIS-3M/USDC is defined as: min(max(100.0 * [1 + (F-S)/S], 75.0), 125.0), where `F` references a basket of BTC futures that are set to expire in March and `S` represents a basket of BTC spot prices. ETH-BASIS-3M/USDC is defined in the same way, except the `F` and `S` correspond to `ETH` futures and spot. Similarly, the `6M` `BTC` and `ETH` price identifiers would be the same formula except with futures that are set to expire in `June` instead of `March`.  
```

UMIPs/umip-35.md
Tommy1231232: Yes you are right. Lets just keep the brackets so [BTC-BASIS-3M]/USDC

UMIPs/umip-35.md
```diff
@@ -0,0 +1,204 @@
...
+
+    1 minute
+
+3. **Input processing**
+
+    None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+
+4. **Result processing** 
+
+    - **Note** - a result processing of "median" is more resilient to market manipulation versus a result processing of "average"
```
smb2796, line 170: The unused questions and notes can be removed from the actual umip

UMIPs/umip-35.md
```diff
@@ -0,0 +1,204 @@
...
+
+# Security considerations
+
+**Example questions**
+
+1. Where could manipulation occur?
+
+As discussed in the `What is the potential for the price to be manipulated on the chosen exchanges?`, manipulation could occur at either at the spot layer or the futures layer on each of the each of the aforementioned exchanges. To disrupt BTC or ETH spot of futures on the biggest exchanges globally would require capital at the highest scale, and the dislocation, were it to be caused, would be short-dated due to the efficiency of those markets. 
+
+Further, towards expiration, the front-month futures will become less liquid and easier to manipulate in terms of required capital. As such, we will ensure that the expiration of the synths using these price identifiers will expire a week _before_ any of the constituent futures in the basket expire. This way, we will minimize this attack vector.
```
smb2796, line 186: These will just roll into new 3m/6m futures, correct?

smb2796: Disregard. I see you noted this below.

UMIPs/umip-35.md
```diff
@@ -0,0 +1,204 @@
...
+
+Mentioned above. 
+
+3. Do the instructions for determining the price provide people with enough certainty?
+
+Yes.
+
+4. What are current or future concern possibilities with the way the price identifier is defined?
+
+When the 3m and 6m expire, we'd have to recreate another set of price identifiers to encode maturity into the price identifier. However, when I spoke to Matt recently, I understand that there are upcoming DVM upgrades that'd enable passing in additional state so as to not create price identifier bloat, or in this case, recreating one every time there is a new future.
```
smb2796, line 200: Could we just have the pid naturally use the current futures baskets? That way when rollover happens, it would still reference the most used futures baskets from FTX/Huobi/Binance.

UMIPs/umip-35.md
```diff
@@ -0,0 +1,204 @@
...
+
+    
+4. How often is the provided price updated?
+
+   The lower bound on the price update frequency is a second.
+
+5. Provide recommended endpoints to query for historical prices from each market listed. 
+
+   FTX, use this endpoint: https://ftx.com/api/markets/BTC/USDT/candles?resolution=15&limit=1&start_time=1608241140&end_time=1608241150
+   Huobi, use this endpoint: https://api.hbdm.com/market/history/trade?symbol=BTC_CQ&size=100
```
smb2796, line 66: Doesnt look like this endpoint returns enough historical data

UMIPs/umip-35.md
chicfilabae: Okay got it!

UMIPs/umip-35.md
```diff
@@ -0,0 +1,204 @@
...
+
+**5. Collateral Decimals** - 6 decimals.
+
+**6. Rounding** - Rounded to 6 decimals in accordance with above.
+
+- **Note** - this should always be less than or equal to the `Intended Collateral Currency` field.
+
+<br>
+
+# RATIONALE
```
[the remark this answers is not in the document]
chicfilabae: Traditionally, the definition of basis is futures minus spot, we're just sticking to convention for people's familiarity to CeFi world.

UMIPs/umip-35.md
chicfilabae: Yeap!

smb2796:
Awesome UMIP and thanks for creating a price feed. Because review is needed for the price feed as well, we think this will need to wait for voting until the price feed PR is merged in.

Also:

- The Huobi endpoint doesn't seem to return enough price data to be used for voting. I'm wondering if is there another market that could be used or another way we could query for this.
- Interested in more detail on how the futures are rolled. Wondering if there is some way we could we bake this into the price id and roll to referencing different futures as they are created.

UMIPs/umip-35.md
chicfilabae: Done

UMIPs/umip-35.md
chicfilabae: Done!

UMIPs/umip-35.md
chicfilabae: Yeap, good idea.

Will remove this.

chicfilabae: Hmmm, good thought here, in our minds the issue when we were considering something like this is the ambiguity you get when new futures are released before the current ones expire. i.e. 2 weeks before all CeFi futures mature, new futures are launched and the "definition" of 3m / 6m gets ambiguous there.

As mentioned, we want to have the optionality to define what we view as the "3m" and "6m" maturities to maximize economic security for the sponsors.  

I see your point here, though. It's my understanding the DVM upgrades will be out by then such that we can pass in context and we wouldn't have to create new price identifiers.

chicfilabae:
Both valid points, we may just switch out to OKex if no one can find a reasonable historical Huobi price feed.

On the second point, see my comment above.

Regarding the PR merge into `protocol`, you think it's acceptable to have a gist with that code as we've done in the past or nah?

UMIPs/umip-35.md
chicfilabae: Agree

UMIPs/umip-35.md
chicfilabae: Appears Matt just said late Feb for the DVM upgrades, so think this would be resolved before the first 3M expire

chicfilabae:
Based on more discussion, we decided to replace Huobi with okex as the API is more reliable + historical API is more readily accessible, and the futures nomenclature explicitly specify the expiry date.

UMIPs/umip-35.md
chicfilabae: Fixed.

smb2796:
> Regarding the PR merge into protocol, you think it's acceptable to have a gist with that code as we've done in the past or nah?

I'd say nah not needed. Could just link directly to the PR or price feed once merged.

UMIPs/umip-35.md
smb2796: Sounds like reasonable justification to leave as is for the first go around. The passed context you refer to is something that we are calling "ancillary data". The contracts themselves will also need to support using this which will be allowed from an accompanying EMP upgrade. I believe you would need to deploy new EMPs to use this and legacy EMPs would not be supported. @mrice32 @chrismaree @nicholaspai can one of you confirm this?

This price id may also need to be amended to allow for generalization and instructions about calculating price based upon the passed context. Optionally I think you could also do this now by using the instructions you provide as the "go-to" price calculation methodology, but note the specific parameters that could be overridden by passed ancillary data.

UMIPs/umip-35.md
```diff
@@ -0,0 +1,200 @@
...
+
+3. **Input processing**
+
+    None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+
+4. **Result processing** 
+
+    For BTC futures prices, the median of the prices from the 3 exchanges should be used. For BTC spot prices, the median of the prices from the 3 spot exchanges should be used. Same for ETH.
+
+    As an example, for [BTC-BASIS-3M], the formula is min(max(100.0 * [1 + (F-S)/S], 75.0), 125.0), where F would be a median of the 3 futures prices as mentioned above and S would be a median as well of the 3 spot prices.
```
smb2796, line 170: You should note results rounding here as well.

UMIPs/umip-35.md
```diff
@@ -0,0 +1,200 @@
...
+   Okex, use this endpoint: https://www.okex.com/api/futures/v3/instruments/BTC-USD-210625/history/candles?start=2021-01-20T02:31:00.000Z&granularity=60
+   Binance, use this endpoint: https://dapi.binance.com/dapi/v1/aggTrades?symbol=btcusd_210326&limit=100&startTime=1611437831000&endTime=1611537831000
+
+6.  Do these sources allow for querying up to 74 hours of historical data? 
+
+   Yes
+
+7.  How often is the provided price updated?
+
+   The lower bound on the price update frequency is a second.
```
smb2796, line 75: Small nitpick, but it looks like Okex history updates every minute not second.

chicfilabae: Fair point

UMIPs/umip-35.md
chicfilabae: Yeap, we will redeploy once more w/ the new EMP contracts

UMIPs/umip-35.md
chicfilabae: Done

UMIPs/umip-35.md
chicfilabae: Done

UMIPs/umip-35.md
chrismaree: ye, that's right. to take advantage of the Ancillary data you'll need new EMPs & possibly new identifies that encode the definition of these extra data fields. For now, the way you guys have it define here I think is fine as it is basically a short-term fix until we've set up the new system.

smb2796 (approved):
Due to there being no outstanding comments, I am going to approve and merge in as Last Call. Once the price feed has been merged, we can update to link to the merged code rather than the PR.

Can you just change the UMIP title and filename to umip-40?

UMIPs/umip-35.md
```diff
@@ -0,0 +1,200 @@
+## HEADERS
+| UMIP-35     |                                                                                                                                          |
```
smb2796, line 2:
```suggestion
| UMIP-40     |                                                                                                                                          |
```

smb2796 (approved):

PR #149, opened 2021-01-21
Change price feed from Uniswap to Bitfinex

abg4:
After speaking with Sean, we realized price feeds aren't going to be able to natively support the ETH to USD normalization. We chose to submit an amendment to change the source to use the Bitfinex AMPLUSD price as it is high volume. Let me know if there are any questions regarding the proposed change.


smb2796 (approved):
Looks good to me! Just overall easier to use all USD denominated sources, so thanks for making this change.

Tommy1231232 (approved):

PR #127, opened 2021-01-14
Add Compound USDC Borrow APR as Price Identifier

evanmays:

PR #150, opened 2021-01-24
Add CAR token two-hour TWAP price identifier

evanmays:
This identifier is for creating a CAR (Compound Annualized Rate) token. CAR will be a futures token. The price identifier resolves to the trading price of CAR pre-expiry. When CAR expires, the price identifier resolves to the rolling 30 day total return.

This means a CAR token is priced at the expected APR on a specific month in the future


evanmays:
Merged into https://github.com/UMAprotocol/UMIPs/pull/127

PR #151, opened 2021-01-26
Draft for addition of uVOL-BTC-APR21 as a Price Identifier

smb2796:


UMIPs/umip-uvol-draft.md
```diff
@@ -0,0 +1,103 @@
+## Headers
+| UMIP-uVOL-DRAFT    |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add uVOL-BTC-APR21 as a price identifier              |
```
nicholaspai, line 4: Why does the `-APR21` have to be in the name of the identifier? Can't the expiration timestamp be changed by the deployer once an identifier is approved?

smb2796: Yes, but it isn't really about the expiry timestamp. It is because voters need to know exactly when they should switch from calculating the synth twap to calculating the realized volatility

nicholaspai: Doesn't that mean that future uVOL products would have to create their own identifiers? Sidenote this is where the auxillary data incoming feature will be huge.

smb2796: Yep! One of the reasons why I'm so excited about it - fewer redundant price ids. Since that feature is supposed to be in before the first uVOL contract would expire in April, the next uVOL pid would be generalized to have no pre-defined timestamps

chrismaree: yup. for now it'll need to do this extra transformation within the identifier, just like uGAS does.

UMIPs/umip-uvol-draft.md
```diff
@@ -0,0 +1,103 @@
...
+
+## SUMMARY
+
+The DVM should support price requests for the 30-day daily volatility of BTC/USD.
+
+## MOTIVATION
+
+The DVM currently does not support the 30-day day volatility of BTC/USD.
+
+The volatility of Bitcoin is a major focus in the crypto and traditional finance world.  Will the recent high volatility in Bitcoin continue throughout the year?  Is Bitcoin too volatile to be considered a viable currency or safe haven asset?  Whether volatility is considered high or low or where it will be in the future is always a debate.
```
chrismaree, line 17: nit: all the text below is full of double spaces after full stops and question marks. please fix

UMIPs/umip-uvol-draft.md
```diff
@@ -0,0 +1,103 @@
...
+
+[Here](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js) is a reference implementation for an off-chain price feed that calculates the TWAP of a token based on Uniswap price data. This feed should be used as a convenient way to query a realtime or historical price, but voters are encouraged to build their own off-chain price feeds.
+
+## TECHNICAL SPECIFICATIONS
+- Price Identifier Name: uVOL-BTC-APR21
+- Base Currency: uVOL-BTC-APR21
+- Quote currency: USDC
+- Intended Collateral Currency: USDC
+- Collateral Decimals: 0.000001 (6 decimals in more general trading format)
```
chrismaree, line 60: We normally define this just as the number of decimals in the collateral currency (i.e this would just be 6).

UMIPs/umip-uvol-draft.md
smb2796: Fixed!

UMIPs/umip-uvol-draft.md
smb2796: Also fixed. Thanks for catching this.

smb2796:
Added some more explicit instructions to the price determination methodology for the realized volatility calculation. Please let me know if this is clear enough.

nicholaspai (approved):
Awesome work LGTM

PR #156, opened 2021-02-02
Add WBTCUSD and USDWBTC as price identifiers

OpenDAO-Logan:


UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,77 @@
+## Headers
+| UMIP-#    |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add WBTCUSD, USDWBTC as price identifiers              |
+| Authors    | Logan (Logan@opendao.io)                               |
+| Status     | Draft                                                                                                                                 |
+| Created    | February 2, 2021                                                                                                                           |
```
smb2796, line 7: Can you create an UMA Discourse discussion introducing this UMIP and add another row and add another row linking to that discussion? https://discourse.umaproject.org/

smb2796:
Hi @OpenDAO-Logan! Looks good and excited for WBTC. A few weeks ago we moved to a new price identifier format. Can you edit to conform to this standard? -> https://github.com/UMAprotocol/UMIPs/blob/master/price-identifier-template.md

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,77 @@
...
+At the time of writing, WBTC is trading at $33,767.47 with a 24-hour trading volume of $346,285,873. There is a circulating supply of 117,000 WBTC coins with a max supply of 117,000. Uniswap is currently the most active market trading it. 
+
+More information on the WBTC network can be found on the website: https://wbtc.network/
+
+## Technical Specification
+The definition of this identifier should be:
+- Identifier name: WBTCUSD
+- Base Currency: WBTC
+- Quote Currency: USD
+- Exchanges: Coinbase Pro, FTX, Crypto.com
```
smb2796, line 27: These seem to be relatively low volume compared to something like the WBTC/WETH pools in Uniswap and Sushiswap. Seems to me that it would make more sense to use those, and then do an ETH/USD conversion. Wondering what other people think though.

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,79 @@
...
+
+- Rounding: Closest, 0.5 up
+- Pricing Interval: 60 seconds
+- Dispute timestamp rounding: down
+
+
+
+## Rationale
+
+Prices are primarily used by Priceless contracts to calculate a synthetic token’s redemptive value in case of liquidation or expiration. Contract counterparties also use the price index to ensure that sponsors are adequately collateralized. More complex computations (like incorporating additional exchanges, calculating a TWAP or VWAP, or imposing price bands, etc.) have the potential to add a greater level of precision and robustness to the definition of this identifier, particularly at settlement of each expiring synthetic token.
```
smb2796, line 57: I'd be interested to hear the rationale on why these price identifiers are necessary to include, instead of just using the BTCUSD and USDBTC price ids.

pemulis: Copying over from Discourse:

I feel very strongly that you cannot treat wrapped or bridged tokens the same as the native token, even though their prices stay pretty close to each other if everything is going well. All of these are separate tokens in constant flux relative to each other, which is why Curve’s DEX exists.

And, as a practical matter, there’s always a chance, however unlikely, of a catastrophic event for a particular token that causes it to trade at a discount relative to its peers. If that token is used as collateral but priced based on a different token that hasn’t experienced catastrophe, the synthetic will be broken.

Given all of that, creating a new USDWBTC/WBTCUSD price identifier feels totally appropriate. I would support a renBTCUSD/USDrenBTC price identifier pair, as well.

https://discourse.umaproject.org/t/adding-wbtc-usd-price-identifier-umip/124/3

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,79 @@
+## Headers
```
Tommy1231232, line 1: Can you use the latest price ID template? There are some questions on here that are not older versions of UMIPs

https://github.com/UMAprotocol/UMIPs/blob/master/price-identifier-template.md

OpenDAO-Logan: Yep, doing a last minute update as we speak.

UMIPs/umip-draft_wbtc_price_identifier.md
OpenDAO-Logan: Agreed!

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+The lower bound on the number of queries allowed per hour is >> 1000.
+
+
+11) What would be the cost of sending 15,000 queries?
+
+Approximately $0
+
+
+## Price Feed Implementation
+Associated WBTC price feeds are available via Cryptowat.ch, Sushiswap, and Uniswap. Conversion from WBTC/ETH to WBTC/USD price needed for Binance and Sushiswap data. No other further feeds required.
```
smb2796, line 106: Agreed on these implementation details. However this is still something that cannot be supported immediately, and would need to be built out

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Technical Specifications
```
smb2796, line 108: You should add a tech specs for USD/WBTC as well

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+
+The addition of WBTCUSD and USDWBTC will serve users at projects that are partnered with companies (such as OpenDAO) which are able to utilize the UMA architecture. More directly, this furthers adoption of the protocol by encouraging a convergence of capital from different projects and by potentially increasing TVL significantly.
+
+Currently the most liquid exchange with USD or stablecoin markets for WBTC is BKEX (17.7% volume) while most overall trade involving WBTC is on Uniswap (~30% volume). BKEX will not be used for defining the price ID, as it is not available through Cryptowatch; however, due to the large volume and high liquidity of WBTC, it should be acceptable to use other markets, even if less voluminous. WBTC can currently be traded on a wide variety of exchanges, including what are widely considered to be several top tier platforms.
+
+In the current setting, there will need to be a significant event that erodes confidence in WBTC and the token for it to be a security or PR concern.
+
+
+
+## Implementation
```
smb2796, line 137: This section should cover exactly how a voter should calculate these prices. Usually follows a pattern like:
- Hit x, y, z endpoints at the price request timestamp
- Make conversions to USD from ETHUSD rate (can refer to umip-6 for this)
- Take median
- Round to x decimals

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+
+WBTC/ETH https://api.binance.com/api/v3/ticker/price?symbol=WBTCETH
+
+Sushiswap:WBTC/ETH
+
+https://sushiswap.fi/token/0x2260fac5e5542a773aa44fbcfedf7c193bc2c599
+
+Uniswap:WBTC/USDC
+
+https://info.uniswap.org/pair/0xb4e16d0168e52d35cacd2c6185b44281ec28c9dc
```
daywiss, line 52: @smb2796 is this ok that its not linking to an API, its linking to dapp page. Same for history

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+
+Approximately $0
+
+
+## Price Feed Implementation
+Associated WBTC price feeds are available via Cryptowat.ch, Sushiswap, and Uniswap. Conversion from WBTC/ETH to WBTC/USD price needed for Binance and Sushiswap data. No other further feeds required.
+
+## Technical Specifications
+
+Price Identifier Name: WBTC/USD
```
daywiss, line 110: @smb2796 do we usually simplify these to just btc/usd or is this a case we want specifically wbtc

UMIPs/umip-draft_wbtc_price_identifier.md
OpenDAO-Logan: Will add

UMIPs/umip-draft_wbtc_price_identifier.md
smb2796: With renBTC, we just used BTCUSD. But I don't see any issue with using WBTCUSD. It's actually more correct, just some extra steps

smb2796:
As discussed on the review call and in Discourse, this is being moved to next week's voting cycle to allow for more time to determine the best price calculation methodology.

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+
+The DVM currently does not support the WBTC/USD or USD/WBTC price index. Supporting the WBTCUSD price identifier would enable UMA partners (like OpenDAO or others) to provide access to the asset via the UMA architecture.
+
+2. Please provide an example of a person interacting with a contract that uses this price identifier.
+
+The addition of this price identifier would enable DeFi users to deploy WBTC as collateral for minting a stable coin, or for other purposes. 
+
+3. Consider adding market data
+
+At the time of writing, WBTC is trading at $33,767.47 with a 24-hour trading volume of $346,285,873. There is a circulating supply of 117,000 WBTC coins with a max supply of 117,000. Uniswap is currently the most active market trading it.
```
smb2796, line 25: Can we update these numbers? WBTC has moved considerably since this was first proposed

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+2) Which specific pairs should be queried from each market?
+
+Binance: WBTC/ETH, Sushiswap WBTC/ETH, Uniswap: WBTC/USDC
+
+
+3) Provide recommended endpoints to query for real-time prices from each market listed.
+
+Binance: 
+
+WBTC/ETH https://api.binance.com/api/v3/ticker/price?symbol=WBTCETH
```
smb2796, line 44: These should really be api endpoints. For ones available on CW, you should provide the CW links similar to OCEAN/USD. For sushiswap, you could provide a subgraph query.

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+4) How often is the provided price updated?
+
+The lower bound on the price update frequency is a minute.
+
+
+5) Provide recommended endpoints to query for historical prices from each market listed.
+
+Binance: WBTC/ETH 
+
+https://api.binance.com/api/v3/klines?symbol=WBTCETH&interval=1d
```
smb2796, line 64: Same comment as above. These should be from the sources that are used in the price feed

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+3. Consider adding market data
+
+At the time of writing, WBTC is trading at $33,767.47 with a 24-hour trading volume of $346,285,873. There is a circulating supply of 117,000 WBTC coins with a max supply of 117,000. Uniswap is currently the most active market trading it.
+More information on the WBTC network can be found on the website: https://wbtc.network/
+
+## Markets and Data Sources
+
+1) What markets should the price be queried from? It is recommended to have at least 3 markets.
+
+Binance, Sushiswap, and Uniswap should be used to construct the price.These 3 exchanges comprise a significant amount of WBTC trade volume and have available pricefeeds on Cryptowatch.
```
smb2796, line 32: Sushiswap unfortunately isn't available on cryptowatch but is easily one of the top 3 most voluminous exchanges. We have a price feed being built that **may** be able to support this, but for right now we would need to build a price feed that would be able to query sushiswap and also make the normalization conversions from ETH to USD

smb2796: It looks like the price feed being written will be able to support this and nothing custom other than a configuration would need to be added. @mrice32 can you confirm this?

UMIPs/umip-draft_wbtc_price_identifier.md
mhairimca: Is this a perhaps a more general issue that needs resolved, now that we are supporting different "flavours" of key currencies, esp btc and usd?

UMIPs/umip-draft_wbtc_price_identifier.md
smb2796: Yeah, good point. Left comments to this effect

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+2. Please provide an example of a person interacting with a contract that uses this price identifier.
+
+The addition of this price identifier would enable DeFi users to deploy WBTC as collateral for minting a stable coin, or for other purposes. 
+
+3. Consider adding market data
+
+At the time of writing, WBTC is trading at $33,767.47 with a 24-hour trading volume of $346,285,873. There is a circulating supply of 117,000 WBTC coins with a max supply of 117,000. Uniswap is currently the most active market trading it.
+More information on the WBTC network can be found on the website: https://wbtc.network/
+
+## Markets and Data Sources
```
smb2796, line 28: Can you bullet all of these answers just to make it more readable?

UMIPs/umip-draft_wbtc_price_identifier.md
smb2796: We had some discussion about this in Discourse. I view it as more of a product design question - contract launchers can use whichever they feel is more fitting

UMIPs/umip-draft_wbtc_price_identifier.md
smb2796: This should be supported in the near future with a combination of the new expression price feed, and a sushiswap price feed that is being built out. Once the sushiswap feed is finished, all you would need to do is define a price feed config [here](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js).

UMIPs/umip-draft_wbtc_price_identifier.md
```diff
@@ -0,0 +1,155 @@
...
+
+1) What markets should the price be queried from? It is recommended to have at least 3 markets.
+
+Binance, Sushiswap, and Uniswap should be used to construct the price.These 3 exchanges comprise a significant amount of WBTC trade volume and have available pricefeeds on Cryptowatch. 
+
+
+2) Which specific pairs should be queried from each market?
+
+Binance: WBTC/ETH, Sushiswap WBTC/ETH, Uniswap: WBTC/USDC
```
smb2796, line 37: The highest volume by far for Binance is WBTC/BTC. Also for Uniswap, it looks like the highest volume is WBTC/ETH. I think we should use these instead.

smb2796:
@mrice32. This price feed would use WBTC/BTC from Binance (available on cryptowatch), Uniswap WBTC/ETH and Sushiswap WBTC/ETH. This would all be normalized to a WBTC/USD or USD/WBTC rate. I believe we should be able to support this with a combination of the ExpressionPriceFeed, UniswapPriceFeed, CryptowatchPriceFeed and the Sushiswap price feed being built out. Does this sound correct to you?

UMIPs/umip-draft_wbtc_price_identifier.md
chrismaree: Yup. Ideally you would want to use a median of cryptowatchx for Binance, Uniswap & Sushiswap feeds. We have never pointed our default uniswap feeds at a SushiSwap pair but in their it should work without any issues. 

To get this to work we'll need to do the following:
1) Add support to the uniswap feed to support SushiSwap.
2) add a default config, as Sean defined.

smb2796:
Outcome of the UMIP Community Review Call (Feb 16th).
- This is dependent on the UniswapPriceFeed working for Sushiswap prices. This will be tested shortly, and, if it does, the price feed should be good to go. If not, we will have to move this to next week and work on building out a Sushiswap price feed.
- Still some outstanding content issues. Mainly, I think we should switch the price sources to using WBTC/BTC for Binance and WBTC/ETH for Uniswap.
- Recommendation is to add a 1 minute TWAP specification for the Uniswap and Sushiswap prices. This will help prevent any attempted manipulation of the pool prices.

akhanubis:
There are a few broken links in the UMIP:

Sushiswap WBTC/ETH should be https://app.sushiswap.fi/pair/0xceff51756c56ceffca006cd410b03ffc46dd3a58
Uniswap:WBTC/USDC should be https://info.uniswap.org/pair/0x004375dff511095cc5a197a54140a24efef3a416 (although if you end up switching to WBTC/ETH then it's https://info.uniswap.org/pair/0xbb2b8038a1640196fbe3e38816f3e67cba72d940)

smb2796:
Tested the Uniswap price feed and it works for Sushiswap prices. If you can address the other comments, we should be good to go.

UMIPs/umip-draft_wbtc_price_identifier.md
mrice32: Is there a reason we need to have market price here? I think daily volume, supply, and most active/liquid market metrics make sense.

UMIPs/umip-draft_wbtc_price_identifier.md
mrice32: This is an api endpoint, no? I think this is 100% reasonable. Would be nice to have a cryptowatch link, too!

UMIPs/umip-draft_wbtc_price_identifier.md
mrice32: Do we have a _proposed_ way to link to a uniswap "endpoint"?

UMIPs/umip-draft_wbtc_price_identifier.md
mrice32: Totally agree -- but I think having a binance endpoint listed here is useful too, so no need to remove it... thoughts?

mrice32 (commented):
> @mrice32. This price feed would use WBTC/BTC from Binance (available on cryptowatch), Uniswap WBTC/ETH and Sushiswap WBTC/ETH. This would all be normalized to a WBTC/USD or USD/WBTC rate. I believe we should be able to support this with a combination of the ExpressionPriceFeed, UniswapPriceFeed, CryptowatchPriceFeed and the Sushiswap price feed being built out. Does this sound correct to you?

This is roughly how I see this price feed being configured:
```js
WBTCUSD : {
  type: "expression"
  expression: "median(WBTCBTC_BINANCE * BTCUSD, WBTCETH_UNISWAP * ETHUSD, WBTCETH_SUSHISWAP * ETHUSD)" ,
  customFeeds: {
    WBTCBTC_BINANCE: {
      type: "cryptowatch",
      ...
    },
    WBTCETH_UNISWAP: {
      type: "uniswap",
      ...
    },
    WBTCETH_SUSHISWAP: {
      type: "uniswap", // Note: this would be a uniswap feed pointed at a sushi address
      ...
    }
  }
},
USDWBTC: {
  type: "expression",
  expression: "1 / WBTCUSD"
}
```

smb2796:
> Do we have a proposed way to link to a uniswap "endpoint"?

We have previously used subgraph queries. Endpoint probably isn't the right word for this section and the template should be updated accordingly. From our call today, I think all that is really necessary is providing the Uniswap or Sushiswap pair address.

> Totally agree -- but I think having a binance endpoint listed here is useful too, so no need to remove it... thoughts?

Agreed, no reason to remove this.

UMIPs/umip-draft_wbtc_price_identifier.md
smb2796: No not really, but if we have them they should be accurate. I'd be in favor of removing the market price.

UMIPs/umip-draft_wbtc_price_identifier.md
smb2796: Agreed, no reason to remove this.

UMIPs/umip-draft_wbtc_price_identifier.md
smb2796: We have previously used subgraph queries. Endpoint probably isn't the right word for this section and the template should be updated accordingly. From our call today, I think all that is really necessary is providing the Uniswap or Sushiswap pair address.

UMIPs/umip-draft_wbtc_price_identifier.md
smb2796: Yes, agreed. My mistake. Would be nice to have cryptowatch as well though.

smb2796 (approved):
Going to merge in and make the suggested updates. @OpenDAO-Logan please submit a pr if any of the updates don't align with your expectations.

PR #160, opened 2021-02-03
Add DAIPHP and PHPDAI as price identifiers

schystz:


schystz:
Below is the PR for the price feeds implementation:
https://github.com/UMAprotocol/protocol/issues/2474

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,69 @@
+## Headers
+
+| UMIP-38    |                                                                                    |
+| ---------- | ---------------------------------------------------------------------------------- |
+| UMIP Title | Add DAIPHP as a price identifier                                                   |
+| Authors    | Chris Verceles (chris.verceles@halodao.com), James Orola (james.orola@halodao.com) |
+| Status     | Draft                                                                              |
+| Created    | Feb 3, 2021                                                                        |
```
smb2796, line 8: Can you create an UMA Discourse discussion introducing this UMIP and add another row and add another row linking to that discussion? https://discourse.umaproject.org/

smb2796:
Hi @schystz! Looking good. We recently moved to a new price identifier format with some additional questions. Could you edit to conform to this new standard -> https://github.com/UMAprotocol/UMIPs/blob/master/price-identifier-template.md

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,69 @@
+## Headers
+
+| UMIP-38    |                                                                                    |
```
Tommy1231232, line 3: Can you use the latest price identifier template?

https://github.com/UMAprotocol/UMIPs/blob/master/price-identifier-template.md

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,69 @@
...
+
+Ultimately, how one queries the price feeds should be varied and determined by the voter to ensure that there is no central point of failure.
+
+While it's important for tokenholders to have redundancy in their sources, bots and users that interact with the system in realtime need fast sources of price information. In these cases, it can be assumed that the price feed mean is accurate enough.
+
+Below are reference implementation for an offchain price feed based on the CoinMarketCap and CoinGecko API. These feeds should be used as a convenient way to query the price in realtime, but should not be used as a canonical source of truth for voters. Users are encouraged to build their own offchain price feeds that depend on other sources.
+
+| Price Feed Source | Implementation                                                                                                                                                  |
```
pemulis, line 60: This is a great use case! Very excited to see this in the UMIPs.

For price feeds, it's good to provide a suggested implementation so that voters can all use the same reasonable data source when calculating prices, and only need to go off-script if it looks like there is something wrong with the implementation price feed or if the data source is believed to be inaccurate or manipulated in some way.

It would also be great if the price feeds could look directly at exchange data, rather than CoinGecko and CoinMarketCap. Ideally, we could take a median of three exchange prices and use that as the market consensus, unless 2/3 (or all three) of the exchanges are considered by the voters to be off for some reason. You don't want to be _too_ vague, or tokenholders may not be able to reach consensus on the valid price, since multiple factions could have equally plausible ways of calculating slightly different prices.

If there is a specific implementation specified by the UMIP, everyone can go with that except in the hopefully rare occasion that the majority agrees the implementation is unreasonable or the data sources are broken, at which point they can do the hard work of agreeing on an alternative implementation that satisfies the majority.

schystz:
> Hi @schystz! Looking good. We recently moved to a new price identifier format with some additional questions. Could you edit to conform to this new standard -> https://github.com/UMAprotocol/UMIPs/blob/master/price-identifier-template.md

Hi @smb2796 , thanks for the feedback so far. We have updated the template and is now on the process of filling out the answers to the rest of the questions.

UMIPs/umip-draft_add_daiphpid.md
schystz: Updated :)

UMIPs/umip-draft_add_daiphpid.md
ctverceles: Thanks for the feedback @pemulis! we actually are in the process of adding in a DAI:PHP endpoint from a Philippine exchange and will add it in the moment its available - however they are the only ones who have an exposed endpoint (that we asked them specifically for) AND have decent volume to back that price up. Other exchanges simply don't offer the endpoint access freely at the moment (we're actually working on getting the others to open it right now, but would like DAIPHP on mainnet to get going for our initial product)

UMIPs/umip-draft_add_daiphpid.md
ctverceles: done @smb2796 ! https://discourse.umaproject.org/t/adding-dai-php-price-identifier/135

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,204 @@
...
+    - CMC: `https://pro-api.coinmarketcap.com/v1/cryptocurrency/quotes/latest?symbol=DAI&convert=PHP&CMC_PRO_API_KEY=<free tier api key>`
+
+    - CG: `https://api.coingecko.com/api/v3/simple/token_price/ethereum?contract_addresses=0x6b175474e89094c44da98b954eedeac495271d0f&vs_currencies=php`
+    
+    - Bloom Exchange DAI:PHP endpoint being opened up
+    
+4. How often is the provided price updated?
+
+    - CMC: every 60 seconds
+    - CG: unknown
```
smb2796, line 62: This seems problematic. We should at least try to get an estimation on how frequently this price updates

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,204 @@
...
+
+6.  Do these sources allow for querying up to 74 hours of historical data? 
+
+    - CMC: Yes
+    - CG: Yes
+
+7.  How often is the provided price updated?
+
+    - CMC: every 5 minutes
+    - CG: unknown
```
smb2796, line 80: Same here. Would be good to get an update estimation

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,204 @@
...
+
+# TECHNICAL SPECIFICATIONS
+
+**1. Price Identifier Name** - DAIPHP
+
+**2. Base Currency** - DAI
+
+**3. Quote currency** - PHP
+
+- If your price identifier is a currency pair, your quote currency will be the
```
smb2796, line 125: You can remove questions that do not apply

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,204 @@
...
+
+- Please be aware that the value of any UMA synthetic token is the value of the price identifier in units of the collateral currency used. If a contract’s price identifier returns 1, and is collateralized in renBTC, each synthetic will be worth 1 renBTC. In most cases, the value of your quote currency and intended collateral currency should be equal.
+
+- [Response - if applicable]
+
+**4. Intended Collateral Currency** - DAI
+
+- Does the value of this collateral currency match the standalone value of the listed quote currency? 
+
+    - [ANSWER]
```
smb2796, line 137: Yes it does

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,196 @@
...
+    - CMC: `https://pro-api.coinmarketcap.com/v1/cryptocurrency/quotes/latest?symbol=DAI&convert=PHP&CMC_PRO_API_KEY=<free tier api key>`
+
+    - CG: `https://api.coingecko.com/api/v3/simple/token_price/ethereum?contract_addresses=0x6b175474e89094c44da98b954eedeac495271d0f&vs_currencies=php`
+    
+    - Bloom Exchange DAI:PHP endpoint being opened up
+    
+4. How often is the provided price updated?
+
+    - CMC: every 60 seconds
+    - CG: according to CG, price updated every 1 to 10 minutes [FAQ #6](https://www.coingecko.com/id/faq)
```
smb2796, line 62: All of this data should be added for Bloom as well

ctverceles: Yes thanks will add it in once they open up the endpoint any day now. It'll appear here https://www.bloomremit.net/developers/2/exchange_rates

smb2796:
As discussed on the review call and in Discourse, this is being moved to next week's voting cycle to allow for more time to integrate the Bloom api and finalize the price feed.

ctverceles:
A bit of an update here: Bloom hasn't gotten back to us on that endpoint, so we're ok to just go with CG and CMC as one of the options that was discussed as which is already implemented here https://github.com/UMAprotocol/protocol/pull/2480

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,204 @@
...
+
+1. What are the financial positions enabled by creating this synthetic that do not already exist?
+
+    - Supporting the DAIPHP price identifier would enable the creation of a Philippine Peso stablecoin (UBE), backed by DAI. Token minters could go short on the DAI/PHP index, while token holders could go long or use synthetic PHP for functional purposes.
+
+2. Please provide an example of a person interacting with a contract that uses this price identifier. 
+
+    - in trading pairs on Philippine cryptocurrency exchanges
+
+    - basis for on chain, on demand liquidity in cross border remittance (our team is starting with the Singapore Philippine corridor with ZKSync, Argent and various on/off ramps)
```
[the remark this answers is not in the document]
ctverceles: Done! @daywiss

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,191 @@
...
+10. If there is a free tier available, how many queries does it allow for?
+
+    - CMC: Yes, 10K call credits /mo
+    - CG: Yes, 100 requests/minute
+
+11.  What would be the cost of sending 15,000 queries?
+
+     - CMC: $29/ mo for Hobbyist plan which bumps call credits to 40K /mo
+     - CG: Free!
+
```
mhairimca, line 99: Is there any cost associated with adding Bloom as the third data source?

UMIPs/umip-draft_add_daiphpid.md
mhairimca: Is the intention to add Bloom as a third price source?

smb2796: That was the original intention, but sounds like we are moving away from that.

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,191 @@
...
+| UMIP [#]          |                                                                                    |
+|-------------------|------------------------------------------------------------------------------------|
+| UMIP Title        | Add DAIPHP as a price identifier                                                   |
+| Authors           | Chris Verceles (chris.verceles@halodao.com), James Orola (james.orola@halodao.com) |
+| Status            | Draft                                                                              |
+| Created           | Feb 3, 2021                                                                        |
+| Link to Discourse | [Discourse](https://discourse.umaproject.org/t/adding-dai-php-price-identifier/135)                                                                             |
+
+<br />
```
smb2796, line 11:
```suggestion
```

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,191 @@
...
+
+# SUMMARY 
+
+The DVM should support price requests for the DAI/PHP price index.
+
+<br />
```
smb2796, line 17:
```suggestion
```

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,191 @@
...
+
+# MOTIVATION
+
+This section should clearly explain the types of financial products that will be created and the mechanics of an example financial product using this price identifier. Please answer the following questions:
+
+1. What are the financial positions enabled by creating this synthetic that do not already exist?
+
+    - Supporting the DAIPHP price identifier would enable the creation of a Philippine Peso stablecoin (UBE), backed by DAI. Token minters could go short on the DAI/PHP index, while token holders could go long or use synthetic PHP for functional purposes.
```
smb2796, line 25: Why are you using the DAIPHP rate and not the PHPDAI rate? If the intention is to create a synthetic PHP stablecoin, I believe you would need PHPDAI.

With DAIPHP, 1 DAI = 48 PHP -> your price is ~48. This means that if your collateral was DAI, each synth that you mint would be worth 48 DAI, not 48 PHP. Essentially each synth would be worth 48 * 48 (2304) PHP. 

With PHPDAI, 1 PHP = .021 DAI -> your price is .021. Using DAI as collateral, each synth would be worth .021 DAI or the equivalent of 1 PHP.

ctverceles: Thanks for detailing that @smb2796. We sorta took https://github.com/HaloDAO/UMIPs/blob/umip-dai-php-price-id/UMIPs/umip-19.md wording as template since they were also proposing a fiat synthetic and then planned on using a liquidator `PRICE_FEED_CONFIG` config with `invertPrice` property set to `true` to get the PHP:DAI rate (for example, DAI:PHP = 48 and 1/48 = 0.21 DAI). Were we thinking along the correct lines there?

smb2796: Yes, you are thinking correctly for the price feed. You will just need to propose the actual price identifier that you want to use in the UMIP. So even if you are querying for the DAI:PHP rate, but are planning to use the PHP:DAI rate, you will need to propose PHP:DAI and specify in the implementation section that voters should take the inverse of DAI:PHP.

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,199 @@
...
+The DVM should support price requests for the DAIPHP and PHPDAI price index.
+
+# MOTIVATION
+
+The DVM currently does not support the DAIPHP or PHPDAI price index. Supporting the DAIPHP and PHPDAI price identifiers would enable the creation of a synthetic DAIPHP and PHPDAI stablecoin, backed by DAI as collateral. Token minters could go short on DAI/PHP index, while token holders could go long or use synthetic PHP for functional purposes.
+
+Examples of a person interacting with a contract that uses this price identifier would be;
+
+    - in trading pairs on Philippine cryptocurrency exchanges
+
```
chrismaree, line 22:
```suggestion
```

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,199 @@
...
+   - CoinGecko
+
+2. Which specific pairs should be queried from each market?
+
+   - DAI:PHP
+
+3. Provide recommended endpoints to query for real-time prices from each market listed.
+
+   - CMC: `https://pro-api.coinmarketcap.com/v1/cryptocurrency/quotes/latest?symbol=DAI&convert=PHP&CMC_PRO_API_KEY=<free tier api key>`
+
```
chrismaree, line 45:
```suggestion
```

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,199 @@
...
+<br />
+
+# PRICE FEED IMPLEMENTATION
+
+Link to the UMA protocol PR:
+
+- https://github.com/UMAprotocol/protocol/pull/2480
+
+Link to the price feed pull request.
```
chrismaree, line 99: your link goes to an issue, not a feed.
```suggestion
Link to the price feed pull issue:
```

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,199 @@
...
+    - CMC: $29/ mo for Hobbyist plan which bumps call credits to 40K /mo
+    - CG: Free!
+
+<br />
+
+# PRICE FEED IMPLEMENTATION
+
+Link to the UMA protocol PR:
+
+- https://github.com/UMAprotocol/protocol/pull/2480
```
chrismaree, line 97: These feeds are almost ready to merge (95% of the way there!). just need 2 additional unit tests and sign off from @nicholaspai

schystz: Thanks for the latest feedback @chrismaree! We have already added the unit tests and committed the suggested changes for this PR as well. Looking forward to the review call later :)

UMIPs/umip-draft_add_daiphpid.md
```diff
@@ -0,0 +1,197 @@
...
+
+- CG and CMC provide a readily available methodology for consolidating and validating price data between a particular crypto asset and fiat quote across exchanges worldwide (see https://www.coingecko.com/en/methodology and https://coinmarketcap.com/api/faq/ )
+
+Our initial approach in lieu of not having access to a "raw" exchange based DAI:PHP or PHP:DAI pair was to source and consolidate component price feeds of DAI:USD and USD:PHP, but it seems CG already does this with their [OpenExchangeRates](https://openexchangerates.org/) integration. In the future, the [HaloDAO](https://halodao.com/) team (or any other team making use of this) will add more price feed sources as the DAI - PHP market builds volume and listings (which we are pushing them to do).
+
+Additionally, both CG and CMC sources offer free and publicly accessible DAI:PHP endpoints, which we would then invert in the price feed implementation to arrive at the PHP:DAI rate.
+
+<br>
+
+# IMPLEMENTATION
```
smb2796, line 169: This implementation section should be a step by step guide to voting. Something like:

1. Voters should query these two endpoints at the price request timestamp rounded down to the nearest available timestamp
2. Voters should take the median of the results
3. Round to x decimal places

smb2796 (approved):
Going to merge in and make a few small formatting fixes on my side.

PR #168, opened 2021-02-10
UMIP draft - YAM ETH, USD price iden.

rossgalloway:
initial pull request to add YAM/ETH and YAM/USD price identifiers


UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,312 @@
+
+## HEADERS
+| UMIP [xx]     |                                                                                                                                  |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add YAM-ETH, YAM-USD, and USD-YAM as price identifiers]                                                                                                  |
```
smb2796, line 5: Always best when we don't need to use punctuation in price id names. What do you think about YAMETH, YAMUSD and USDYAM?

smb2796: Also you list ETHYAM below, but not here. Please include in the title as well

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,312 @@
...
+| Authors    | Ross (ross@yam.finance)
+| Status     | Draft                                                                                                                                  |
+| Created    | January 28th, 2021  
+| Forum Post | https://discourse.umaproject.org/t/add-yam-usd-and-yam-eth-price-identifiers/171
+
+//Note that an UMIP number will be assigned by an editor. When opening a pull request to submit your UMIP, please use an abbreviated title in the filename, umip-draft_title_abbrev.md.
```
smb2796, line 11: You can remove this

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,312 @@
...
+//The title should be 44 characters or less.
```
smb2796, line 12: And this

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,312 @@
...
+ETH/YAM 
+YAM/USD
+USD/YAM 
+
+
+# MOTIVATION
+
+The DVM currently does not support the YAMETH, ETHYAM, YAMUSD or USDYAM price indices.
+
+Supporting the YAMUSD price identifier would enable the creation of a YAM backed stablecoin, built using one of the perpetual stablecoin frameworks being developed on UMA. YAM token holders can utilize this as a hedging tool, and could go long or use it for other financial purposes. There is also potential for the YAM DAO itself to use YAM reserves or newly minted tokens as a community backed line of credit to mint YAM backed stablecoins to support its mission or to add liquidity to UMA synths supported on Degenerative.finance.
```
smb2796, line 28: This really only applies to USDYAM, right? YAMUSD would be more useful for something like synthetic YAM

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,312 @@
...
+
+    - No limit currently
+
+11.  What would be the cost of sending 15,000 queries?
+
+     - $0
+
+# PRICE FEED IMPLEMENTATION
+
+//To allow for the creation of bots that can programmatically calculate prices off-chain to liquidate and dispute transactions, you must create a price feed following the UMA Protocol format (outlined below). This price feed is also necessary to calculate developer mining rewards.
```
smb2796, line 98: I'm assuming this is just a filler until the price feed gets figured out. Are you planning on submitting a pr for a price feed?

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,312 @@
...
+1. What markets should the price be queried from? It is recommended to have at least 3 markets.
+
+Sushiswap - https://sushiswap.fi/pair/0x0f82e57804d0b1f6fab2370a43dcfad3c7cb239c
+Huobi  
+Gate.io  
+Uniswap - https://info.uniswap.org/token/0x0aacfbec6a24756c20d41914f2caba817c0d8521 
+
+2.  Which specific pairs should be queried from each market?
+
+YAM/ETH on sushiswap and uniswap
```
smb2796, line 53: Curious why you chose four sources here? Huobi seems to be much higher volume than Gate.io and is available on Cryptowatch, so I'd suggest removing Gate.io.

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,312 @@
...
+
+More information on YAM can be found on the website: https://yam.finance/
+
+# MARKETS & DATA SOURCES
+
+ **Required questions**
+
+1. What markets should the price be queried from? It is recommended to have at least 3 markets.
+
+Sushiswap - https://sushiswap.fi/pair/0x0f82e57804d0b1f6fab2370a43dcfad3c7cb239c
```
smb2796, line 46: No need to link to the interfaces

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,312 @@
...
+2.  Which specific pairs should be queried from each market?
+
+YAM/ETH on sushiswap and uniswap
+YAM/USDT on Huobi
+YAM/USDT on Gate
+ETH/USD per UMIP 6
+
+2. Provide recommended endpoints to query for real-time prices from each market listed. 
+
+    - Sushiswap graph explorer
```
smb2796, line 60: Can you provide the actual queries? Should be similar to what is shown here: https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-39.md#obtaining-the-wbtcusd-price

UMIPs/YAMUSDETH Price iden umip
rossgalloway: Yeah, honestly I am not sure what YAM/USD would be used for? Short Token?  Just figured I would include it since it is simple to do so.

UMIPs/YAMUSDETH Price iden umip
rossgalloway: uniswap is the one with the least liquidity and volume, but I could take gate out.

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,312 @@
...
+//To meet this criteria, please include a link to a PR to the UMA [protocol repo](https://github.com/UMAprotocol/protocol) with an example price feed that inherits this [PriceFeedInterface](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/PriceFeedInterface.js). 
+
+//Classes of price identifiers that are supported by default and require no additional price feed include.
+
+//- Any currency pair available on [Cryptowatch](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js)
+
+//- [Uniswap prices](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js)
+
+//- [Balancer prices](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js)
+
```
mhairimca, line 109: Do we need actual price feed links?

smb2796: Yes, definitely. I believe this is just left over as a filler, because this section still needs to be completed. @rossgalloway can you remove this to avoid confusion?

UMIPs/YAMUSDETH Price iden umip
smb2796: Fair, yeah Uniswap looks to be really low vol so let's remove that instead and leave Gate

UMIPs/YAMUSDETH Price iden umip
smb2796: Up to you all if you want to include. Could be used to short, but I'm not sure if that is something that is actually needed

UMIPs/umip-draft-yam-collateral.md
```diff
@@ -0,0 +1,51 @@
+## Headers
+| UMIP-38  |                                                                                                                                         |
```
smb2796, line 2: Can you delete this file from your pull request?

UMIPs/newUmipYamEthSlpCollateral
```diff
@@ -0,0 +1,48 @@
+## Headers
+| UMIP-xx  |                                                                                                                                         |
```
smb2796, line 2: This will need to be in a separate pull request and won't be included for this week's votes

rossgalloway: oh, shit I borked something when playing around in my clone.  My bad. I'll fix this

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,289 @@
...
+**For YAM/ETH and ETH/YAM **
+
+1. **What prices should be queried for and from which markets?**
+
+    - YAM/ETH Price from Sushiswap (https://app.sushiswap.fi/pair/0x0f82e57804d0b1f6fab2370a43dcfad3c7cb239c)
+
+    - ETH/YAM Price from Sushiswap (https://app.sushiswap.fi/pair/0x0f82e57804d0b1f6fab2370a43dcfad3c7cb239c)
+    
+    //should we add YAM/ETH from Huobi and gate.io even though they have pretty shitty volume? or calculate the YAM/ETH from YAM/USDT on those exchanges?
```
smb2796, line 215: I'd be in favor of calculating YAM/ETH from YAM/USDT. You could just do the conversion with the USD/ETH rate.

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,289 @@
...
+
+    - None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+
+4. **Result processing** 
+
+      - Round to 6 decimals
+
+<br>
+
+**For USD/YAM:**
```
smb2796, line 257: You can remove this section and just have USD/YAM and YAM/USD together.

UMIPs/YAMUSDETH Price iden umip
rossgalloway: So to clarify, we would get the final YAM/ETH price using the native sushiswap price and then the calculated prices from huobi and gate, going from YAM/USDT to YAM/ETH?

And then do the opposite in getting YAM/USD. calculate from sushiswap ETH/YAM and use native prices from the CEXs?

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,289 @@
+
+## HEADERS
+| UMIP [xx]     |                                                                                                                                  |
```
smb2796, line 3: This should be a markdown file. Right now all of the formatting is messed up because it isn't markdown

UMIPs/YAMUSDETH Price iden umip
```diff
@@ -0,0 +1,289 @@
...
+Gate.io  
+
+2.  Which specific pairs should be queried from each market?
+
+YAM/ETH on sushiswap 
+YAM/USDT on Huobi
+YAM/USDT on Gate
+ETH/USD per UMIP 6
+
+2. Provide recommended endpoints to query for real-time prices from each market listed.
```
smb2796, line 54:
```suggestion
3. Provide recommended endpoints to query for real-time prices from each market listed. 
```

UMIPs/YAMUSDETH Price iden umip
smb2796: bump

rossgalloway: These aren't ready yet. I will update with what I have but am not confident that they are correct. These, along with the price implimentation are being worked on my other members of the team

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,301 @@
...
+
+    - Uniswap graph explorer
+    https://thegraph.com/explorer/subgraph/uniswap/uniswap-v2
+
+
+5. Provide recommended endpoints to query for historical prices from each market listed. 
+
+    * YAM: Returns ETH/YAM price
+
+        ```{
```
smb2796, line 67: This formatting should be corrected. Each expression should be wrapped in its own code block, like this:

```
{
  token(
    id:"0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48", 
    block: {number: 11849560}
  )
  {
    derivedETH
  }
} 
```

UMIPs/YAMUSDETH Price iden umip
smb2796: I don't think any work needs to be done on the price implementation. This should be able to be supported by a new price feed that we just added, called the [ExpressionPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js). The `ExpressionPriceFeed` allows you to combine price feeds in mathematical expressions. This should make any USD/ETH + YAM/ETH combination relatively easy.

Since you are using a Sushiswap rate, this is also dependent on a Sushiswap price feed, but this is also something that we are building and should be ready soon.

Once these are ready, all you should need to do is define a price feed config [here](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js).

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,301 @@
...
+
+Sushiswap 
+Huobi  
+Gate.io  
+
+2.  Which specific pairs should be queried from each market?
+
+YAM/ETH on sushiswap 
+YAM/USDT on Huobi
+YAM/USDT on Gate
```
smb2796, line 51: Seems like Gate.io is really low volume. I'd say it's more detrimental to include this than not. What about only using the Sushiswap YAM/ETH price and the Huobi YAM/USDT price?

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,313 @@
...
+
+    - None. Human intervention in extreme circumstances where the result differs from broad market consensus.
+
+4. **Result processing** 
+
+      - Round to 6 decimals
+
+<br>
+
+**For USD/YAM:**
```
smb2796, line 281: You can erase this and just add USD/YAM to the header with YAM/USD.

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,313 @@
...
+The addition of YAMETH, ETHYAM, YAMUSD, and USDYAM fits into a larger goal of advancing the adoption of the UMA protocol by allowing YAM to be used as collateral for minting derivatives on the Degenerative platform, as well as other UMA based synthetics. This furthers adoption of the protocol by encouraging a convergence of capital from different projects and increasing TVL.
+
+Using the Sushiswap Price should give the most accurate price for YAM/ETH on the market as it has the deepest liquidity. All AMM pools should be queried using a 2-hour TWAP to prevent flash-loan attacks and liquidations.
+
+The YAM/USDT pools on Huobi and Gate.io have the most volume. Much more than YAM/ETH.
+
+
+# IMPLEMENTATION
+
+**For YAM/ETH and ETH/YAM **
```
smb2796, line 231: Can you provide instructions, similar to what you have for YAM/USD?

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,313 @@
...
+**For YAM/USD:** 
+
+1. **What prices should be queried for and from which markets?**
+
+ YAM/ETH Price from Sushiswap (https://sushiswap.fi/pair/0x0f82e57804d0b1f6fab2370a43dcfad3c7cb239c)
+Query the ETH/USD Price as per UMIP-6
+Multiply the YAM/ETH price by the ETH/USD price to get the YAM/USD price
+Query the YAM/USDT price on Gate.io
+Query the YAM/USDT Price on Huobi 
+Take the median of prices acquired from steps 3-5 to get the final YAM/USD price
```
smb2796, line 264: Add rounding to this, and the step below for each price identifier,

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,313 @@
...
+
+      - When more approved exchanges and price feeds are added, results should be calculated using the median price.
+ 
+<br>
+
+**For YAM/USD:** 
+
+1. **What prices should be queried for and from which markets?**
+
+ YAM/ETH Price from Sushiswap (https://sushiswap.fi/pair/0x0f82e57804d0b1f6fab2370a43dcfad3c7cb239c)
```
smb2796, line 259: Can you number these steps?

smb2796:
@mrice32. This price feed would use YAMETH from Sushiswap, and YAMUSDT from Huobi (available on cryptowatch). I believe we should be able to support this with a combination of ExpressionPriceFeed, CryptowatchPriceFeed and the Sushiswap price feed being built out. Does this sound correct to you?

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,313 @@
+
+## HEADERS
+| UMIP [xx]     |                                                                                                                                  |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add YAMETH, ETHYAM, YAMUSD, and USDYAM as price identifiers]                                                                                                  |
```
chrismaree, line 5:
```suggestion
| UMIP Title | Add YAMETH, ETHYAM, YAMUSD, and USDYAM as price identifiers                                                                                                  |
```

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,313 @@
...
+USD/YAM 
+
+
+# MOTIVATION
+
+The DVM currently does not support the YAMETH, ETHYAM, YAMUSD or USDYAM price indices.
+
+Supporting the YAMUSD price identifier would enable the creation of a YAM backed stablecoin, built using one of the perpetual stablecoin frameworks being developed on UMA. YAM token holders can utilize this as a hedging tool, and could go long or use it for other financial purposes. There is also potential for the YAM DAO itself to use YAM reserves or newly minted tokens as a community backed line of credit to mint YAM backed stablecoins to support its mission or to add liquidity to UMA synths supported on Degenerative.finance. 
+
+A user would lock YAM in order to mint a new overcollateralized dollar-pegged token.  This price feed would be used to determine the liquidation price of the collateral. If the Perpetual token has a funding rate, the price feed would also be used to determine it and would dictate whether there is a premium on repayment to return the YAM collateral.
```
chrismaree, line 27:
```suggestion
A user would lock YAM in order to mint a new overcollateralized dollar-pegged token. This price feed would be used to determine the liquidation price of the collateral. If the Perpetual token has a funding rate, the price feed would also be used to determine it and would dictate whether there is a premium on repayment to return the YAM collateral.
```

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,313 @@
...
+
+    - No limit currently
+
+11.  What would be the cost of sending 15,000 queries?
+
+     - $0
+
+# PRICE FEED IMPLEMENTATION
+
+In progress
```
chrismaree, line 117: seems like cryptowatch already has Huobi, Gate.io & Uniswap feeds for YAM. 
image

To finish up your feed we'll need to add a new feed for SushiSwap which can then be used within a median with the other feeds. Fortunately, there is another UMIP currently in review (https://github.com/UMAprotocol/UMIPs/pull/156/files for WBTCUSD & USDWBTC) which also requires a SushiSwap feed. Ideally, both of these UMIPs can rely on the same feed implementation.

You'll then simply need to define a [default config](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js) to Medianizer over the feeds defined within your implementation.

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,313 @@
...
+2.  Which specific pairs should be queried from each market?
+
+YAM/ETH on sushiswap 
+YAM/USDT on Huobi
+YAM/USDT on Gate
+ETH/USD per UMIP 6
+
+3. Provide recommended endpoints to query for real-time prices from each market listed. 
+
+    - Sushiswap graph explorer
```
Tommy1231232, line 56: Instead of using the graph we should pull the price from the source (i.e., markets). Can you update?

smb2796: This is on-chain data, so you the graph is just a way of getting this information, similarly to how cryptowatch is used to get CEX prices. We've been putting subgraph queries in here in previous UMIPS, but talked today about removing these entirely and just defining the pool addresses within the UMIP.

Sorry for the back and forth @rossgalloway, but can you just remove the subgraph queries and instead just list the pool addresses?

smb2796:
Outcome of the community call today was that:
- This is dependent on the UniswapPriceFeed working for Sushiswap prices. This will be tested shortly, and, if it does, the price feed should be good to go. If not, we will have to move this to next week and work on building out a Sushiswap price feed.
- Recommendation is to add a 1 minute TWAP to the Sushiswap price to prevent manipulation.
- A couple of outstanding comments on content structuring.

Overall should be a go if the UniswapPriceFeed works for Sushiswap.

UMIPs/YAMUSDETH Price iden umip.md
```diff
@@ -0,0 +1,313 @@
...
+            token(
+                id:"0x0aacfbec6a24756c20d41914f2caba817c0d8521", 
+                block: {number: 11849560}
+            )
+            {
+                derivedETH
+            }
+        } 
+
+    * USDC: Returns ETH/USD price. Divide ETH/YAM by ETH/USD to get USD/YAM
```
smb2796, line 78: This isn't necessary, since we are using ETH/USD per UMIP 6

smb2796:
Tested querying for the Sushiswap price with the UniswapPriceFeed. Everything works correctly. If you can address the other outstanding comments, this should be good to go.

UMIPs/YAMUSDETH Price iden umip
rossgalloway: Per discussion with Sean, uniswap has been added back in since gate.io has almost no liquidity in YAM/ETH and YAM/USDT does not allow withdrawals and is not available on cryptowatch

smb2796 (approved):
Thanks for making those changes. Going to merge in and make a few formatting changes, but overall looks pretty good.

PR #170, opened 2021-02-11
Add USD UNI-V2 LP token price identifiers

dev-1-lp-dollar:
This is a proposal to add USD-UNI-V2-WBTC-ETH as a supported price identifier.

The price identifier will allow LP Dollar to offer fixed borrowing costs to UNI-V2-WBTC-ETH liquidity providers, and serve as a tool for the UMA ecosystem to create synthetics with UNI-V2-WBTC-ETH LP positions as collateral.


smb2796:
Seems like all of the price identifiers that you just submitted could be housed in one price id umip. Same for all of the collateral currencies.

dev-1-lp-dollar:
The trouble is that each pair has different underlying oracle requirements (for example, getting the WBTC vs the UMA price). But I can combine all of them if you would like.

smb2796:
>The trouble is that each pair has different underlying oracle requirements (for example, getting the WBTC vs the UMA price). But I can combine all of them if you would like.

That would be best, and then just specifying the different requirements for each within the same UMIP

dev-1-lp-dollar:
> > The trouble is that each pair has different underlying oracle requirements (for example, getting the WBTC vs the UMA price). But I can combine all of them if you would like.
> 
> That would be best, and then just specifying the different requirements for each within the same UMIP

Combined them here and for the collateral UMIP.

smb2796:
Awesome, thanks for doing this. Because the UMIP submissions missed the weekly submission deadline, we unfortunately need to move these to next week's review cycle. The great thing about this is that we are already building out a price feed that should work for these! The price feed should be finished by next week, so we will be setup well to vote in the following cycle.

dev-1-lp-dollar:
> Awesome, thanks for doing this. Because the UMIP submissions missed the weekly submission deadline, we unfortunately need to move these to next week's review cycle. The great thing about this is that we are already building out a price feed that should work for these! The price feed should be finished by next week, so we will be setup well to vote in the following cycle.

Yeah, no problem, thanks for the feedback. Regarding timeline, meaning it will be discussed on Monday and voted on next Friday? Not too familiar with the cycles.

dev-1-lp-dollar:
According to the calendar, as far as I can tell, it looks like the PR was opened on Wednesday before the deadline to meet the review on Monday.

dev-1-lp-dollar:
Yeah it looks like I posted in Discord community and Discourse in time but took my time PRing it in here. :/

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+  
+    = (1 / 520.3440264478901) * 1e18 
+  
+    = 1921805477092654  
+    ```
+
+
+## Technical Specifications
+
+Price Identifier Name: USD-UNI-V2-WBTC-ETH
```
Tommy1231232, line 727: For all the names lets update to USD/UNI-V2-WBTC-ETH. (this makes it easier to distinguish the base and quote currency)

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+| Created    | February 09, 2021           |
+| Link to Discourse | [Discourse](https://discourse.umaproject.org/t/add-usd-uni-v2-wbtc-eth-as-a-price-identifier/148)                                 |
+
+## Summary
+
+The DVM should support price requests for the following Uniswap V2 LP tokens: WBTC-ETH, USDC-ETH, UNI-ETH, UMA-ETH.
+
+## Motivation
+
+The DVM currently does not support Uniswap V2 LP tokens.
```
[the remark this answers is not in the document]
dev-2-lp-dollar: Correct from my understanding. V1 or V3 would be separate liquidity pools from what this price ID is expected to reference for V2. 

Because they're separate liquidity pools, they have a separate exchange rate, and therefore can't be treated equal. It's reasonable to expect the prices across these different pools (of the same underlying assets) to be quite close due to the nature of arbitrage. However, I wouldn't expect them to be equal. 

Let me know if this line of thought makes sense, or if you had a different POV please!

UMIPs/umip-usd-uni-v2-lp.md
dev-2-lp-dollar: @Tommy1231232 good idea :+1: , will do.

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+
+By enabling LP token price identifiers, UMA will open the door for the creation of synths with CFMM (constant function market maker) LP positions.
+
+The LP Dollar team will use the price identifiers to enable fixed borrowing costs for Uniswap V2 liquidity providers.
+
+## Data Sources & Price Feed Implementation
```
Reinis-FRP, line 22: Can you split this out to `MARKETS & DATA SOURCES`, `PRICE FEED IMPLEMENTATION` and `IMPLEMENTATION` sections?
Do you have a link to reference implementation script that combines all steps described in this section?

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+Rounding: Round to nearest 18 decimal places (nineteenth decimal place digit >= 5 rounds up and < 5 rounds down)  
+
+Price Identifier Name: USD-UNI-V2-UMA-ETH  
+Base Currency: USD  
+Quote currency: UNI-V2-UMA-ETH  
+Intended Collateral Currency: UNI-V2-UMA-ETH  
+Collateral Decimals: 18  
+Rounding: Round to nearest 18 decimal places (nineteenth decimal place digit >= 5 rounds up and < 5 rounds down)
+
+## Rationale
```
Reinis-FRP, line 755: Can you elaborate the rationale section on the design motivation (e.g. reasoning for selecting particular markets, price robustness, potential for price manipulation)?

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+
+## Security Considerations
+Adding these new price identifiers should not effect the security to the DVM or priceless financial contract users. However, anyone deploying a new priceless token contract referencing these identifiers should take care to parameterize the contract appropriately to avoid the loss of funds for synthetic token holders. Additionally, the contract deployer should ensure that there is a network of liquidators and disputers ready to perform the services necessary to keep the contract solvent.
+
+The Uniswap smart contracts are among the highest quality in decentralized finance, safely locking up and handling billions of dollars in different assets.
+
+Uniswap as a price feed is online 24/7 with no downtime as it's a protocol hosted on the Ethereum blockchain. Therefore, there will be no forced existence of time gaps in price data - it will always be available.
+
+These pairs are susceptible to price volatility, and have moved in the ranges of 10-30% within 24 hours, with the more extreme side of the range highly unlikely, but still possible. A high collateralization requirement (120%+) should be set to mitigate this potential volatility risk.
```
mhairimca, line 766: Are there any risks to manipulation of the balance of the assets within a Uniswap pool using flashloans?

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+
+    Base Currency: UNI  
+    Quote Currency: USD  
+
+    Exchanges: Coinbase Pro (UNI:USD), Kraken (UNI:USD), Bitfinex (UNI:USD)
+    Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.  
+    Price Steps: 0.01 (2 decimals in more general trading format)  
+    Rounding: Closest, 0.5 up  
+    Pricing Interval: 60 seconds  
+    Dispute timestamp rounding: down
```
akhanubis, line 503: Can you add Result Processing: Median to make it clearer? Also applies to UMA:USD.

akhanubis (commented):
There are some extra `)` in the price calculation examples

UMIPs/umip-usd-uni-v2-lp.md
mhairimca: Should the decimals here be 18 to match the collateral currency?
@smb2796  - could you clarify this as I'm not sure.

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
pemulis: > Correct from my understanding. V1 or V3 would be separate liquidity pools from what this price ID is expected to reference for V2.
> 
> Because they're separate liquidity pools, they have a separate exchange rate, and therefore can't be treated equal. It's reasonable to expect the prices across these different pools (of the same underlying assets) to be quite close due to the nature of arbitrage. However, I wouldn't expect them to be equal.
> 
> Let me know if this line of thought makes sense, or if you had a different POV please!

Agreed!

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+    ```
+
+4) Fourth, use the UNI-V2-USDC-ETH total supply of LP tokens and USD value of reserves to calculate the USD value of each LP token (UNI-V2-USDC-ETH:USD). USDC is treated as equal to USD.
+
+    Example calculation of UNI-V2-USDC-ETH:USD:
+
+    ```
+    UNI-V2-USDC-ETH:USD =
+  
+    (USD value of USDC reserves + USD value of WETH reserves) / UNI-V2-USDC-ETH total supply of LP tokens)
```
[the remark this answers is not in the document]
pemulis: I think it makes sense to include that additional step, since USDC is almost always slightly off-peg in one direction or the other, and the effect is amplified by pooling a bunch of USDC together.

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+  
+    = (1 / 520.3440264478901) * 1e18 
+  
+    = 1921805477092654  
+    ```
+
+
+## Technical Specifications
+
+Price Identifier Name: USD/UNI-V2-WBTC-ETH
```
pemulis, line 727: @mrice32 would the `/` in the price identifier cause any problems with our bot clients, or would it be escaped automatically? Cf. the expression price feed.

UMIPs/umip-usd-uni-v2-lp.md
pemulis: I think that the query will return the balance of the assets at the beginning of the specified block, before any transactions, but I could be wrong.

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+
+The DVM should support price requests for the following Uniswap V2 LP tokens: WBTC-ETH, USDC-ETH, UNI-ETH, UMA-ETH.
+
+## Motivation
+
+The DVM currently does not support Uniswap V2 LP tokens.
+
+By enabling LP token price identifiers, UMA will open the door for the creation of synths with CFMM (constant function market maker) LP positions.
+
+The LP Dollar team will use the price identifiers to enable fixed borrowing costs for Uniswap V2 liquidity providers.
```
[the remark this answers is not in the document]
dev-1-lp-dollar: It is similar to the yield dollar ETH & renBTC fixed borrowing that the UMA team has been supporting. I can elaborate, but it was more just mean to be a brief mention of what is possible with the price identifier.

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+
```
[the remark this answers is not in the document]
dev-1-lp-dollar: Okay I can elaborate on what we'll be using it for

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: We haven't released a reference implementation of the script but surely can. 

Note, we wouldn't want to conflict with anything the UMA team is putting out, however, and in the discord there was mention of UMA building a price feed for this UMIP. Can someone on the UMA team comment here?

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+     ```
+
+3) Third, the ETHUSD price must be queried and used to calculate the USD value of the WETH reserves. 
+
+    The methodology to query the price of ETHUSD is similar to [UMIP-6](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-6.md), with a different exchange list and decimal precision.  
+
+    Base Currency: ETH  
+    Quote Currency: USD  
+
+    Exchanges: Coinbase Pro (ETH:USD), Kraken (ETH:USD), Bitfinex (ETH:USD), Bitstamp (ETH:USD)
```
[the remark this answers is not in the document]
dev-1-lp-dollar: We can recommend endpoints for sure. I'm not sure what you mean about an api key though.

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+
+    Base Currency: ETH  
+    Quote Currency: USD  
+
+    Exchanges: Coinbase Pro (ETH:USD), Kraken (ETH:USD), Bitfinex (ETH:USD), Bitstamp (ETH:USD)  
+    Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.  
+    Price Steps: 0.01 (2 decimals in more general trading format)  
+    Rounding: Closest, 0.5 up  
+    Pricing Interval: 60 seconds  
+    Dispute timestamp rounding: down
```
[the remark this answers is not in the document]
dev-1-lp-dollar: good question - i think relative to the requested price feed timestamp makes the most sense to me. will add the clarification if that sounds good

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+    
+    = 97499.896966146357068372 * 1716.12 
+    
+    = 167321523.18154308
+    ```
+    
+4) Fourth, the BTC:USD price must be queried and used to calculate the USD value of the WBTC reserves. 
+
+    The methodology to query the price of BTC:USD is similar to [UMIP-7](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-7.md), with a different exchange list and decimal precision.  
+
```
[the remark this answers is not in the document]
dev-1-lp-dollar: For BTC:USD, we think two decimals is appropriate given that most endpoints we found only returned that many anyway. For the BTC:USD price, this will not effect the calculation much.

Decimals are important in the steps that follow getting the BTC:USD price, so we can be more specific there about the decimals there.

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
dev-1-lp-dollar: I think it is fair to treat USDC as equal to 1 USD. I suspect users would expect it to be treated as such, and given free exchange of USDC for USD on Coinbase Pro I think it is reasonable given market structure. 

If it trades off peg anywhere, I would suspect that to be due to trading fees on certain exchanges. But would be interested to see anywhere it is trading off peg.

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: Sure thing!

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: okay

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: My understanding is that using web3 or subgraph calls return the value at the end of the specific block. I.e. if i call `web3.queryContractState(199029)` it would return the state after block 199029 has been processed.

Because we are reading the reserves off chain at the end of the block (also applies if it queries at the beginning), then there is no potential for flashloan manipulation. FYI, I addressed a similar comment in [discourse here](https://discourse.umaproject.org/t/add-usd-uni-v2-wbtc-eth-as-a-price-identifier/148/9?u=dev-1-lpdollar).

There is the potential that a whale could suck it up and mess with the reserves by sending a huge buy or sell, but they would immediately be arbed by bots (costing them large amounts of money) and could not use a flash loan.

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
akhanubis: I'm in favor of assuming 1 USDC = 1 USD.
What would be the price sources for USDC/USD?

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
akhanubis: UMA team has released a template for price id UMIPs a few days ago: https://github.com/UMAprotocol/UMIPs/blob/master/price-identifier-template.md
One of the questions that the UMIP should provide an answer for is if you need an API key to query data from the diff price sources.

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
smb2796: I agree. We've traditionally treated USDC as equivalent to USD, and it seems reasonable to do the same here.

UMIPs/umip-usd-uni-v2-lp.md
Tommy1231232: @smb2796 can we use the same price feed that was done for Badger?

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
Reinis-FRP: Yes, it would be helpful to include brief explanation from the perspective of users (e.g. mint synth and sell for borrowers)

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
Reinis-FRP: relative to the requested price feed timestamp seems reasonable, so better to add it

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+The DVM currently does not support Uniswap V2 LP tokens.
+
+By enabling LP token price identifiers, UMA will open the door for the creation of synths with CFMM (constant function market maker) LP positions.
+
+The LP Dollar team will use the price identifiers to enable fixed borrowing costs for Uniswap V2 liquidity providers.
+
+## Data Sources & Price Feed Implementation
+
+| Uniswap V2 WBTC-ETH   |                                                                                                 |
```
smb2796, line 24: The names of all of these price ids need to be changed to omit spaces and also include the currency that these are being paired with. Suggest: `USD/Uniswap_V2_WBTC_ETH`

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
+## Headers
+| UMIP-usd-uni-v2-lp-DRAFT    |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add USD UNI-V2 LP token price identifiers              |
```
smb2796, line 4: Can you list all of the price identifier names here?

UMIPs/umip-usd-uni-v2-lp.md
smb2796: We should be able to yes. See this [pr](https://github.com/UMAprotocol/protocol/pull/2576/files) for an example of how to price LP tokens.

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+     ```
+     
+     web3.js Response
+     ```
+     167105037364529719
+     ```
+
+3) Third, the ETHUSD price must be queried and used to calculate the USD value of the WETH reserves. 
+
+    The methodology to query the price of ETHUSD is similar to [UMIP-6](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-6.md), with a different exchange list and decimal precision.
```
smb2796, line 129: Why is this change necessary?

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+
+3) Third, the ETHUSD price must be queried and used to calculate the USD value of the WETH reserves. 
+
+    The methodology to query the price of ETHUSD is similar to [UMIP-6](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-6.md), with a different exchange list and decimal precision.  
+
+    Base Currency: ETH  
+    Quote Currency: USD  
+
+    Exchanges: Coinbase Pro (ETH:USD), Kraken (ETH:USD), Bitfinex (ETH:USD), Bitstamp (ETH:USD)  
+    Input Processing: None. Human intervention in extreme circumstances where the result differs from broad market consensus.
```
smb2796, line 135: Please indent all of these

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+    USD value of WETH reserves = WETH in reserves * WETH:USD = 
+    
+    = 97499.896966146357068372 * 1716.12 
+    
+    = 167321523.18154308
+    ```
+    
+4) Fourth, the BTC:USD price must be queried and used to calculate the USD value of the WBTC reserves. 
+
+    The methodology to query the price of BTC:USD is similar to [UMIP-7](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-7.md), with a different exchange list and decimal precision.
```
smb2796, line 155: Why is this necessary? Also it would seem to make more sense to use the WBTCUSD price id, rather than BTCUSD

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+    
+    = 3667.03647028 * 45938.30 = 168457421.48266372
+    ```
+
+5) Fifth, use the UNI-V2-WBTC-ETH total supply of LP tokens and USD value of reserves to calculate the USD value of each LP token (UNI-V2-WBTC-ETH:USD).
+
+    Example calculation of UNI-V2-WBTC-ETH:USD:
+  
+    ```
+    UNI-V2-WBTC-ETH:USD =
```
smb2796, line 182: Great example

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+    ```
+    USD:UNI-V2-WBTC-ETH = 
+  
+    (1 / UNI-V2-WBTC-ETH:USD) * 1e18
+  
+    = (1 / 2009388525.6835613) * 1e18 
+  
+    = 497663835.15
+  
+    = 497663835
```
smb2796, line 203: This is confusing. The implementation should walk through how to calculate this price id before scaling. The voter dapp should abstract any necessary scaling away from a voter, and right now it seems like you are calculating the value of 1e18 USD in LP tokens.

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+     ```
+
+     web3.js Response
+     ```
+     2572499047307646516
+     ```
+
+3) Third, the ETHUSD price must be queried and used to calculate the USD value of the WETH reserves. 
+
+    The methodology to query the price of ETHUSD is similar to [UMIP-6](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-6.md), with a different exchange list and decimal precision.
```
smb2796, line 311: Since this is used in multiple places, I suggest defining a separate section for the alternative ETHUSD calculation to avoid redundancy.

UMIPs/umip-usd-uni-v2-lp.md
smb2796: I now see you added all of these at the end of the section, but please also add to `Summary`, `Headers` and the tope of each of these sections

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,766 @@
...
+    
+    = 131583140.29680333
+    ```
+
+4) Fourth, the UNI:USD price must be queried and used to calculate the USD value of the UNI reserves.  
+
+    Base Currency: UNI  
+    Quote Currency: USD  
+
+    Exchanges: Coinbase Pro (UNI:USD), Kraken (UNI:USD), Bitfinex (UNI:USD)
```
smb2796, line 498: Interesting that we have two UMIPs in the same week proposing different implementations for UNI/USD. Why did you choose these sources? The other price identifier uses Binance/OKEx/CB Pro, which seems more reasonable to me. Bitfinex and Kraken look to be lower volume than Binance and OKEx

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
dev-1-lp-dollar: Added 👍

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
dev-1-lp-dollar: Added 👍

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
dev-1-lp-dollar: Added 👍

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: Rationale was to use USD markets when possible rather than USDT markets

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: Initially this was written because it seemed the USD:BTC price should carry more decimals than specified in UMIP-7. However, as the example calculation price shows, we just need BTC:USD, so it should be fine to reuse UMIP-7.

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: i think swapping out Kraken for Binance is probably warranted given it only introduces one USDT market in the median calculation. And since Kraken has the lowest volume of the three.

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: Changed to use UMIP-6

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: Changed to use UMIP-7

UMIPs/umip-usd-uni-v2-lp.md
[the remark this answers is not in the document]
dev-1-lp-dollar: Added language to be more specific about decimals in calculation steps, and per other comments, have switched to using UMIP-7 for BTC:USD

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: Edited the example to take this into account

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: Regarding WBTCUSD price id, this UMIP seems to be missing clarification on the ETH:USD conversation (unless I missed it), and seems to add a decent amount of complexity. Unless there are objections, we believe BTC:USD will serve users (and liquidatiors and disputers) of the price id well.

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: A lot of the text here was taken out with the reference to UMIP-6. I think it is fine to leave this as the redundancy helps the examples flow better (i.e. rather than jumping from place to place in the doc).

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: Added Binance, took out Kraken

UMIPs/umip-usd-uni-v2-lp.md
smb2796: @dev-1-lp-dollar thoughts on this name change?

UMIPs/umip-usd-uni-v2-lp.md
smb2796: To @Reinis-FRP's point. Due to the amount of price ids in this umip, I think it is fine to combine `Markets & Data Sources` with the voting `Implementation`. I do think a separate `Price Feed Implementation` would work well however. In the `Price Feed Implementation` section, you can just explain that it would be the same model for all of these price identifiers. These should use a combination of the [CryptowatchPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js), [LPPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/LPPriceFeed.js) and [ExpressionPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js).

As a meta point, I think it is becoming apparent that we should make it a requirement that new price identifiers need a [default price feed config](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js) even when using existing price feeds. @dev-1-lp-dollar if you are able to submit a pr for this and link it in `Price Feed Implementation` that would be great. If not, I think that is fine given that this hasn't historically been a requirement.

UMIPs/umip-usd-uni-v2-lp.md
smb2796: Thanks, looks good to me.

UMIPs/umip-usd-uni-v2-lp.md
smb2796: That's fine

UMIPs/umip-usd-uni-v2-lp.md
smb2796: That seems reasonable to me

UMIPs/umip-usd-uni-v2-lp.md
smb2796: It has [ETHUSD](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-51.md#implementation) conversion. I'd argue that given the custodial nature of WBTC and given its tendency to trade off peg to BTC, it is necessary to calculate with WBTCUSD rather than BTCUSD. You could always define your own WBTCUSD implementation here if umip-51 does not seem suitable. What do you think?

smb2796:
@dev-1-lp-dollar @dev-2-lp-dollar curious to hear your thoughts on the comment I left around WBTC. Other than that, I think the main outstanding issue is the price identifier [names](https://github.com/UMAprotocol/UMIPs/pull/170#discussion_r580274794).

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,700 @@
+## Headers
+| UMIP-usd-uni-v2-lp-DRAFT    |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add USD UNI-V2 WBTC-ETH, USDC-ETH, UNI-ETH, UMA-ETH LP token price identifiers              |
```
smb2796, line 4: Can you please update these names to match the ones in `Technical Specifications`

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,700 @@
...
+    = (1 / 520.3440264478902)
+  
+    0.001921805477092653  
+    ```
+
+
+## Technical Specifications
+
+Price Identifier Name: USD/UNI-V2-WBTC-ETH  
+Base Currency: USD
```
smb2796, line 662: Please bullet all of these

UMIPs/umip-usd-uni-v2-lp.md
```diff
@@ -0,0 +1,700 @@
...
+    USD value of WETH reserves = WETH in reserves * WETH:USD = 
+    
+    = 1350.358508316793260065 * 1716.12 
+    
+    = 2317377.24329262
+    ```
+
+4) Fourth, the UMA:USD price must be queried and used to calculate the USD value of the UMA reserves.  
+
+    Base Currency: UMA
```
smb2796, line 609: Please bullet these and all similar sections

UMIPs/umip-usd-uni-v2-lp.md
dev-1-lp-dollar: I haven't seen data to suggest WBTC trades off peg by more than minting or redemption fees, but would be interested to see it. Can you share this? 

I have two thoughts / concerns that makes me hesitant: 

1) As a minter, I would anticipate 1 WBTC to be equal to 1 BTC.
2) UMIP-51 looks robust enough for a price oracle but seems to add non-trivial overhead to calculating the liquidation price.

So, I suppose the amount by which it trades off peg would determine the robustness here. I wouldn't want users to get liquidated because it traded off peg and BitGo wasn't hacked and WBTC was redeemable for 1 BTC - redemption cost. But I also see the value in handling the custodial nature here.

smb2796: Fair points. I generally meant the difference caused by minting + redemption fees, but you're right this difference is somewhat negligible.

I'd generally argue that the actual act of calculating the WBTCUSD price is no more difficult than for BTCUSD, but I recognize that this is subjective.

If some WBTC black swan event did happen, WBTC started trading way off peg from BTC and this price id was not representative of the actual LP token values, voters could always come to a consensus around using the WBTC rates instead. This level of flexibility is one of the coolest things about the DVM imo, and gives us an option if this ever does become a bigger concern.

Overall you've convinced me and I feel fine with leaving it as is!

smb2796 (approved):
Looks like all comments have been addressed except for a few small formatting issues. Will fix those on my end. Going to merge in and move to `Last Call`.

PR #171, opened 2021-02-11
Add UNI-V2 LP tokens as whitelisted collateral currencies

dev-1-lp-dollar:
This is a proposal to add the following UNI-V2 LP tokens as approved collateral currencies: WBTC-ETH, USDC-ETH, UNI-ETH, and UMA-ETH. 

Supporting the collateral will allow LP Dollar to offer fixed borrowing costs to UNI-V2 liquidity providers, and serve as a tool for the UMA ecosystem to create synthetics with UNI-V2 LP positions as collateral.


UMIPs/umip-uni-v2-lp-collateral.md
```diff
@@ -0,0 +1,65 @@
...
+## Summary
+This UMIP will add the following UNI-V2 LP tokens as approved collateral currencies: WBTC-ETH, USDC-ETH, UNI-ETH, and UMA-ETH. 
+
+This will involve adding the currencies to the whitelist and adding a flat final fee to charge per-request. The proposed final fees are listed in the table below. 
+
+| Currency (LP token) | Final Fee in LP token | Final Fee in Dollars |
+|---------------------|-----------------------| -------------------- | 
+| WBTC-ETH | 0.00000025 | $500 USD |
+| USDC-ETH | 0.0000000465 | $400 USD |
+| UNI-ETH | 0.000000000000297 | $400 USD |
```
Tommy1231232, line 19: Can you explain how you got these decimal numbers? They seem a bit low

Tommy1231232: UNI-ETH has 300 mill of liquidity and 360,000 outstanding LP tokens. This means that every LP token is worth around $800, so really it should be around .5-.6

dev-1-lp-dollar: Thanks for pointing this out. I just updated the calculations and detailed the numbers used.

UMIPs/umip-uni-v2-lp-collateral.md
```diff
@@ -0,0 +1,76 @@
...
+
+
+
+## Motivation
+UNI-V2 LP tokens represent ownership in CFMM (constant function market maker) positions on Uniswap. 
+
+The underlying assets include ETH, WBTC - a currency pegged to Bitcoin, USDC - a USD stablecoin, UNI - Uniswap's governance, and UMA - this projects own governance token. All tokens are widely used across the cryptocurrency space, and account for a large amount of activity in the decentralized finance space.
+ 
+UNI-V2 LP tokens as a collateral type are expected to have a variety of deployments. They will open the door for the creation of synthetic assets with CFMM positions.
+
```
mhairimca, line 39: Could you elaborate on why a user may wish to use LP tokens as collateral.

chicfilabae: Maybe also a comment or two regarding how the impermanent loss that a user faces by holding LP tokens might affect their collateral value, and any implications for getting liquidated?

UMIPs/umip-uni-v2-lp-collateral.md
```diff
@@ -0,0 +1,76 @@
...
+## Rationale
+The rationale behind this change is giving deployers more useful collateral currency options. This is an advancement into a better type of collateral.
+
+$400 USD equivalent was chosen as the final fee because it is equal to or above the mimimum of already approved coins.
+
+## Implementation
+
+This change has no implementation other than proposing the two aforementioned governance transactions that will be proposed.
+
+## Security Considerations
```
akhanubis, line 69: Similarly to https://github.com/UMAprotocol/UMIPs/pull/189, there are some implications related to having UMA as collateral (in this case as UNI UMA-ETH) that should be clearly laid out in this section.

pemulis: See my reply here: https://github.com/UMAprotocol/UMIPs/pull/189#discussion_r579422123

I think that locking up UMA-ETH LP tokens enhances the UMA security mechanism by removing UMA from the market, at least temporarily. The LP would need to pay down their debt to unlock their LP collateral and withdraw it from the pool, which is a necessary step to using the UMA tokens to mount an attack on the DVM or dump the tokens on the market where an attacker could accumulate them.

UMIPs/umip-uni-v2-lp-collateral.md
pemulis: I don't think impermanent loss harms the value of the LP tokens, since it's a relative measure of your profit from being a liquidity provider vs. your profit from holding the two tokens independently. The LP token holds its value, but the provider loses money (at least temporarily) compared to what they would have made by following a different investment strategy.

The value of an LP token itself should appreciate over time due to the collection of fees, even in cases where the liquidity provider is suffering brutal impermanent loss. The way a liquidation may occur is if the two tokens composing the liquidity pool are losing value compared to a synthetic minted with the LP token as collateral. For instance, if you had a WBTC/ETH LP token as collateral to mint synthetic dollars, and then the entire crypto market crashes compared to the dollar. In that case, someone using the LP token as collateral could get liquidated, but that's not really an impermanent loss issue.

UMIPs/umip-uni-v2-lp-collateral.md
```diff
@@ -0,0 +1,76 @@
...
+UNI = $21.05  
+UMA = $28.98
+```
+ 
+| Currency (LP token) |  Total Reserves | LP Token Supply | LP Token in Dollars | Final Fee in LP token |
+| ------------------- | --------------- | --------------- | ------------------- | -------------------- |
+| WBTC-ETH | WBTC: 4,004.25660 <br /> ETH: 108,354.69868 | 0.183766117503018792 | ```((4,004.25660 * 48,659.32) + (108,354.69868 * 1,803.90)) / 0.183766117503018792``` = $2,123,924,962.41 | `400 / 2,123,924,962.41` = 0.00000018833057056127
+| USDC-ETH | USDC: 176,760,773.71622 <br /> ETH: 97,320.84342 | 2.961635437353718411 | ```((176,760,773.71622 * 1) + (97,320.84342 * 1,803.90)) / 2.961635437353718411``` = $118,960,571.14 | `400 / 118,960,571.14` = 0.00000336245863790664
+| UNI-ETH  | UNI: 6,619,997.29037 <br /> ETH: 77,263.63438 | 362,909.728355933628948352 | ```((6,619,997.29037 * 21.05) + (77,263.63438 * 1,803.90)) / 362,909.728355933628948352``` = $768.03 | `400 / 768.03` = 0.52081298907594755413
+| UMA-ETH  | UMA: 81,710.86344 <br /> ETH: 1,310.95349 | 8,717.786395937747945885 | ```((81,710.86344 * 28.98) + (1,310.95349 * 1,803.90)) / 8,717.786395937747945885``` = $542.89 | `400 / 542.89` = 0.73679750962441746947
```
pemulis, line 29: I double-checked all of the math here, and compared the numbers to uniswap.info and Etherscan. Obviously, the specific numbers of LP token supply, token balance, and token price change from day to day, but this document is within a reasonable range of the current situation in these pools, and the final fee calculated in LP tokens looks good!

UMIPs/umip-uni-v2-lp-collateral.md
```diff
@@ -0,0 +1,76 @@
...
+
+As mentioned above, the asset is packaged from Uniswap, a decentralized protocol on the Ethereum blockchain. Uniswap is one of the most popular, proven, and secure smart contract protocols to exist in decentralized finance and beyond. They've succesfully locked up and handled billions of dollars over the course of multiple years, with multiple extensive auditing from top names in the industry. There is the theoretical risk that the pools on Uniswap could be exploited and lead to a rapid loss of value in the currency proposed here which would require fast response to ensure solvency for any financial products built using this as a collateral type. However, due to the nature of Uniswap's long-proven track record around quality and security this is extremely unlikely. 
+
+The new price identifiers that are intended to be used with these collateral currencies will need to be specified to 18 decimals of precision in order to comply with UNI-V2 LP token's precision.
```
pemulis, line 76: This sentence can be cut, assuming people are using the latest contracts. The new EMP and Perpetual contracts launched this week allow price identifiers and collateral currencies to have different decimals of precision. For example, if you wanted to issue a synthetic with six decimals of precision against an LP token collateral with 18 decimals of precision, that would be fine.

UMIPs/umip-uni-v2-lp-collateral.md
dev-1-lp-dollar: I would agree with @pemulis here. It seems to me like there is risk to having UMA in a market where it can be borrowed, but here the LP tokens would be used as collateral exclusively.

UMIPs/umip-uni-v2-lp-collateral.md
dev-1-lp-dollar: Thanks!

UMIPs/umip-uni-v2-lp-collateral.md
akhanubis: Excellent

UMIPs/umip-uni-v2-lp-collateral.md
```diff
@@ -0,0 +1,76 @@
+# Headers
+| UMIP-#     |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add UNI-V2 LP tokens as whitelisted collateral currencies              |
```
smb2796, line 4: Please add all of the collateral names to the title

UMIPs/umip-uni-v2-lp-collateral.md
smb2796: Also a great way to structure this. Thank you for making this consumable

UMIPs/umip-uni-v2-lp-collateral.md
dev-2-lp-dollar: Great, thanks for the clarification! Removing now.

UMIPs/umip-uni-v2-lp-collateral.md
smb2796: @dev-1-lp-dollar these final fees are actually unnecessarily precise in my opinion, especially since $400 is just an estimated threshold and the final fee will obviously fluctuate in dollar value. Also some of these are not actually possible, given that they have more than 18 decimals.

Here are my suggestions for :
-  WBTC-ETH: 0.0000002: ~$425 with the numbers you provided
-  USDC-ETH: 0.0000035: ~$415
-  UNI-ETH: 0.55: ~$425
-  UMA-ETH: 0.8: ~$430

What do you think? If you are fine with this, can you please update here and also in `Technical Specifications`?

dev-1-lp-dollar: Sounds good to me. Just updated.

smb2796 (approved):
Looks great to me! Going to merge in and move to last call

PR #174, opened 2021-02-11
Add UNI-V2-USDC-ETH as a supported collateral currency

dev-1-lp-dollar:


dev-1-lp-dollar:
Merged with https://github.com/UMAprotocol/UMIPs/pull/171

PR #190, opened 2021-02-17
Add proposal for various protocol altcoins as price identifier pairs against USD

jaxrtech:
This is the "sibling" PR to #189 to add the matching price identifier against USD for the same collateral currencies:

-  AAVE (Aave)
-  LINK (Chainlink)
-  SNX (Synthetix Network Token)
-  UMA (UMA Governance Token)
-  UNI (Uniswap Token)

+ @smb2796 

**Edit**: We're in the process of editing this to be against USDC instead of USDT (especially since Balancer does not have USDT pools)


UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+The DVM currently does not support the AAVE/USD or USD/AAVE index.
+
+Cost: Pricing for this index is easy to access through open centralized exchange APIs, and cross-exchange price discrepancies are typically negligible.
+
+## Markets & Data Sources
+Coinbase Pro, Binance, and OKEx should be used to construct the price. These 3 exchanges comprise a significant amount of AAVE trade volume and have available pricefeeds on Cryptowatch.
+
+ 1. Which specific pairs should be queried from each market?
+
+    - Coinbase Pro: AAVE/USD
```
mhairimca, line 25: Should this be USDC?

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
+ **Headers**
+| UMIP [#]   |      |
+|------------|------|
+| UMIP Title | Add price identifiers for {AAVE, LINK, SNX, UMA, UNI}/USD and inverse pairs |
+| Authors    | Josh Bowden (josh@ferrosync.io) |
+| Status     | Draft |
+| Created    | February 17th, 2021 |
+<br>
```
mhairimca, line 8: discourse link?

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+
+# AAVE (Aave Token)
+
+## Summary
+The DVM should support price requests for the AAVE/USD, USD/AAVE price index.
+
+## Motivation
```
Reinis-FRP, line 15: Can you please elaborate on the types of financial products that will be created and the mechanics of an example financial product using this price-id?
The same comment applies to all other token price-id `Motivation` sections below.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+     - Approximately $5
+
+### Price Feed Implementation
+
+Associated AAVE price feeds are available via Cryptowatch. No other further feeds required.
+
+### Technical Specifications
+
+#### AAVE/USD:
+  -  Price Identifier Name: AAVE/USD
```
Reinis-FRP, line 76: Since `AAVEUSD` has less than 8 characters, slash character can be omitted from the price-id.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+  -  Quote Currency: USD
+  -  Intended Collateral Currency: USDT
+  -  Collateral Decimals: 6
+  -  Rounding: Round to nearest 6 decimal places (seventh decimal place digit >= 5 rounds up and < 5 rounds down)
+  -  Does the value of this collateral currency match the standalone value of the listed quote currency?: YES
+  -  Is your collateral currency already approved to be used by AAVE financial contracts?: YES
+
+#### USD/AAVE:
+
+  -  Price Identifier Name: USD/AAVE
```
Reinis-FRP, line 87: Since `USDAAVE` has less than 8 characters, slash character can be omitted from the price-id.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+      - Approximately $5
+
+## Price Feed Implementation
+
+Associated LINK price feeds are available via Cryptowatch. No other further feeds required.
+
+### Technical Specifications
+
+#### LINK/USD:
+-  Price Identifier Name: LINK/USD
```
Reinis-FRP, line 183: Since `LINKUSD` has less than 8 characters, slash character can be omitted from the price-id.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+-  Quote Currency: USD
+-  Intended Collateral Currency: USDT
+-  Collateral Decimals: 6
+-  Rounding: Round to nearest 6 decimal places (seventh decimal place digit >= 5 rounds up and < 5 rounds down)
+-  Does the value of this collateral currency match the standalone value of the listed quote currency?: YES
+-  Is your collateral currency already approved to be used by LINK financial contracts?: YES
+
+#### USD/LINK
+
+-  Price Identifier Name: USD/LINK
```
Reinis-FRP, line 194: Since `USDLINK` has less than 8 characters, slash character can be omitted from the price-id.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+ 
+## Price Feed Implementation
+
+Associated SNX price feeds are available via Cryptowatch. No other further feeds required.
+
+### Technical Specifications
+
+#### SNX/USD
+
+-  Price Identifier Name: SNX/USD
```
Reinis-FRP, line 282: Since `SNXUSD` has less than 8 characters, slash character can be omitted from the price-id.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+-  Quote Currency: USD
+-  Intended Collateral Currency: USDT
+-  Collateral Decimals: 6
+-  Rounding: Round to nearest 6 decimal places (seventh decimal place digit >= 5 rounds up and < 5 rounds down)
+-  Does the value of this collateral currency match the standalone value of the listed quote currency?: YES
+-  Is your collateral currency already approved to be used by SNX financial contracts?: YES
+
+#### USD/SNX
+
+-  Price Identifier Name: USD/SNX
```
Reinis-FRP, line 293: Since `USDSNX` has less than 8 characters, slash character can be omitted from the price-id.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+
+    - Coinbase Pro LINK/USD: <https://cryptowat.ch/charts/COINBASE-PRO:LINK-USD>
+    - Binance LINK/USDT: <https://cryptowat.ch/charts/BINANCE:LINK-USDT>
+    - OKEx LINK/USDT: <https://cryptowat.ch/charts/OKEX:LINK-USDT>
+
+ 3. Provide recommended endpoints to query for real-time prices from each market listed.
+
+    - Coinbase Pro AAVE/USD: <https://api.cryptowat.ch/markets/coinbase-pro/linkusd/price>
+    - Binance AAVE/USDT: <https://api.cryptowat.ch/markets/binance/linkusdt/price>
+    - OKEx AAVE/USDT: <https://api.cryptowat.ch/markets/okex/linkusdt/price>
```
Reinis-FRP, line 140: Replace `AAVE` with `LINK`

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+
+# UMA (UMA Governance Token)
+
+## Summary
+
+The DVM should support price requests for the UMA/USD, USD/UMA price index.
+
+## Motivation
+
+The DVM currently does not support the LINK/USD or USD/LINK index.
```
Reinis-FRP, line 323: Replace `LINK` with `UMA`

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+
+    - Coinbase Pro: UMA/USD
+    - Binance: UMA/USDT
+    - OKEx: UMA/USDT
+
+ 2. Provide recommended endpoints to query for real-time prices from each market listed.
+
+    - Coinbase Pro AAVE/USD: <https://api.cryptowat.ch/markets/coinbase-pro/umausd/price>
+    - Binance AAVE/USDT: <https://api.cryptowat.ch/markets/binance/umausdt/price>
+    - OKEx AAVE/USDT: <https://api.cryptowat.ch/markets/okex/umausdt/price>
```
Reinis-FRP, line 342: Replace `AAVE` with `UMA`

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+  -  Collateral Decimals: 18
+  -  Rounding: Round to nearest 18 decimal places (nineteenth decimal place digit >= 5 rounds up and < 5 rounds down)
+  -  Does the value of this collateral currency match the standalone value of the listed quote currency?: YES
+  -  Is your collateral currency already approved to be used by AAVE financial contracts?: In progress
+
+### Implementation
+
+Voters should query for the price of AAVE/USD at the price request timestamp on Coinbase Pro, Binance & OKEx. Recommended endpoints are provided in the markets and data sources section.
+
+1.  When using the recommended endpoints, voters should use the open price of the OHLC period that the timestamp falls in.
```
Reinis-FRP, line 100: Which pricing interval for the OHLC period?
The same should be clarified for all other token price-ids below.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+2.  The median of these results should be taken
+3.  The median from step 2 should be rounded to six decimals to determine the AAVEUSD price.
+4.  The value of USDAAVE will follow the exact same process but undergo one additional step: it will be the result of dividing 1/AAVEUSD.
```
Reinis-FRP, line 103: rounding to 18 decimals

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+
+    - Coinbase Pro: LINK/USD
+    - Binance: LINK/USDT
+    - OKEx: LINK/USDT
+
+ 2. Provide recommended websites to manually inspect real-time and historical prices.
+
+    - Coinbase Pro LINK/USD: <https://cryptowat.ch/charts/COINBASE-PRO:LINK-USD>
+    - Binance LINK/USDT: <https://cryptowat.ch/charts/BINANCE:LINK-USDT>
+    - OKEx LINK/USDT: <https://cryptowat.ch/charts/OKEX:LINK-USDT>
```
Reinis-FRP, line 134: Better provide API endpoints, not websites (similarly as for other tokens)

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+-  Is your collateral currency already approved to be used by LINK financial contracts?: In progress 
+
+### Implementation
+
+Voters should query for the price of LINK/USD at the price request timestamp on Coinbase Pro, Binance & OKEx. Recommended endpoints are provided in the markets and data sources section.
+
+ 1.  When using the recommended endpoints, voters should use the open price of the OHLC period that the timestamp falls in.
+ 2.  The median of these results should be taken
+ 3.  The median from step 2 should be rounded to six decimals to determine the LINKUSD price.
+ 4.  The value of USDLINK will follow the exact same process but undergo one additional step: it will be the result of dividing 1/LINKUSD.
```
Reinis-FRP, line 210: rounding to 18 decimals

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+-  Is your collateral currency already approved to be used by SNX financial contracts?: In progress 
+
+### Implementation
+
+Voters should query for the price of SNX/USD at the price request timestamp on Coinbase Pro, Binance & OKEx. Recommended endpoints are provided in the markets and data sources section.
+
+1.  When using the recommended endpoints, voters should use the open price of the OHLC period that the timestamp falls in.
+2.  The median of these results should be taken
+3.  The median from step 2 should be rounded to six decimals to determine the SNXUSD price.
+4.  The value of USDSNX will follow the exact same process but undergo one additional step: it will be the result of dividing 1/SNXUSD.
```
Reinis-FRP, line 309: rounding to 18 decimals

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+-  Is your collateral currency already approved to be used by UMA financial contracts?: In progress 
+
+### Implementation
+
+Voters should query for the price of UMA/USD at the price request timestamp on Coinbase Pro, Binance & OKEx. Recommended endpoints are provided in the markets and data sources section.
+
+1.  When using the recommended endpoints, voters should use the open price of the OHLC period that the timestamp falls in.
+2.  The median of these results should be taken
+3.  The median from step 2 should be rounded to six decimals to determine the UMAUSD price.
+4.  The value of USDUMA will follow the exact same process but undergo one additional step: it will be the result of dividing 1/UMAUSD.
```
Reinis-FRP, line 415: rounding to 18 decimals

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+-  Is your collateral currency already approved to be used by UNI financial contracts?: In progress
+
+### Implementation
+
+Voters should query for the price of UNI/USD at the price request timestamp on Coinbase Pro, Binance & OKEx. Recommended endpoints are provided in the markets and data sources section.
+
+1.  When using the recommended endpoints, voters should use the open price of the OHLC period that the timestamp falls in.
+2.  The median of these results should be taken
+3.  The median from step 2 should be rounded to six decimals to determine the UNIUSD price.
+4.  The value of USDUNI will follow the exact same process but undergo one additional step: it will be the result of dividing 1/UNIUSD.
```
Reinis-FRP, line 515: rounding to 18 decimals

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+
+  -  Price Identifier Name: USD/AAVE
+  -  Base Currency: USD
+  -  Quote Currency: AAVE
+  -  Intended Collateral Currency: AAVE
+  -  Collateral Decimals: 18
+  -  Rounding: Round to nearest 18 decimal places (nineteenth decimal place digit >= 5 rounds up and < 5 rounds down)
+  -  Does the value of this collateral currency match the standalone value of the listed quote currency?: YES
+  -  Is your collateral currency already approved to be used by AAVE financial contracts?: In progress
+
```
Reinis-FRP, line 95: `RATIONALE` section is missing for all price-id tokens (The rationale should flesh out the specification by describing what motivated the design and why particular design decisions were made, as well as any alternative designs that were considered)

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+Associated AAVE price feeds are available via Cryptowatch. No other further feeds required.
+
+### Technical Specifications
+
+#### AAVE/USD:
+  -  Price Identifier Name: AAVE/USD
+  -  Base Currency: AAVE
+  -  Quote Currency: USD
+  -  Intended Collateral Currency: USDT
```
akhanubis, line 79: Most platforms don't support USDT as collateral because of certain lack of trust related to it's peg. Is there a particular reason for you to choose USDT instead of USDC?
If you decide to go with USDT, then Collateral Decimals should be 18.

Also applies to other identifiers quoted in USD.

BreezeRo: I also have a preference for USDC over tether, however I believe tether decimals are indeed 6 not 18, per contract.methods.decimals() @ 0xdac17f958d2ee523a2206206994597c13d831ec7

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+#### USD/SNX
+
+-  Price Identifier Name: USD/SNX
+-  Base Currency: USD
+-  Quote Currency: SNX
+-  Intended Collateral Currency: SNX
+-  Collateral Decimals: 18
+-  Rounding: Round to nearest 18 decimal places (nineteenth decimal place digit >= 5 rounds up and < 5 rounds down)
+-  Does the value of this collateral currency match the standalone value of the listed quote currency?: YES
+-  Is your collateral currency already approved to be used by SNX financial contracts?: In progress
```
mhairimca, line 300: Should this read UMA?

UMIPs/umip-draft_alt_price_identifiers.md
mhairimca: Also a typical user who would want to use a contract built on this price feed.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+2.  The median of these results should be taken
+3.  The median from step 2 should be rounded to six decimals to determine the UNIUSD price.
+4.  The value of USDUNI will follow the exact same process but undergo one additional step: it will be the result of dividing 1/UNIUSD.
+ For both implementations, voters should determine whether the returned price differs from broad market consensus. This is meant to provide flexibility in any unforeseen circumstances as voters are responsible for defining broad market consensus.
+
+<br>
+
+# Security Considerations
+These are the first identifiers proposed for registration with the DVM which are not based on existing trading pairs. Adding these new identifiers by itself poses little security risk to the DVM or priceless financial contract users. However, anyone deploying a new priceless token contract referencing these identifiers should take care to parameterize the contract appropriately to avoid the loss of funds for synthetic token holders. Additionally, the contract deployer should ensure that there is a network of liquidators and disputers ready to perform the services necessary to keep the contract solvent.
```
pemulis, line 521: `These are the first identifiers proposed for registration with the DVM which are not based on existing trading pairs.`

What do you mean?

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: @mhairimca All of Coinbase Pro's tickers are named `USD` since they are against fiat USD. Although somewhat confusing, this is intentional.

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: I've change all of the collateral currencies to USDC. We originally wanted to so for putting collateral in from centralized exchanges without the additionally swap since most are pegged to USDT. Given that Balancer doesn't have any USDT pool (among other reasons as you point out), it's indicative that we should clearly be using USDC.

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: @Reinis-FRP fixed

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: Fixed

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+
+#### AAVE/USD:
+  -  Price Identifier Name: AAVE/USD
+  -  Base Currency: AAVE
+  -  Quote Currency: USD
+  -  Intended Collateral Currency: USDT
+  -  Collateral Decimals: 6
+  -  Rounding: Round to nearest 6 decimal places (seventh decimal place digit >= 5 rounds up and < 5 rounds down)
+  -  Does the value of this collateral currency match the standalone value of the listed quote currency?: YES
+  -  Is your collateral currency already approved to be used by AAVE financial contracts?: YES
```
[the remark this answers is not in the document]
jaxrtech: Fixed those typos

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: Fixed typo 👍🏾

UMIPs/umip-draft_alt_price_identifiers.md
akhanubis: > I also have a preference for USDC over tether, however I believe tether decimals are indeed 6 not 18, per contract.methods.decimals() @ 0xdac17f958d2ee523a2206206994597c13d831ec7

You are right! I was thinking about dai decimals prob.

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: Created post and added link 👍🏾

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,523 @@
...
+
+    - No
+
+ 8. Is there a cost associated with usage?
+
+    - Yes
+
+ 9. If there is a free tier available, how many queries does it allow for?
+
+    - The lower bound on the number of queries allowed per hour is >> 1000.
```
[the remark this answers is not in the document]
jaxrtech: You're correct. I've updated the estimated number of queries for the free tier for all the coins.

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: @Reinis-FRP, @mhairimca. Thanks for your feedback! I've updated the draft with a brief overall motivation. I've copied it below for convenience:
> #### Motivation
>
>The purpose of adding these price identifiers is to be able to mint synthetic tokens priced at the defi protocol's token valued against USD. By doing so, users who mint synthetic tokens and sell them to the market are able to short the protocol. Similarly, users who buy the synthetic token are able to long the protocol.
>
>If the inverse price identifier is used (e.g. USD/UMA) and the same altcoin token is employed as collateral, then the value of the synthetic token can be traded in place of the underlying collateral without having to sell the underlying asset.

Let me know if that helps clarify the purpose of the UMIP

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: I've went on and removed this as it -- well -- isn't the case and is superfluous. Thanks

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: It's 1 minute (going off of previous CryptoWatch UMIPs). I've updated the wording. 👍🏾

UMIPs/umip-draft_alt_price_identifiers.md
smb2796: This looks good to me!

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,541 @@
+ **Headers**
+| UMIP [#]   |      |
+|------------|------|
+| UMIP Title | Add price identifiers for {AAVE, LINK, SNX, UMA, UNI}/USD and inverse pairs |
```
smb2796, line 4: I know there are quite a few, but can you list all of the price identifier names here in full? Also like @Reinis-FRP brought up, I think all of the price ids should have the slash omitted. `UMAUSD` instead of `UMA/USD`

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,541 @@
...
+  -  Is your collateral currency already approved to be used by UMA financial contracts?: YES
+
+#### USD/AAVE:
+
+  -  Price Identifier Name: `USDAAVE`
+  -  Base Currency: USD
+  -  Quote Currency: AAVE
+  -  Intended Collateral Currency: AAVE
+  -  Collateral Decimals: 18
+  -  Rounding: Round to nearest 18 decimal places (nineteenth decimal place digit >= 5 rounds up and < 5 rounds down)
```
smb2796, line 110: I'd be in favor of not rounding all of the USD/XYZ price ids to 18 decimals. If anything this just creates a possibility of voting confusion and seems like an unnecessary level of precision given the prices of these tokens. Something like 6-8 rounding decimals would seem more suitable to me.

mhairimca: I thought everything now had to be 18 decimals?

jaxrtech: It's also possible that we could just have it *stored* at 18 decimals internally on-chain (since most things expect 18 decimals), but just *round* to 6 decimal places. This is mainly because centralized exchanges are not going to quote against USD more than 2-6 decimals so rounding more than that won't do much.

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: Fixed 👍🏾

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: ...never mind, @smb2796 @mhairimca, I forgot that the USDC ERC20 is in fact 6 decimals on-chain. Not sure what the best way to handle this is or just keep it as is? Let me know what you think is best.

smb2796: Yep USDC is 6 decimals on-chain. Previously we needed to be specific about the "Collateral Decimals" that a price id would use, so that we could scale the price identifier accordingly when submitting votes. This is no longer needed in the new contract types, and everything should have "Collateral Decimals" of 18.

"Collateral decimals" is separate from rounding decimals. You can round to fewer decimals than the scaling factor and that is what I was suggesting. What do you think about:
- Collateral decimals: 18
- Rounding: 6 decimals

Rounding with 6 decimals and Aave at $400 would imply a loss of precision of no more than $.0002 which seems acceptable to me.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,541 @@
...
+Associated AAVE price feeds are available via Cryptowatch. No other further feeds required.
+
+### Technical Specifications
+
+#### AAVE/USD:
+  -  Price Identifier Name: `AAVEUSD`
+  -  Base Currency: AAVE
+  -  Quote Currency: USD
+  -  Intended Collateral Currency: USDC
+  -  Collateral Decimals: 6
```
smb2796, line 98: > Yep USDC is 6 decimals on-chain. Previously we needed to be specific about the "Collateral Decimals" that a price id would use, so that we could scale the price identifier accordingly when submitting votes. This is no longer needed in the new contract types, and everything should have "Collateral Decimals" of 18.

Related to this comment, everything should now have `Collateral Decimals` of 18. Sorry about the confusion here. Getting the timing right between price identifiers being usable and the contracts being ready has been difficult. Can you make this correction for all of the USD denominated price ids?

`Collateral Decimals` also now seems like a misleading name. It is really just a scaling factor and doesn't accurately represent USDC's decimals. I'm thinking we should rename this field to `Scaling Decimals`. What do you think?

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,541 @@
...
+ * SNXUSD, USDSNX
+ * UMAUSD, USDUMA
+ * UNIUSD, USDUNI
+
+# Motivation
+
+The purpose of adding these price identifiers is to be able to mint synthetic tokens priced at the defi protocol's token valued against USD. By doing so, users who mint synthetic tokens and sell them to the market are able to short the protocol. Similarly, users who buy the synthetic token are able to long the protocol.
+
+If the inverse price identifier is used (e.g. USD/UMA) and the same altcoin token is employed as collateral, then the value of the synthetic token can be traded in place of the underlying collateral without having to sell the underlying asset.
+
```
smb2796, line 25: @Reinis-FRP had a good point. The rationale section is missing. It may seem obvious to us why the calculation methodology was chosen etc, but may not be for new UMIP readers. Since the rationale should be the same for most of the price ids, I think we can get away with 1 rationale section below `Motivation`.

Could you just explain why the specific markets were chosen in a `Rationale` section? Unless you all had some additional rationale, it should just be something along the lines of "highest volume" and "easily calculated because available on cryptowatch"

smb2796:
Left a few more comments, but looks super close to being ready to merge.

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: Thanks for the feedback, I've added a rational section for each price identifier. Let me know if that covers what you were looking for. 👍🏾

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: lol @smb2796 I must have pushed my next commit around when you were typing this. I've added it per price identifer anyhow.

smb2796: Works for me!

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: @smb2796  It's understandable, I remember there being some discussion on trying to overhaul how decimals are handled. I think "Scaling Decimals" probably makes more sense. Nonetheless, I don't know if there is any relevancy for knowing the number of decimals the collateral's ERC20 has -- or if that is just queried or cached form ERC20 contract? If we need to specify both, I'd maybe call it "Price Feed Scaling Decimals" and "ERC20 Collateral Decimals".

UMIPs/umip-draft_alt_price_identifiers.md
jaxrtech: @smb2796 This seems acceptable, I've renamed it to "Scaling Decimals" and just have them set them both to round to 6 decimals. Additionally updated the "Rationale" accordingly and move it to the top so it's not duplicated each time.

UMIPs/umip-draft_alt_price_identifiers.md
```diff
@@ -0,0 +1,582 @@
...
+If the inverse price identifier is used (e.g. USD/UMA) and the same altcoin token is employed as collateral, then the value of the synthetic token can be traded in place of the underlying collateral without having to sell the underlying asset.  
+
+# Rationale
+For each altcoin token, we choose to have both `XYZUSD` and `USDXYZ` price identifiers (where the symbol `XYZ` is either `AAVE`, `LINK`, `SNX`, `UMA`, `UNI`) to allow for financial products that use either the altcoin token or stable USDC as collateral.
+ * Using USDC as collateral allows for speculating on the price of `XYZUSD` -- risking liquidation if the price of the altcoin increases relative to USD.
+ * Using the altcoin as collateral allows for borrowing synthetic USDC equivalents tracking `USDXYZ` -- risking liquidation if altcoin *decreases* in value relative to USD.
+
+We picked the top 3 exchanges by trading volume to avoid price movements solely to do poor liquidity.
```
smb2796, line 31:
```suggestion
We picked the top 3 exchanges by trading volume to avoid price movements due solely to poor liquidity.
```

smb2796 (approved):
Looks good. Going to merge in and move to Last Call

UMIPs/umip-draft_alt_price_identifiers.md
yokymia: Hey everyone

PR #194, opened 2021-02-19
Update price identifier decimals and deprecate legacy price IDs

smb2796:
This PR does two things.

1. Adds a deprecation notice for any price identifier that is currently being used in a mainnet contract **AND** has a non-18 decimal scaling specification.
2. Edits all identifiers, that do not follow into the first category, to specify 18 decimals of scaling.

This is necessary because the new [EMP](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-54.md) and [Perp](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-53.md) require that all price identifiers be scaled to 18 decimals. Price identifiers that fall into the first category will resolve incorrectly if used with the new factories, hence the deprecation notices.

### DIFF
--- a/UMIPs/umip-13.md
+++ b/UMIPs/umip-13.md
@@ -28,6 +28,7 @@ The definition of this identifier should be:
 - Rounding: Closest, 0.5 up
 - Pricing Interval: 60 seconds
 - Dispute timestamp rounding: down
+- Scaling Decimals: 18
 
 The definition of this identifier should be:
 - Identifier name: USDPERL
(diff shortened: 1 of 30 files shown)


smb2796:
Closing until the new factories are ready to be used

pemulis (approved):
LGTM! Good to get this in with the new EMP and perpetual contracts going into use.

nicholaspai (approved):

PR #198, opened 2021-02-24
Add ETHBTC_FR as a supported price id

smb2796:


UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+
+## IMPLEMENTATION
+Voters should determine which pricing implementation to use depending on when the price request was submitted.
+
+1. Following the specifications in [UMIP-2](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-2.md), query for the ETHBTC price at the disputed funding rate proposal timestamp.
+2. Query for the cumulative funding rate multiplier at the same timestamp.
+3. Multiply the ETHBTC price from step 1 by the CFRM from step 2 to get the fair value of the ETHBTC synthetic (ETHBTC-FV). [ETHBTC * CFRM].
+4. Query for the ETHBTC-PERP 5 minute TWAP from the listed AMM pool. This will return the ETHBTC-PERP's TWAP denominated in USDC.
+5. Subtract the result of step 3 from the result of step 4. [ETHBTC-PERP - ETHBTC-FV].
+6. Divide the result of step 5 by the ETHBTC rate. [ETHBTC-PERP - ETHBTC-FV]/ETHBTC-FV.
```
akhanubis, line 136: If I understand correctly from step 3, ETHBTC * CFRM = ETHBTC-FV, but step 6 refers to ETHBTC-FV as ETHBTC rate. 
The price feed expression in line 84 uses ETHBTC as ETHBTC-FV. What's the relation between those two values?

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+
+Add XYZ rationale including an example walk through of a funding rate application.
+
+A five minute TWAP is used to query the ETHBTC-PERP price. The TWAP is used to decrease the risks of attempted ETHBTC-PERP price manipulation.
+
+## IMPLEMENTATION
+Voters should determine which pricing implementation to use depending on when the price request was submitted.
+
+1. Following the specifications in [UMIP-2](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-2.md), query for the ETHBTC price at the disputed funding rate proposal timestamp.
+2. Query for the cumulative funding rate multiplier at the same timestamp.
```
akhanubis, line 132: How do you calculate the CFRM?

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+- Rounding: Round to nearest 18 decimal places (19th decimal place digit >= 5 rounds up and < 5 rounds down)
+- Synthetic Name: To be added
+- Synthetic Address: To be added
+- AMM Pool Address: To be added
+- AMM Pair: ETHBTC_PERP/USDC
+
+## RATIONALE
+
+To create an ETH/BTC perpetual, an ETHBTC funding rate is required. This funding rate will be used to keep the price of the ETHBTC-PERP synthetic pegged to the ETHBTC rate times the cumulative funding rate multiplier (CFRM). The funding rate will be determined with the following formula:
+- [ETHBTC-PERP - ETHBTC-FV] / ETHBTC-FV / 86400
```
chicfilabae, line 119: Similar in structure to our basis synths, thoughts around adding clamps for the funding rate, akin to many CeFi exchanges?

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+It should be noted that this is an altered price reporting from what has been used historically by UMA contracts. With the introduction of the optimistic oracle and new EMP and Perpetual contract templates, all price and funding rate requests will be handled by the Optimistic Oracle, and will only be sent to the voters of the DVM in the case of a price or funding rate dispute.
+
+*Note*:
+For future funding rates, it is likely that a more generalized funding rate identifier should be created and used. A generalized funding rate identifier would be able to support any funding rate request by making use of ancillary data. Link to UMIP section. This UMIP is reduced in scope and specifically only creates a methodology for the ETHBTC funding rate and is being proposed so that this identifier can be used as an example of how a funding rate would work for a specific synthetic. 
+
+## MARKETS & DATA SOURCES
+
+- Markets:
+
+Balancer: ETHBTC_PERP/USDC
```
chicfilabae, line 32: What's the high level trade-offs around constructing the underlying index from spot exchanges yourselves vs. leveraging the funding rate from a variety of the most liquid CeFi derivs exchanges (i.e. Bitmex, etc.)?

Presumably then you wouldn't have to worry about whether your implementation is better than the optimization around funding rate construction + index construction that other CeFi exchanges have already done over the past few years

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+
+It should be noted that this is an altered price reporting from what has been used historically by UMA contracts. With the introduction of the optimistic oracle and new EMP and Perpetual contract templates, all price and funding rate requests will be handled by the Optimistic Oracle, and will only be sent to the voters of the DVM in the case of a price or funding rate dispute.
+
+*Note*:
+For future funding rates, it is likely that a more generalized funding rate identifier should be created and used. A generalized funding rate identifier would be able to support any funding rate request by making use of ancillary data. Link to UMIP section. This UMIP is reduced in scope and specifically only creates a methodology for the ETHBTC funding rate and is being proposed so that this identifier can be used as an example of how a funding rate would work for a specific synthetic.
```
mhairimca, line 26: should there be a link in here (line 3)?

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+
+Without an expiry date to keep a synthetic token pegged to its underlying price, UMA’s perpetual contract requires that a funding rate be levied when there is a difference between the synthetic’s price and the underlying index. This will pressure the overvalued side to unwind its position or encourage the undervalued side to create a larger position. This mechanic is similar to what centralized exchanges use to keep their perpetual synthetics inline with its underlier
+
+Funding rate proposals and disputes function similarly to normal UMA price requests. At any time, a proposer can propose a new funding rate in return for a reward. If the funding rate proposal is not disputed within its liveness period, that proposed rate is then used to continuously adjust the value of 1 synthetic token. 
+
+It should be noted that this is an altered price reporting from what has been used historically by UMA contracts. With the introduction of the optimistic oracle and new EMP and Perpetual contract templates, all price and funding rate requests will be handled by the Optimistic Oracle, and will only be sent to the voters of the DVM in the case of a price or funding rate dispute.
```
Reinis-FRP, line 23: What is meant by "With the introduction of the optimistic oracle"? From [UMA FAQ](https://umaproject.org/faq.html) I understand that Optimistic Oracle = DVM which already is the current mechanism for resolving disputed price requests. Is the reference here made to new version of Optimistic Oracle? Is there a high level overview document on how new Optimistic Oracle version should be functioning?

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+- AMM Pool Address: To be added
+- AMM Pair: ETHBTC_PERP/USDC
+
+## RATIONALE
+
+To create an ETH/BTC perpetual, an ETHBTC funding rate is required. This funding rate will be used to keep the price of the ETHBTC-PERP synthetic pegged to the ETHBTC rate times the cumulative funding rate multiplier (CFRM). The funding rate will be determined with the following formula:
+- [ETHBTC-PERP - ETHBTC-FV] / ETHBTC-FV / 86400
+- `ETHBTC-FV` denotes the ETHBTC price gathered with the methodology shown in [UMIP-2](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-2.md) multiplied by the CFRM.
+- `ETHBTC-PERP` denotes the five minute TWAP of the synthetic created with this funding rate identifier. This synth will be pooled with USDC. 
+- 86400 is the number of seconds per day. Assuming all other prices stay constant, this effectively gives the funding rate per second that would need to be applied to move a synthetic token's value back to fair value in one day.
```
Reinis-FRP, line 122: What is the rationale for selecting 1 day as target period for reaching fair value? Is it universal period for all perpetuals or only for this particular price-id?

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+
+## RATIONALE
+
+To create an ETH/BTC perpetual, an ETHBTC funding rate is required. This funding rate will be used to keep the price of the ETHBTC-PERP synthetic pegged to the ETHBTC rate times the cumulative funding rate multiplier (CFRM). The funding rate will be determined with the following formula:
+- [ETHBTC-PERP - ETHBTC-FV] / ETHBTC-FV / 86400
+- `ETHBTC-FV` denotes the ETHBTC price gathered with the methodology shown in [UMIP-2](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-2.md) multiplied by the CFRM.
+- `ETHBTC-PERP` denotes the five minute TWAP of the synthetic created with this funding rate identifier. This synth will be pooled with USDC. 
+- 86400 is the number of seconds per day. Assuming all other prices stay constant, this effectively gives the funding rate per second that would need to be applied to move a synthetic token's value back to fair value in one day.  
+
+Add XYZ rationale including an example walk through of a funding rate application.
```
Reinis-FRP, line 124: Yes, this should be particularly helpful in understanding how the selected funding rate is applied:
- How the funding rate translates to CFRM?
- How CFRM is recalculated and stored?
- How CFRM is applied when evaluating potential liquidation? Is it applied to perpetual synthetic value derived from Balancer pool and original underlying price-id is not used anymore?

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+
+A five minute TWAP is used to query the ETHBTC-PERP price. The TWAP is used to decrease the risks of attempted ETHBTC-PERP price manipulation.
+
+## IMPLEMENTATION
+Voters should determine which pricing implementation to use depending on when the price request was submitted.
```
Reinis-FRP, line 129: Currently there is only one pricing implementation here. Would this UMIP be updated with secondary implementation and when that should be used?

UMIPs/ethbtc-funding-rate.md
Reinis-FRP, line 132: Is CFRM stored on-chain or voters should calculate it locally from historical funding rates?

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+## SUMMARY
+
+The DVM should support price requests for a funding rate for the ETH/BTC UMA perpetual. This funding rate represents the change in a synthetic token's value per second.  
+
+*Note*:
+This funding rate identifier requires querying for the price of the synthetic token that will be created by a contract using this identifier. Throughout this UMIP, this synthetic will be referred to as ETHBTC-PERP. Since this synthetic has not yet been created, its specific details can not yet be included. This UMIP will be updated once the ETHBTC-PERP has been created to include its address, token name and AMM pool address.
+
+## MOTIVATION
+
+Without an expiry date to keep a synthetic token pegged to its underlying price, UMA’s perpetual contract requires that a funding rate be levied when there is a difference between the synthetic’s price and the underlying index. This will pressure the overvalued side to unwind its position or encourage the undervalued side to create a larger position. This mechanic is similar to what centralized exchanges use to keep their perpetual synthetics inline with its underlier
```
Reinis-FRP, line 19: Is there an arbitrage mechanism for bringing perpetual synthetic token value back to its underlying? Or the system relies solely on sponsors optimizing for their effective collateral ratio?

What is expected synthetic price deviation from underlying as it is unlikely that any sponsor would want to mint at minimum collateral ratio and redeem/mint instantly as price changes

UMIPs/ethbtc-funding-rate.md
smb2796: Nope, I was using an old UMIP as a template and this slipped through. Thanks for pointing this out! Updated to remove.

UMIPs/ethbtc-funding-rate.md
smb2796: The Optimistic Oracle is a new settlement and price proposal layer before the DVM. It is explained in this [UMIP](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-52.md) and documentation is on its way. I will add the links to the UMIPs in the meantime. Do you think this is adequate?

Reinis-FRP: Yes, link to UMIP-52 is definitely adequate here.

Do I understand correctly that [Technical Architecture of the DVM](https://docs.umaproject.org/oracle/tech-architecture) would be revised as part of Docs update? It would be really helpful when trying to grasp the big picture.

UMIPs/ethbtc-funding-rate.md
smb2796: Yes! Thank you. Added

UMIPs/ethbtc-funding-rate.md
smb2796: The perpetual contracts themselves have clamps built in. Deployers specify a max and min funding rate on deployment, but you are right that this should probably be factored in to this calculation.

UMIPs/ethbtc-funding-rate.md
smb2796: Hm good question. These don't seem directly comparable to me though. First of all, a perp using this funding rate could conceivably trade at a very different level from something like a Bitmex perp. This could be due to liquidity differences/availability to different markets etc, and could necessitate very different funding rates imo. Also, this might not apply for this specific funding rate, but it's very possible that perps are created that are not available on CeFi exchanges - so a funding rate method will need to eventually be created.

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+
+Funding rate proposals and disputes function similarly to normal UMA price requests. At any time, a proposer can propose a new funding rate in return for a reward. If the funding rate proposal is not disputed within its liveness period, that proposed rate is then used to continuously adjust the value of 1 synthetic token.
```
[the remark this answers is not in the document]
smb2796: Yes, they need to post a bond which they lose if they are disputed successfully. Added some language in for this.

UMIPs/ethbtc-funding-rate.md
smb2796: Yes, it will be. Although the technical architecture of the DVM has largely remained the same. The Optimistic Oracle just creates another settlement layer before the DVM.

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+## SUMMARY
+
+The DVM should support price requests for a funding rate for the ETH/BTC UMA perpetual. This funding rate represents the change in a synthetic token's value per second.  
+
+*Note*:
+This funding rate identifier requires querying for the price of the synthetic token that will be created by a contract using this identifier. Throughout this UMIP, this synthetic will be referred to as ETHBTC-PERP. Since this synthetic has not yet been created, its specific details can not yet be included. This UMIP will be updated once the ETHBTC-PERP has been created to include its address, token name and AMM pool address.
+
+## MOTIVATION
+
+Without an expiry date to keep a synthetic token pegged to its underlying price, UMA’s perpetual contract requires that a funding rate be levied when there is a difference between the synthetic’s price and the underlying index. This will pressure the overvalued side to unwind its position or encourage the undervalued side to create a larger position. This mechanic is similar to what centralized exchanges use to keep their perpetual synthetics in line with its underlier
```
chrismaree, line 19:
```suggestion
Without an expiry date to keep a synthetic token pegged to its underlying price, UMA’s perpetual contract requires that a funding rate be levied when there is a difference between the synthetic’s price and the underlying index. This will pressure the overvalued side to unwind its position or encourage the undervalued side to create a larger position. This mechanic is similar to what centralized exchanges use to keep their perpetual synthetics in line with its underlier.
```

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+
+Funding rate proposals and disputes function similarly to normal UMA price requests. On request, a proposer can post a bond and propose a new funding rate in return for a reward. If the funding rate proposal is not disputed within its liveness period, that proposed rate is then used to continuously adjust the value synthetic token sponsors' debt. If the proposed funding rate is successfully disputed, the proposer will lose their bond. The bond amount varies between each perpetual contract.
+
+It should be noted that this is an altered price reporting from what has been used historically by UMA contracts. With the introduction of the [Optimistic Oracle](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-52.md) and new [EMP](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-54.md) and [Perpetual contract](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-53.md) templates, all price and funding rate requests will be handled by the Optimistic Oracle, and will only be sent to the voters of the DVM in the case of a price or funding rate dispute.
```
chrismaree, line 23: the new EMP contracts will still use the DVM for liquidation disputes.

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+To determine the ETHBTC_PERP synth price on Balancer, the [BalancerPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js) is used.
+
+To combine these rates in a mathematical expression, the [ExpressionPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js) is used.
+
+Because this uses existing price feeds, the only additional work that is required is adding a new [default price feed config](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js). This cannot be added to the protocol repo until this perpetual synthetic is created, because the price feed config will require the related pool address. It should follow this pattern:
+
+```
+ETHBTC_FR: {
+    type: "expression",
+    expression: "(ETHBTC_PERP - ETH/BTC) / ETH/BTC / 31536000",
```
chrismaree, line 84: please explain `31536000` magic number.

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+- AMM Pool Address: To be added
+- AMM Pair: ETHBTC_PERP/USDC
+
+## RATIONALE
+
+To create an ETH/BTC perpetual, an ETHBTC funding rate is required. This funding rate will be used to keep the price of the ETHBTC-PERP synthetic pegged to the ETHBTC rate times the cumulative funding rate multiplier (CFRM). The funding rate will be determined with the following formula:
+- [ETHBTC-PERP - ETHBTC-FV] / ETHBTC-FV / 31536000
+- `ETHBTC-FV` denotes the ETHBTC price gathered with the methodology shown in [UMIP-2](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-2.md) multiplied by the CFRM.
+- `ETHBTC-PERP` denotes the five minute TWAP of the synthetic created with this funding rate identifier. This synth will be pooled with USDC. 
+- 31536000 is the number of seconds in a year. Assuming all other prices stay constant, this effectively gives the funding rate per second that would need to be applied to move a synthetic token's value back to fair value in one year.
```
chrismaree, line 122: this is a great point here and is something that I've always had a hard time quantifying with funding rates. What this means effectively is that if I'm on the wrong side of the trade (holding an undervalued synth for example) I'd need a wait in this position for a full year until the funding rate "corrected" the miss pricing in the synth to make me "whole". 

Would it make sence to decrease this duration to make the "correction" factor happen faster? maybe like a month? What is the argument for setting it to 1 year in duration? is this perhaps a too weak force to ensure a strong peg? 

Clearly, we will need to experiment and test to find the right duration. I'd just like to understand why we are starting with this duration vs another duration.

UMIPs/ethbtc-funding-rate.md
smb2796: Yes good point. I updated the language to reflect this

UMIPs/ethbtc-funding-rate.md
```diff
@@ -0,0 +1,146 @@
...
+- Rounding: Round to nearest 18 decimal places (19th decimal place digit >= 5 rounds up and < 5 rounds down)
+- Synthetic Name: To be added
+- Synthetic Address: To be added
+- AMM Pool Address: To be added
+- AMM Pair: ETHBTC_PERP/USDC
+
+## RATIONALE
+
+To create an ETH/BTC perpetual, an ETHBTC funding rate is required. This funding rate will be used to keep the price of the ETHBTC-PERP synthetic pegged to the ETHBTC rate times the cumulative funding rate multiplier (CFRM). The funding rate will be determined with the following formula:
+- [ETHBTC-PERP - ETHBTC-FV] / ETHBTC-FV / 31536000
```
Tommy1231232, line 119: Can we make this more readable? Don't want people to get confused thinking the dash means to subtract.  E.g., [[ETHBTC-PERP] - [ETHBTC-FV]] / [ETHBTC-FV] / 3153600 Note brackets are names

UMIPs/ethbtc-funding-rate.md
Reinis-FRP: I also consider 1y correction period too long given the volatility of most crypto assets.
If I understand correctly, the only force how the funding rate brings perp in sync with underlying is by applying CFRM when evaluating collateralization for potential liquidation - hence, it is really a soft peg, so the force of funding rate should be quite strong. Hence, I am in favor of 1week - max 1month adjustment periods. It could be even shorter, but we also do not want to put too much burden and gas costs for funding rate proposers.

smb2796: I agree that this period needed to be reduced. After some discussion, I updated this to be 1 day instead of 1 year. This is modeled after FTX's funding rate calculation.

I don't think the gas costs for funding rate proposers should be too much of a concern. Funding rate proposers are given a reward for proposing a correct funding rate. This should make up for the gas costs.

UMIPs/ethbtc-funding-rate.md
smb2796: Yep, it is stored on-chain. Added some directions for how to get the CFRM.

UMIPs/ethbtc-funding-rate.md
smb2796: This was a mistake, so thanks for pointing it out! Updated to use ETHBTC-FV instead of ETHBTC in this step.

UMIPs/ethbtc-funding-rate.md
smb2796: Added some instructions for this within the `Implementation` section.

UMIPs/ethbtc-funding-rate.md
```diff
@@ -115,32 +120,31 @@ ETHBTC_FR: {
...
+1. Following the specifications in [UMIP-2](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-2.md), query for the 1-hour ETHBTC TWAP, ending at the disputed funding rate proposal timestamp.
+2. For each distinct price query in the 1-hour TWAP, query for the cumulative funding rate multiplier (CFRM) at the same timestamps.
+3. For each timestamp within the 1-hour TWAP, the CFRM and ETHBTC rate should be multiplied to get the 1-hour TWAP of ETHBTC * CFRM. 
+4. Query for the Token Scaling Multiplier (TSM) for the ETHBTC_PERP, and multiply this by the result from step 3 to get the ETHBTC-FV.
+5. Query for the ETHBTC-PERP 1-hour TWAP from the listed AMM pool. This will return the ETHBTC-PERP's TWAP denominated in USDC.
+6. Subtract the result of step 4 from the result of step 5. [ETHBTC-PERP - ETHBTC-FV].
+7. Divide the result of step 6 by the ETHBTC-FV rate from step 4. [ETHBTC-PERP - ETHBTC-FV]/ETHBTC-FV.
+8. Divide the result from step 6 by 86400 (# of seconds in a day) to get the funding rate per second. This should then be multiplied by -1. Voters should then round this result to 18 decimal places.
```
akhanubis, line 141: I think it should be "Divide the result from step 7"

smb2796: Good catch. Updated

UMIPs/ethbtc-funding-rate.md
```diff
@@ -115,32 +120,31 @@ ETHBTC_FR: {
...
-2. Query for the cumulative funding rate multiplier at the same timestamp.
-3. Multiply the ETHBTC price from step 1 by the CFRM from step 2 to get the fair value of the ETHBTC synthetic (ETHBTC-FV). [ETHBTC * CFRM].
-4. Query for the ETHBTC-PERP 5 minute TWAP from the listed AMM pool. This will return the ETHBTC-PERP's TWAP denominated in USDC.
-5. Subtract the result of step 3 from the result of step 4. [ETHBTC-PERP - ETHBTC-FV].
-6. Divide the result of step 5 by the ETHBTC-FV rate from step 3. [ETHBTC-PERP - ETHBTC-FV]/ETHBTC-FV.
-7. Divide the result from step 6 by 31536000 (# of seconds in a year) to get the funding rate per second. Voters should then round this result to 18 decimal places.
+1. Following the specifications in [UMIP-2](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-2.md), query for the 1-hour ETHBTC TWAP, ending at the disputed funding rate proposal timestamp.
+2. For each distinct price query in the 1-hour TWAP, query for the cumulative funding rate multiplier (CFRM) at the same timestamps.
```
Reinis-FRP, line 135: Does it imply querying CFRM for the end of each 60 second interval within the 1-hour TWAP period?

smb2796: Yes it does and good question. Updated to clarify this.

UMIPs/ethbtc-funding-rate.md
smb2796: Added some clarification on the rationale for choosing `86400`

mhairimca (approved):

PR #204, opened 2021-03-08
Add uTVL_KPI_UMA  as a Price Identifier 

mhairimca:


UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,352 @@
+
+## HEADERS
+| UMIP [#]     |                                                                                                                                  |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | [Key Performance Indicator; Total Value Locked UMA Price Identifier                                                                                                  |
```
smb2796, line 5: To follow the existing convention, the title should just be: "Add uTVL_KPI_UMA as a price identifier".

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,370 @@
...
+https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-7.md
+Should other non-algorithmic derivations of BTC to be added to the accepted collateral currency whitelist prior to expiry, these should also be similarly grouped and  follow the calculation methods detailed in UMIP 7, see rationale for further details.
+Note that the price identifier of UMIP 7 is depreciated, see rationale for further discussion
+
+
+	3. USD 	- assumed to be exactly 1USD. 
+Should other non-algorithmic USD stablecoins be added to the accepted collateral currency whitelist prior to expiry, these should also be similarly grouped and assumed to be worth exactly 1USD
+Note - See rationale for further discussion
+
+	4. uLQ 	- Uniswap (calculations performed as per UMIP 59)
```
smb2796, line 98: What is uLQ and sLQ?

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,370 @@
...
+ - The use of these dashboards was considered but rejected due to the potential for manipulation and the differences between their calculations of the values of the collateral assets and the value as determined by the relevant UMIPs.
+
+
+<br>
+
+# IMPLEMENTATION
+
+ - The dollar value of each of the contracts should be calculated using the UMIPs and guidence in the Markets and Data section.  
+ - These should then be summed to obtain the total value locked (TVL) measured in dollars.
+ - The TVL as measured in dollars should then be divided by 10^8
```
smb2796, line 315: Shouldn't this be divided by 10^9, not 10^8? I think it'd be clearer to just say that this needs to be divided by 1 billion

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,370 @@
...
+ - These currently monitor different contracts and calculate the value slightly differently.
+
+ - The use of these dashboards was considered but rejected due to the potential for manipulation and the differences between their calculations of the values of the collateral assets and the value as determined by the relevant UMIPs.
+
+
+<br>
+
+# IMPLEMENTATION
+
+ - The dollar value of each of the contracts should be calculated using the UMIPs and guidence in the Markets and Data section.
```
smb2796, line 313: I think we need to cover which contracts need to be used in this calculation within this section. For right now, I believe the rules would be:

1. Any EMP contract
2. Any Perpetual contract
3. Any perpetual pool party contract

smb2796, line 313: Let's also be explicit about the fact that we are using the total value locked of the collateral currency in each contract.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,370 @@
...
+ - YAM, 
+ - Aave, 
+ - Link, 
+ - SNX, 
+ - UMA, 
+ - UNI
+
+It is proposed that these are treated in the above groups for the purposes of determining markets and data sources.  For discussion on this see #Rationale 
+
+ **Required questions**
```
smb2796, line 80: This is all good information, but I wonder if it is overcomplicating things. Perhaps we should just define some heuristics around how to choose the pricing source for each collateral. It is likely that there will be additional collateral types that are not covered by this list, and defining rules rather than explicit instructions for each would cover these.

What do you think about?
- For collateral types that have a Collateral/USD price identifier already, use the instructions from that price id.
- For types that do not, use the price from the highest 24 hour volume exchange.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: Uniswap Liquidity tokens and Sushiswap liquidity tokens 

I'm suggesting that any other Uniswap or Sushiswap LQ tokens that are approved as collateral types follow these conventions unless there is good reason not to.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: Oh, yeah, thats changed from the original KPI options paper.

Will update, I've referred to 100 million in a couple of places.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,370 @@
...
+
+	4. uLQ 	- Uniswap (calculations performed as per UMIP 59)
+https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-59.md
+Should other Uniswap liquidity tokens be added to the accepted collateral currency whitelist prior to expiry, these should also be similarly calculated.
+Note - see rationale for further discussion
+
+
+	5. sLQ 	- Sushiswap (calculations performed as per UMIP 39)
+https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-39.md
+Should other Uniswap liquidity tokens be added to the accepted collateral currency whitelist prior to expiry, these should also be similarly calculated.
```
akhanubis, line 106: Should say Sushiswap

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,370 @@
...
+Should other non-algorithmic derivations of BTC to be added to the accepted collateral currency whitelist prior to expiry, these should also be similarly grouped and  follow the calculation methods detailed in UMIP 7, see rationale for further details.
+Note that the price identifier of UMIP 7 is depreciated, see rationale for further discussion
+
+
+	3. USD 	- assumed to be exactly 1USD. 
+Should other non-algorithmic USD stablecoins be added to the accepted collateral currency whitelist prior to expiry, these should also be similarly grouped and assumed to be worth exactly 1USD
+Note - See rationale for further discussion
+
+	4. uLQ 	- Uniswap (calculations performed as per UMIP 59)
+https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-59.md
```
akhanubis, line 99: This UMIP doesn't define LP/USD, only USD/LP. LP/USD is the result of the previous to last step of the implementation, so I'd add the step number to this reference.

It also applies to 7. bBadger - Sushiswap

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,370 @@
...
+Should other Uniswap liquidity tokens be added to the accepted collateral currency whitelist prior to expiry, these should also be similarly calculated.
+Note - see rationale for further discussion
+
+
+	5. sLQ 	- Sushiswap (calculations performed as per UMIP 39)
+https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-39.md
```
akhanubis, line 105: I don't think this UMIP is the best example for SLPs calculation because the SLPs in UMIP 39 are wrapped through a  Badger Sett vault so the calculation takes into account the pricePerFullShare of the vault.

 I think Uniswap and Sushiswap calculations can be unified, they are the same implementation wise.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,370 @@
...
+
+3. Four non-algorithmic dollar stablecoins (DAI, USDC, USDT, rDai]) are approved as collateral within the UMA Protocol. 
+It is assumed for this purpose that 1DAI=1USDC=1USDT=1rDAI=$1.  While this assumption may be somewhat inaccurate, it has minimal impact on the overall calculation, cuts down on processing complexity and works on a “common sense” attitude that a dollar stablecoin is worth exactly $1.
+There are other forms of dollar stablecoins which are not approved as collateral within the UMA protocol.  To maintain consistency it is suggested that this assumption is made for any other non-algorithmic stablecoin. 
+
+4. Four Uniswap Liquidity Tokens (wBTC-Eth, USDC-Eth, UNI-Eth, UMA-Eth) are approved as collateral within the UMA protocol. 
+It is recommended that any additional proposal to add Uniswap liquidity tokens as approved collateral currencies are evaluated against UMIP 59 and follow those market and data sources unless there is good reason for deviance
+
+5. One Sushisaw Liquidity Token ( [bwBTC/ETH SLP) is approved as collateral in the UMA Protocol.
```
prabhaav, line 294: Sushisaw -> sushiswap

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,370 @@
...
+
+ - There is no need for price processing.  This is a snapshot based on a particular time, however it may be useful for TVL Options holders to have oversight of the ongoing TVL and consequently the value of their options on an ongoing basis.  There are currently two dashboards that track the TVL of UMA - 
+	 - [SimpleID](https://monitor.simpleid.xyz/uma)
+ 	- [Yuen](https://docs.google.com/spreadsheets/d/e/2PACX-1vSEMURxiVQuu6jSDp2zmI7kdKKaJjgmhWNiVjwStyJekDx9hWgclKzm_yv9iyj82IRP4d9dZ8rgvCCB/pubhtml) 
+
+ - These currently monitor different contracts and calculate the value slightly differently.
+
+ - The use of these dashboards was considered but rejected due to the potential for manipulation and the differences between their calculations of the values of the collateral assets and the value as determined by the relevant UMIPs.
```
prabhaav, line 306: What methodologies do the relevant UMIPs use to calculate the value?

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: Whats a perpetual pool party contract?  Is that the optimistic oracle?

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: This doesnt work for Liquidity tokens,  but yes....this can be simplified a lot.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: There is a bit of variety in markets and data sources.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: Have added that in.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: I dont think its relevant for markets and data (as the same sources apply), however have added it into the rationale to be explicit about where the calculation should come from.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: I've stuck with the Weth, BTC and USD stablecoin definitions and grouped the liquidity tokens together to follow UMIP 59.  The rest can be either UMIPs or derived from highest volume dollar markets. I've also added in the rationale notes that where there is no dollar market, the highest volume eth market should be used, then dollar value calculated from UMIP 6.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+
+    - Divide TVL by 10^9 and apply the food and ceiling price rounding.
+
+<br>
+
+# Security considerations
+
+1. Where could manipulation occur?
+
+ - Negligible opportunities for manipulation.
```
nicholaspai, line 286: Is the TVL calculated as a TWAP at all? Otherwise I could see a pretty simple manipulation where a KPI holder deposits a ton of collateral into a contract last second before the snapshot and then withdraws. @chrismaree and I have been discussing this offline.

smb2796: I'm in favor of this, but think the TWAP would need to be considerably long to prevent this manipulation. If we did something like 2 hours, is it really that much more painful to just keep the capital in the contracts for 2-hours rather than a few seconds? Something like a week might make sense.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: Have combined uniswap and sushiswap calculations to follow UMIP 59

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: I think this is a good point, but unsure how to resolve given the variety of markets.

Chris suggested that we could use coingecko for individual ERC20s but there is the problem of LP tokens which dont have a dollar price feed.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+
+11.  What would be the cost of sending 15,000 queries?
+
+ - There is no need to sent 15, 000 queries for this price identifier as it does not require bots. 
+
+<br>
+
+# PRICE FEED IMPLEMENTATION
+
+As there is no requirement for ongoing monitoring through liquidation or dispute bots, a price feed is not required.  The only requirement is a query of the UMA TVL statistic at the timestamp 00.00(UTC) on 30th June according to the data and markets as defined above.
```
akhanubis, line 162: So Risk Labs is the only allowed sponsor? Is this price id a one use price id (based on synth expiry time being hardcoded into the price id spec itself and the fact that proper collateralization up to 2 UMA price is a given by Risk Labs)?

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+ - These should then be summed to obtain the total value locked (TVL) measured in dollars.
+
+
+4. **Result processing** 
+
+    - Divide TVL by 10^9 and apply the food and ceiling price rounding.
+
+<br>
+
+# Security considerations
```
Reinis-FRP, line 282: What are implications of rogue liquidation attacks on the KPI options program? E.g. if at some point TVL reaches $2bn, and someone tries liquidating part of sponsors' debt position at uTVL_KPI_UMA>2, which is disputed and DVM returns price of 2 - do I understand correctly, that the sponsor's debt position would still be settled at 2? Now, if at expiration TVL has fallen (e.g. to $1bn), the sponsor would not be able to get out excess collateral at price of 1 because part of debt already has been settled for the price of 2.
Sure, there is no economic gain for the liquidator (just losses), but is there a risk that part of UMA collateral could get stuck in the contract?

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+There are other forms of wrapped BTC which are not approved as collateral within the UMA protocol.  To maintain consistency it is suggested that this assumption is made for any other similar non-algorithmic wrapped Bitcoin.  
+
+3. Four non-algorithmic dollar stablecoins (DAI, USDC, USDT, rDai]) are approved as collateral within the UMA Protocol. 
+It is assumed for this purpose that 1DAI=1USDC=1USDT=1rDAI=$1.  While this assumption may be somewhat inaccurate, it has minimal impact on the overall calculation, cuts down on processing complexity and works on a “common sense” attitude that a dollar stablecoin is worth exactly $1.
+There are other forms of dollar stablecoins which are not approved as collateral within the UMA protocol.  To maintain consistency it is suggested that this assumption is made for any other non-algorithmic stablecoin. 
+
+4. FiveLiquidity Tokens (wBTC-Eth, USDC-Eth, UNI-Eth, UMA-Eth, bwBTC/ETH_SLP) are approved as collateral within the UMA protocol.
```
akhanubis, line 229: I think bwBTC/ETH_SLP should be moved to the native ERC20 tokens because of the extra step of pricePerFullShare. Using UMIP59 to calculate bwBTC/ETH_SLP/USD would return a lower price (higher if the badger sett gets drained) than the real one.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
akhanubis: Wouldn't applying full TWAP to the final price (aka to every underlying price) be cumbersome for sources that don't natively provide it?

Can averaging snapshots over the suggested period alleviate this (although exploiter could easily jump in and out)?

What about using random snapshots over the period, where the random timestamps are decided after expiry via any public shareable source of randomness (not sure how good the entropy would be, but something like using the hash of the first block after expiry as seed)? So everyone knows how to calculate the random TS beforehand but nobody knows what the chosen TS will be up until expiry.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+    - From the list of ExpiringMultiPartyCreator (available at https://github.com/UMAprotocol/protocol/blob/master/packages/affiliates/payouts/devmining-status.json), query all EMP addresses from createExpiringMultiParty() and call pfc() on all of them, repeat same steps for PerpetualCreator and just check the createPerpetual()  event, to identify the amount of each collateral type locked in UMA contracts. 
+
+ - The dollar value of each contract should then be calculated using the details supplied in the Markets and Data section referencing the relevant UMIPs
+
+ - These should then be summed to obtain the total value locked (TVL) measured in dollars.
+
+
+4. **Result processing** 
+
+    - Divide TVL by 10^9 and apply the food and ceiling price rounding.
```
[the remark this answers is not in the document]
mhairimca: oops!

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: It seems unlikely given the lack of economic gain, but in such a case we could settle the dispute about which is the correct TVL using the DVM

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: I think it is a one-use UMIP, if we were to create new KPI options we would probably put forward a new PI

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: It could go in as a native ERC20 token section to use UMIP39, or grouped together with the other LPs, as per UMIP 59. 

Which do you think is better?

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
mhairimca: See comment above.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+ - The dollar value of each of the contracts should be calculated using the UMIPs and guidence in the Markets and Data section. 
+ - Where there is no price identifier UMIP, the price should be queried from the highest volume USD* market
+
+2. **Pricing interval**
+
+    - This will vary for each value however the relevant UMIPs provide a guide to the pricing interval for each
+
+3. **Input processing**
+
+    - From the list of ExpiringMultiPartyCreator (available at https://github.com/UMAprotocol/protocol/blob/master/packages/affiliates/payouts/devmining-status.json), query all EMP addresses from createExpiringMultiParty() and call pfc() on all of them, repeat same steps for PerpetualCreator and just check the createPerpetual()  event, to identify the amount of each collateral type locked in UMA contracts.
```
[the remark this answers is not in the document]
mhairimca: @daywiss could you comment?
@smb2796 suggested this method and I'm not sure how practical the proposed alternative above is.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+## HEADERS
+| UMIP [#]     |                                                                                                                                  |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | [Add uTVL_KPI_UMA  as a Price Identifier
+                                                                                                  |
+| Authors    | [Hart Lambur hart@umaproject.org, Mhairi McAlpine mhairi@umaproject.org]
+	       
+| Status     | [Draft]
+
+| Created    | [4th March 2021]
```
smb2796, line 11: The formatting here is a little off. Suggest pulling from another UMIP

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+| Created    | [4th March 2021] 
+
+| [Link to Discourse   -  https://discourse.umaproject.org/t/add-utvl-kpi-uma-as-a-price-identifier/318]                                                                               
+<br>
+<br>
+<br>
+
+# SUMMARY 
+
+Key Performance Indicator(KPI) options are synthetic tokens that redeem to protocol tokens with the redemption value determined by performance against that indicator. One example of a KPI is Total Value Locked (TVL).
```
smb2796, line 20:
```suggestion
Key Performance Indicator (KPI) options are synthetic tokens that redeem to protocol tokens with the redemption value determined by performance against that indicator. One example of a KPI is Total Value Locked (TVL).  
```

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+| [Link to Discourse   -  https://discourse.umaproject.org/t/add-utvl-kpi-uma-as-a-price-identifier/318]                                                                               
+<br>
+<br>
+<br>
+
+# SUMMARY 
+
+Key Performance Indicator(KPI) options are synthetic tokens that redeem to protocol tokens with the redemption value determined by performance against that indicator. One example of a KPI is Total Value Locked (TVL).  
+
+This UMIP enables the DVM to support price requests based on the TVL of UMA
```
smb2796, line 22:
```suggestion
This UMIP enables the DVM to support price requests based on the TVL of UMA.
```

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+
+A synthetic option is minted against a base collateral, in this case UMA, which expires at 00.00(UTC) June 30th 2021.
```
smb2796, line 24: I don't think this would be specific to a certain expiry date, would it?

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+The value locked is calculated using the methods included in previous UMIPs which have price identifies associated with collateral types where available, collateral types which do not have an associated Price Identifier, or which are not yet approved will follow these presidents where available. 
+
+The dollar value of each of the collateral types is then summed to provide the total value locked.
+
+  
+
+
+# MOTIVATION
+
+The primary motivation for the development of KPI options is to allow protocols to incentivise Defi users to assist them to reach the protocol's identified targets by leveraging their community resources by sharing their value with their community members.
```
smb2796, line 37:
```suggestion
The primary motivation for the development of KPI options is to allow protocols to incentivize DeFi users to assist them to reach the protocol's identified targets by leveraging their community resources by sharing their value with their community members.  
```

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+The dollar value of each of the collateral types is then summed to provide the total value locked.
+
+  
+
+
+# MOTIVATION
+
+The primary motivation for the development of KPI options is to allow protocols to incentivise Defi users to assist them to reach the protocol's identified targets by leveraging their community resources by sharing their value with their community members.  
+
+Total Value Locked(TVL) is a frequently quoted key performance indicator and one which has a level of prominence in key Defi dashboards as an indicator of the health of a protocol.  Within UMA, TVL plays the additional function of being the lower floor for the value of 0,5 $UMA, due to the design of UMA and the COC>POC inequality.
```
smb2796, line 39:
```suggestion
Total Value Locked(TVL) is a frequently quoted key performance indicator and one which has a level of prominence in key DeFi dashboards as an indicator of the health of a protocol.  Within UMA, TVL plays the additional function of being the lower floor for the value of 0.5 $UMA's market cap, due to the design of UMA and the COC>POC inequality.
```

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+
+1. What are the financial positions enabled by creating this synthetic that do not already exist?
+
+  - The DVM does not currently support any form of KPI options.  This synthetic token will allow the creation of tokens which expire to a set rate of the collateral asset tokens based on a pre-identified bounded ratio as determined by the TVL of the protocol at the time of expiry.
+
+2. Please provide an example of a person interacting with a contract that uses this price identifier. 
+
+ - A protocol may choose to leverage its community and/or its reputation by minting TVL Options for its token which can be redeemed to a token amount as determined by the TVL of the protocol at the expiry point.
+ - A protocol community member, tokenholder, voter or proximal Defi prtocol participant may be gifted a TVL option by a protocol as an incentive to build the TVL of the protocol within the option timeframe and redeem at expiry.
```
smb2796, line 48:
```suggestion
 - A protocol community member, tokenholder, voter or proximal DeGi protocol participant may be gifted a TVL option by a protocol as an incentive to build the TVL of the protocol within the option timeframe and redeem at expiry.
```

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+# MARKETS & DATA SOURCES
+
+There are a variety of assets approved as Collateral within UMA's priceless contracts.
+A list of these assets is [available on our docs site](https://docs.umaproject.org/uma-tokenholders/approved-collateral-currencies).  To maintain consistency with existing price identifier UMIPs, it is suggested that different markets are queried for different collateral types. These are grouped below. 
+
+ - [wEth] - the wrapped native token of the Ethereum Network (henceforth referred to as ETH)
+ - [renBTC, wBTC] - wrapped ERC20 versions of Bitcoin (henceforth referred to as BTC)
+ - [DAI, USDC, USDT, rDai]  - ERC20 dollar stable coins (henceforth referred to as USD)
+
+Additionally there are a number of liquidity tokens that are accepted as collateral, as well as "native" ERC20 tokens.
```
smb2796, line 64:
```suggestion
Additionally there are a number of LP tokens that are accepted as collateral, as well as "native" ERC20 tokens.
```

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+Note - see rationale for further discussion
+
+
+	5. Native ERC20 Tokens 
+		- where a price identifier UMIP exists that defines a dollar value, this should be used.
+		- where no price identifier exists,the highest volume USD* markets should be used.
+
+**Note - see rationale for further discussion**
+
+2.  Which specific pairs should be queried from each market?
```
smb2796, line 99: Suggest grouping all of these sections as the answers are all the same

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
smb2796: We should also list this [script](https://github.com/UMAprotocol/protocol/blob/master/packages/merkle-distributor/kpi-options-helpers/calculate-uma-tvl.ts), which calculates UMA TVL.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+
+**5. Collateral Decimals** 
+
+ - 18
+
+
+**6. Rounding** - Round to 2 decimal places.
+
+ - If the price is 1.025, then this would be rounded up to 1.03. If the price is 1.0249999, then this would be rounded down to 1.02.
+ - if the value returned is less than 0.05 round up to 0.1 to provide a floor price.
```
smb2796, line 206:
```suggestion
 - if the value returned is less than 0.01 round up to 0.1 to provide a floor price.
```

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
```diff
@@ -0,0 +1,310 @@
...
+
+
+<br>
+
+# IMPLEMENTATION
+
+ - The contracts to be tracked include 
+	- Any EMP contract
+ 	- Any Perpetual contract
+	- Any perpetual pool party contract
```
smb2796, line 250: Are we including perpetual pool party contracts? I think this is a bit of an open question

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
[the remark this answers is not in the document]
smb2796: That is actually a list of ExpiringMultiParty's, not the Creator. Using this list is unnecessary since we are finding all EMP's from the creator.

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
Reinis-FRP: What if the value returned is between 0.01 and 0.1? I understood that everything below 0.1 should be rounded up to 0.1 for floor price

UMIPs/Add uTVL_KPI_UMA  as a Price Identifier.md
akhanubis: I think it should go as native linked with UMIP39, because it's slightly diff than a standard LP token.

smb2796 (approved):
I think a few things need to be cleaned up, but going to merge in and mark as Last Call while the outstanding issues are addressed!

PR #205, opened 2021-03-09
Add DPI and INDEX as collateral

GottliebFreudenreich:


UMIPs/umip-60.md
```diff
@@ -0,0 +1,67 @@
...
+
+ [dpi]: https://etherscan.io/token/0x1494CA1F11D487c2bBe4543E90080AeBa4BA3C2b
+
+---
+
+## Rationale
+The rationale behind this change is giving deployers more useful collateral currency options. This is an advancement into a new types of collateral and would allow DPI and INDEX holders to create options and synthetic assets within the UMA ecosystem.
+
+$400 USD equivalent was chosen as the final fee because it is equal to or above the mimimum of already approved coins.
+
```
mhairimca, line 56: Final fee should be specified in units of the collateral currency.

UMIPs/umip-60.md
```diff
@@ -0,0 +1,67 @@
...
+## Implementation
+
+This change has no implementation other than proposing the two aforementioned governance transactions that will be proposed.
+
+### Security Considerations
+
+Since the underlying tokens are persistently valuable tokens, including the token should impose no additional risk to the protocol.
+
```
mhairimca, line 64: What about a scenario where one of the tokens within the index falls to zero.
I'm guessing that wouldnt actually pose a risk, but it would be good to  note how the index would work in this scenario.

UMIPs/umip-60.md
```diff
@@ -0,0 +1,67 @@
...
+| Authors    | Gottlieb Freudenreich (gottlieb.freudenreich@gmail.com)
+| Status     | Proposal |
+| Created    | 2020-03-08 |
+| Discourse Link | https://discourse.umaproject.org/t/add-index-and-dpi-as-collateral/326
+
+# Summary
+
+This UMIP will add INDEX and DPI to the supported collateral currencies on the global whitelist contract, allowing the usage of these 2 assets as collateral currencies.
+More information on INDEX/DPI can be found on the website: https://www.indexcoop.com/
+
```
mhairimca, line 14: Edit to "INDEX and DPI" to clarify that this is each individual token, not the market between them?

GottliebFreudenreich: Okay, I'll fix that :)

UMIPs/umip-60.md
GottliebFreudenreich: I specified units in line 28 and line 46. This was more about explaining my rationale behind that. But I can also edit it here.

UMIPs/umip-60.md
GottliebFreudenreich: Thanks for addressing this! If a token collapses, the DPI value would drop by the value of that token within DPI.
The fund only allows minining and withdrawing in the token number ratios with all the tokens being put in / withdrawn in the ratio set by the monthly balance. This is unlike an AMM / balance pool type structure where people can deposit a single token and remove others.

UMIPs/umip-60.md
```diff
@@ -0,0 +1,69 @@
...
+This UMIP will add INDEX and DPI to the supported collateral currencies on the global whitelist contract, allowing the usage of these 2 assets as collateral currencies.
+More information on INDEX and DPI can be found on the website: https://www.indexcoop.com/
+
+# Proposed Collateral Currencies
+
+## INDEX (Index Cooperative Token)
+### Motivation 
+
+INDEX is the governance token which presides over the Index Cooperative the curater of the DeFi Pulse Index. 
+The Index Cooperative has 130 million TVL and could utilize its governance token and/or treasury funds as collateral within the UMA ecosystem.
```
Reinis-FRP, line 21: 130 million TVL is USD currency?

GottliebFreudenreich: Thanks for pointing it out. Fixed.

UMIPs/umip-60.md
mhairimca: I think that would probably be better just for clarity.

smb2796 (approved):
Looks good to go

PR #206, opened 2021-03-09
Add DPI and INDEX as Price Identifier in ETH and USD

GottliebFreudenreich:


UMIPs/umip-61.md
```diff
@@ -0,0 +1,377 @@
...
+
+# IMPLEMENTATION
+
+**For INDEX/ETH and ETH/INDEX**
+
+    1. Query INDEX/ETH Price from Uniswap using 1 minute TWAP (0x3452a7f30a712e415a0674c0341d44ee9d9786f9).
+    2. Query INDEX/ETH Price from Sushiswap using 1 minute TWAP (0xa73df646512c82550c2b3c0324c4eedee53b400c).
+    3. Query INDEX/ETH Price from Balancer using 1 minute TWAP (0xcf19a7c81fcf0e01c927f28a2b551405e58c77e5).
+    4. Take the median of prices acquired from steps 1, 2, and 3 and round to 18 decimals get the final INDEX/ETH price.
+    5. (for ETH/INDEX) Take the Inverse of the result of step 7 (1/ INDEX/ETH) and round to 18 decimals to get the ETH/INDEX price.
```
Reinis-FRP, line 337: Probably reference to step 4 for INDEX/ETH

Reinis-FRP, line 337: use inverse of INDEX/ETH before rounding and round the inverse or inverse the rounded INDEX/ETH and then round again?
The same question applies to all other inverse pairs below

GottliebFreudenreich: Sorry.. Changed that

GottliebFreudenreich: I would inverse the rounded INDEX/ETH and then round again. I'm open for other suggestions.

UMIPs/umip-64.md
```diff
@@ -0,0 +1,377 @@
...
+
+## INDEX/USD 
+
+**1. Price Identifier Name** - INDEXUSD
+
+**2. Base Currency** - INDEX
+
+**3. Quote currency** - USD
+
+**4. Intended Collateral Currency** - USDT
```
akhanubis, line 149: It's an intent, so it's not final/blocking, but is there a particular reason to choose USDT over DAI or USDC?

UMIPs/umip-64.md
```diff
@@ -0,0 +1,377 @@
...
+
+## INDEX/ETH
+
+**1. Price Identifier Name** - INDEXETH
+
+**2. Base Currency** - INDEX
+
+**3. Quote currency** - ETH
+
+**4. Intended Collateral Currency** - ETH
```
akhanubis, line 199: It should be WETH. Same for DPI/ETH

UMIPs/umip-64.md
```diff
@@ -0,0 +1,377 @@
...
+
+    - No limit currently
+
+11.  What would be the cost of sending 15,000 queries?
+
+     - $0
+
+# PRICE FEED IMPLEMENTATION
+
+These price identifiers use the [UniswapPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js), [BalancerPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js) and [ExpressionPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js).
```
akhanubis, line 113: Should MedianizerPriceFeed be included here considering that implementations takes the median from the 3 sources?

UMIPs/umip-64.md
```diff
@@ -0,0 +1,377 @@
...
+    5. (for ETH/INDEX) Take the Inverse of the result of step 4 (1/ INDEX/ETH) and round to 18 decimals to get the ETH/INDEX price.
+
+**For INDEX/USD and USD/INDEX** 
+
+    1. Query INDEX/ETH Price from Uniswap using 1 minute TWAP (0x3452a7f30a712e415a0674c0341d44ee9d9786f9).
+    2. Query INDEX/ETH Price from Sushiswap using 1 minute TWAP (0xa73df646512c82550c2b3c0324c4eedee53b400c).
+    3. Query INDEX/ETH Price from Balancer using 1 minute TWAP (0xcf19a7c81fcf0e01c927f28a2b551405e58c77e5).
+    4. Query the ETH/USD Price as per UMIP-6.
+    5. Multiply the INDEX/ETH prices in steps 1, 2 and 3 by the ETH/USD price to get the respective INDEX/USD prices.
+    6. Take the median of prices acquired from step 4 and round to 6 decimals to get the INDEX/USD price.
```
akhanubis, line 346: From step 5? Same for DPI/USD

UMIPs/umip-61.md
akhanubis: My personal opinion is to leave rounding as the last step to avoid carrying inaccuracies over to the following steps, but I'm not sure what the general consensus is.
I would split step 4 into two steps (median and rounding) so step 5 can use the ETH/INDEX median price before rounding

mhairimca (commented):
nothing to add

UMIPs/umip-64.md
GottliebFreudenreich: There is no particular reason. I would also prefer to use DAI.

UMIPs/umip-64.md
GottliebFreudenreich: I added MedianizerPriceFeed.js for implementation

UMIPs/umip-64.md
GottliebFreudenreich: Thanks for mentioning it. I'll change this.

UMIPs/umip-61.md
GottliebFreudenreich: That's a great idea which provides more flexebility and also helps readability.

UMIPs/umip-64.md
GottliebFreudenreich: That's correct! I changed the ordering.

UMIPs/umip-64.md
```diff
@@ -0,0 +1,381 @@
+## Headers
+| UMIP-64   |   |
+|------------|---|
+| UMIP Title | Add INDEXETH, INDEXUSD, DPIETH and DPIUSD as price identifier |
```
smb2796, line 4: Can you add the other price identifiers here as well?

UMIPs/umip-64.md
```diff
@@ -0,0 +1,381 @@
...
+    6. (for ETH/DPI) Take the Inverse of the result of step 4 (1/ DPI/ETH) and round to 18 decimals to get the ETH/DPI price.
+
+**For DPI/USD and USD/DPI** 
+
+    1. Query DPI/ETH Price from Uniswap using 1 minute TWAP (0x4d5ef58aac27d99935e5b6b4a6778ff292059991).
+    2. Query DPI/ETH Price from Sushiswap using 1 minute TWAP (0x34b13f8cd184f55d0bd4dd1fe6c07d46f245c7ed).
+    3. Query DPI/ETH Price from Balancer using 1 minute TWAP (0x2aa3041fe813cfe572969216c6843c33f14f9194).
+    4. Query the ETH/USD Price as per UMIP-6.
+    5. Multiply the DPI/ETH prices in steps 1, 2 and 3 by the ETH/USD price to get the respective DPI/USD prices.
+    6. Take the median of prices acquired from step 5
```
smb2796, line 367: It seems like it would make more sense to take the median and then do the ETH/USD conversion. The results would always be the same, but it just seems cleaner to me.

UMIPs/umip-64.md
```diff
@@ -0,0 +1,381 @@
...
+# IMPLEMENTATION
+
+**For INDEX/ETH and ETH/INDEX**
+
+    1. Query INDEX/ETH Price from Uniswap using 1 minute TWAP (0x3452a7f30a712e415a0674c0341d44ee9d9786f9).
+    2. Query INDEX/ETH Price from Sushiswap using 1 minute TWAP (0xa73df646512c82550c2b3c0324c4eedee53b400c).
+    3. Query INDEX/ETH Price from Balancer using 1 minute TWAP (0xcf19a7c81fcf0e01c927f28a2b551405e58c77e5).
+    4. Take the median of prices acquired from steps 1, 2, and 3
+    5. Take the result from step 4 and round to 18 decimals to get the final INDEX/ETH price.
+    6. (for ETH/INDEX) Take the Inverse of the result of step 4 (1/ INDEX/ETH) and round to 18 decimals to get the ETH/INDEX price.
```
smb2796, line 338: I'd suggest rounding to less than 18 decimals. 18 decimals seems like an unnecessary level of precision

UMIPs/umip-64.md
```diff
@@ -0,0 +1,381 @@
...
+The USD/ETH and ETH/USD Price can be calculated as per UMIP-6.
+
+
+# IMPLEMENTATION
+
+**For INDEX/ETH and ETH/INDEX**
+
+    1. Query INDEX/ETH Price from Uniswap using 1 minute TWAP (0x3452a7f30a712e415a0674c0341d44ee9d9786f9).
+    2. Query INDEX/ETH Price from Sushiswap using 1 minute TWAP (0xa73df646512c82550c2b3c0324c4eedee53b400c).
+    3. Query INDEX/ETH Price from Balancer using 1 minute TWAP (0xcf19a7c81fcf0e01c927f28a2b551405e58c77e5).
```
smb2796, line 335: Should there be additional steps for this pool, given that this pool is composed of 4 assets?

UMIPs/umip-64.md
```diff
@@ -0,0 +1,381 @@
...
+
+    - No limit currently
+
+11.  What would be the cost of sending 15,000 queries?
+
+     - $0
+
+# PRICE FEED IMPLEMENTATION
+
+These price identifiers use the [UniswapPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js), [BalancerPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js), [ExpressionPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js) and [MedianizerPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js).
```
smb2796, line 113: I don't think the BalancerPriceFeed will work automatically for a 4-asset pool. @chrismaree what do you think?

UMIPs/umip-64.md
```diff
@@ -0,0 +1,381 @@
...
+- ETH/DPI on Sushiswap
+- ETH/USD per UMIP 6
+
+3. Provide recommended endpoints to query for real-time prices from each market listed. 
+
+    - Uniswap DPI/ETH Pool Address: 0x4d5ef58aac27d99935e5b6b4a6778ff292059991
+    - Uniswap INDEX/ETH Pool Address: 0x3452a7f30a712e415a0674c0341d44ee9d9786f9
+    - Sushiswap DPI/ETH Pool Address: 0x34b13f8cd184f55d0bd4dd1fe6c07d46f245c7ed
+    - Sushiswap INDEX/ETH Pool Address: 0xa73df646512c82550c2b3c0324c4eedee53b400c
+    - Balancer INDEX/ETH Pool Address: 0xcf19a7c81fcf0e01c927f28a2b551405e58c77e5
```
smb2796, line 66: This balancer pool also is weighted at 70/30 INDEX/WETH. This should be noted and probably accounted for within the implementation section.

UMIPs/umip-64.md
chrismaree: yup. it currently does not support this.

UMIPs/umip-64.md
GottliebFreudenreich: Yes, I'll add the inverse pairs too

UMIPs/umip-64.md
GottliebFreudenreich: I'm sorry I just wanted to go with best practices I identified from already approved collateral. What is your suggestion for the number of decimals?

UMIPs/umip-64.md
GottliebFreudenreich: Alright! I'll change the conversion order.

UMIPs/umip-64.md
GottliebFreudenreich: Thanks! I'll note this here and in the implementation section.

UMIPs/umip-64.md
GottliebFreudenreich: Thank you for digging into it! I would very much like to only query DEXes instead of implementing a CEX price feed from e.g. Huobi. What are your thoughts on testing the BalancerPriceFeed with this implementation? Would you rather see the BalancerPriceFeed removed?

UMIPs/umip-64.md
GottliebFreudenreich: Sorry, I'm not really sure what to implement.. Do you have any suggestions or best practices?

UMIPs/umip-64.md
smb2796: Usually 5-8 decimals of rounding are specified (admittedly with a few exceptions!). For price identifiers that are intended to be collateralized in ETH, we usually use 5. So it depends on the value of the collateral that you are planning to use, but I think 5 would be sufficient given that INDEX and DPI are both worth considerably less per than ETH.

UMIPs/umip-64.md
smb2796: Agreed it's not binding, but also as a meta point, this field is slightly irrelevant now and should probably be removed from the price identifier template!

UMIPs/umip-64.md
smb2796: Thank you!

UMIPs/umip-64.md
```diff
@@ -0,0 +1,381 @@
...
+    6. (for ETH/INDEX) Take the Inverse of the result of step 4 (1/ INDEX/ETH) and round to 18 decimals to get the ETH/INDEX price.
+
+**For INDEX/USD and USD/INDEX** 
+
+    1. Query INDEX/ETH Price from Uniswap using 1 minute TWAP (0x3452a7f30a712e415a0674c0341d44ee9d9786f9).
+    2. Query INDEX/ETH Price from Sushiswap using 1 minute TWAP (0xa73df646512c82550c2b3c0324c4eedee53b400c).
+    3. Query INDEX/ETH Price from Balancer using 1 minute TWAP (0xcf19a7c81fcf0e01c927f28a2b551405e58c77e5 - this is a 70/30 INDEX/ETH pool).
+    4. Query the ETH/USD Price as per UMIP-6.
+    5. Take the median of prices acquired from step 1, 2, and 3.
+    6. Multiply the INDEX/ETH price from step 5 by the ETH/USD price from step 4 to get the respective INDEX/USD prices.
```
smb2796, line 347: Small wording thing, but it wouldn't be the "respective INDEX/USD prices" anymore after changing the ordering. It would just be the "median INDEX/USD price".

UMIPs/umip-64.md
smb2796: Cool, thanks!

UMIPs/umip-64.md
GottliebFreudenreich: Alright! I'll set rounding to 5.

UMIPs/umip-64.md
GottliebFreudenreich: Thanks! I'll fix this.

UMIPs/umip-64.md
smb2796: I think the BalancerPriceFeed can be left in. We will just need to test before launching any contracts using this

smb2796 (approved):
Looks like all issues except a couple of questions around the BalancerPriceFeed were addressed. Going to merge in while we test the price feed.

PR #209, opened 2021-03-10
Propose v2 PIDs for all deprecated identifiers

smb2796:


UMIPs/v2-price-identifier.md
```diff
@@ -0,0 +1,35 @@
+## Headers
+| UMIP-xx    |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add USDBTC_V2, BCHNBTC_V2, ELASTIC_STABLESPREAD/USDC_V2, STABLESPREAD/USDC_V2, STABLESPREAD/BTC_V2, AMPLUSD_V2 as supported price identifiers            |
```
smb2796, line 4: Had trouble thinking up good names for these. Open to any ideas!

UMIPs/v2-price-identifier.md
```diff
@@ -0,0 +1,35 @@
...
+## Technical Specifications
+
+Voters should refer to the rationale, technical specifications, implementations and security considerations in each corresponding UMIP. The only change to the implemenentations of these price identifiers are that these should all be scaled 10^18. Voters currently do not have to adjust results for any sort of scaling when voting through the UMA voter dapp, so the result that a voter would input for one of these price ids would be exactly the same as for the deprecated identifiers.
+
+The price identifier pairings are as follows.
+
+| New Identifier | Old Identifier | UMIP |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------| --------------|
+| USDBTC_V2 | USDBTC | [UMIP-7](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-7.md) |
+| BCHNBTC_V2 | BCHNBTC | [UMIP-23](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-23.md) |
```
mhairimca, line 30: There are quite a few dependencies on UMIP 7 (USDBTC), in contrast to the rest of the PIs in this UMIP.

Would it perhaps be better to update  UMIP 7 rather than including it in this one.

smb2796: We're not able to update UMIP-7 because there is a live contract (YD-BTC) that depends on the existing USDBTC price identifier

UMIPs/v2-price-identifier.md
chicfilabae: Eh, I mean it's fine with us

UMIPs/v2-price-identifier.md
mhairimca: Maybe a separate one for BTC then, and when that contract expires update UMIP 7 to point to that one.

smb2796: What would be the reason for doing that? These all are being used by contracts currently - I see no difference between USDBTC and the other identifiers being proposed

mhairimca: So that if future PIs point to a BTC pricefeed as part of the calculation, there is no danger of confusion with any of the other calculations for other tokens.

smb2796:
@mhairimca what do you think? Good to sign off/approve and merge?

Tommy1231232 (approved):

PR #216, opened 2021-03-23
UMIP for project specific TVL price identifiers

bryanjcampbell1:
Signed-off-by: bryanjcampbell1 <bryanjcampbell1@gmail.com>

### DIFF
--- a/UMIPs/umip-70.md
+++ b/UMIPs/umip-70.md
@@ -1,0 +1,86 @@
+## Headers
+ - UMIP 70
+ - Title: Add AAVE_TVL, COMPOUND_TVL, SUSHISWAP_TVL, UNISWAP_TVL, BADGER_TVL, DODO_TVL, BZX_TVL, OPIUM_TVL, and PIEDAO_TVL as supported price identifiers
+ - Author:  Bryan Campbell (bryanjcampbell1@gmail.com)
+ - Status: DRAFT
+ - Created: March, 22, 2021
+
+
+## Summary
+The DVM should support price requests for the following Total Value Locked (TVL) derivatives:
+
+1) Aave TVL divided by 1,000,000,000
+2) Compound TVL divided by 1,000,000,000
+3) SushiSwap TVL divided by 1,000,000,000
+4) Uniswap TVL divided 1,000,000,000
+5) Badger DAO TVL divided 1,000,000,000
+6) Dodo TVL divided 1,000,000,000
+7) bZx TVL divided 1,000,000,000
+8) Tornado Cash TVL divided 1,000,000,000
+9) Opium Network TVL divided 1,000,000,000
+10) PieDAO TVL divided 1,000,000,000
+
+
+## Motivation
+
+The DVM currently does not support the previously mentioned price indices.
+Adding project specific TVL identifiers enables the creation of synthetic assets with a price that tracks the actual use of a DeFi protocol.  
+
+
+## Technical Specification
+
+### PLACEHOLDER_TVL
+- Quote Currency: USD
+- Source: https://data.defipulse.com/
+- Result Processing: None
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad consensus.
+- Rounding: Round to nearest 3 decimal places (fourth decimal place digit >= 5 rounds up and < 5 rounds down)
+- Intended Collateral Type: USDC
+- Scaling Decimals: 18 (1e18)
+- Pricing Interval: 60 minutes
+- Dispute timestamp rounding: down 
+
+
+## Rationale
+
+Price of DeFi Pulse API is free at Trial level and 95 USD for the Starter level.
+It would require 200 GET requests per month in order to use up the credits offered at the Trial level.
+At the time of writing, DeFi Pulse’s metrics are the most popular for measuring a project’s TVL. 
+
+
+## Implementation
+
+The following procedure uses DeFi Pulse Data API and can be utilized to determine the value of any of the price identifiers listed in the title of this UMIP.  To avoid unnecessary repition, "PLACEHOLDER_TVL" and "placeholder" are stand ins for the price id and name of the abovementioned projects. 
+
+API docs can be found at https://docs.defipulse.com/api-docs-by-provider/defi-pulse-data/total-value-locked/total-value-locked
+
+
+1) Sign up for the free Trial account from https://data.defipulse.com/
+2) Call the API with https://data-api.defipulse.com/api/v1/defipulse/api/GetHistory?project=placeholderp&period=1w&api-key=********
+
+The response object is an array of TVL values at hourly increments over the past week.
+
+3) Find the object in the response array that corresponds to timestamp equal to time the synthetic token expired
+4) The value at the key “tvlUSD” is the TVL that we want
+5) Divide the value found at step 4 by 1,000,000,000 to get the settlement value of the PLACEHOLDER_TVL price identifier
+
+
+## ## Backup Calculation
+
+In the case that the the value from DeFi Pulse for any project at expiration is determined to differ from broad market consensus, the following are examples of how voters could arrive at a backup measurement.
+
+1) Use Historical TVL values from the defillama API. Docs can be found at https://0xngmi.gitbook.io/defillama/api
+2) Use TVL values from individual project's subgraphs
+
+
+## Security Considerations
+
+There are concerns around using DeFi Pulse data as the primary source for determining price feeds. 
+
+1) DeFi Pulse alone is not decentralized enough
+2) DeFi Pulse is not fully transparent in how they calculate TVL
+3) DeFi Pulse is fairly arbitrarily choosing what platforms to include in their calculation      
+
+If DeFi Pulse suspends their api service a new price feed source will have to be determined.  DeFi Llama offers a free and completely open API with historical values for TVL.  This is an excellent secondary option.
+
+There are likely some security considerations around using a price feed that is available with an hourly granularity. Contract creators should likely specify longer liquidation and withdrawal liveness times to allow for prices to be accurately reported. There could be large jumps in price resulting from the 1 hour lag that could result in unexpected liquidations for token sponsors.


UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+### PLACEHOLDER_TVL
+- Quote Currency: USD
+- Source: https://data.defipulse.com/
+- Result Processing: None
+- Input Processing: None. Human intervention in extreme circumstances where the result differs from broad consensus.
+- Rounding: Round to nearest 3 decimal places (fourth decimal place digit >= 5 rounds up and < 5 rounds down)
+- Intended Collateral Type: USDC
+- Scaling Decimals: 18 (1e18)
+- Pricing Interval: 60 minutes
+- Dispute timestamp rounding: down
```
Reinis-FRP, line 41: This belongs to `IMPLEMENTATION` section

UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+- Intended Collateral Type: USDC
+- Scaling Decimals: 18 (1e18)
+- Pricing Interval: 60 minutes
+- Dispute timestamp rounding: down 
+
+
+## Rationale
+
+Price of DeFi Pulse API is free at Trial level and 95 USD for the Starter level.
+It would require 200 GET requests per month in order to use up the credits offered at the Trial level.
```
Reinis-FRP, line 47: This belongs to `MARKETS & DATA SOURCES` section

UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+At the time of writing, DeFi Pulse’s metrics are the most popular for measuring a project’s TVL.
```
Reinis-FRP, line 48: Please elaborate on the design motivation and any alternatives considered

UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+## Rationale
+
+Price of DeFi Pulse API is free at Trial level and 95 USD for the Starter level.
+It would require 200 GET requests per month in order to use up the credits offered at the Trial level.
+At the time of writing, DeFi Pulse’s metrics are the most popular for measuring a project’s TVL. 
+
+
+## Implementation
+
+The following procedure uses DeFi Pulse Data API and can be utilized to determine the value of any of the price identifiers listed in the title of this UMIP.  To avoid unnecessary repition, "PLACEHOLDER_TVL" and "placeholder" are stand ins for the price id and name of the abovementioned projects.
```
Reinis-FRP, line 53: repition -> repetition

Reinis-FRP, line 53: Would it be possible to include here a translation table from Project specific PLACEHOLDER_TVL to slug used in defipulse.com? Or you expect that slugs might change and voters always should look it up on defipulse.com?

UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+
+API docs can be found at https://docs.defipulse.com/api-docs-by-provider/defi-pulse-data/total-value-locked/total-value-locked
+
+
+1) Sign up for the free Trial account from https://data.defipulse.com/
+2) Call the API with https://data-api.defipulse.com/api/v1/defipulse/api/GetHistory?project=placeholderp&period=1w&api-key=********
+
+The response object is an array of TVL values at hourly increments over the past week.
+
+3) Find the object in the response array that corresponds to timestamp equal to time the synthetic token expired
```
Reinis-FRP, line 63: Is it possible to narrow down the constructed API request so that it returns only the relevant timestamp TVL?

UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+4) The value at the key “tvlUSD” is the TVL that we want
+5) Divide the value found at step 4 by 1,000,000,000 to get the settlement value of the PLACEHOLDER_TVL price identifier
+
+
+## ## Backup Calculation
+
+In the case that the the value from DeFi Pulse for any project at expiration is determined to differ from broad market consensus, the following are examples of how voters could arrive at a backup measurement.
+
+1) Use Historical TVL values from the defillama API. Docs can be found at https://0xngmi.gitbook.io/defillama/api
```
pemulis, line 72: Instead of this being merely a back-up, should the canonical price feed take the average of DeFi Pulse and defillama (or the median of the two plus a third, if a good third source exists)?

pemulis (commented):
I think this is a really cool metric! I agree with @Reinis-FRP's excellent comments, and added a note that perhaps we should take multiple data sources into account for this price feed, rather than only a single data source (DeFi Pulse) with additional sources as mere backups.

akhanubis:
How important it is for the price IDs to be accurately tracking the real TVL instead of tracking what DefiPulse says it's the TVL? I have used DefiPulse's API in the past and have seen some issues in their data (which is expected considering that calculating TVL is not an easy task with protocols adding contracts, features, pools, etc over time).

To give an example of a wrong value being reported as of now:
DP reports that Compound TVL is 5.866b at 2021-03-26T22 UTC when in reality it is  6.916b (according to Compound itself). This is because DP isnt taking into account the new WBTC2 market that opened in Compound on March 18th, so more than a week ago and still reporting wrong data.
Considering that Compound is going to migrate all the markets in the near future, we can expect these type of discrepancies to come up regularly.

Do the price IDs aim for accuracy or ease of implementation? I think both answers are correct and it's a matter of perspective, but I want to point out the inherent risks of trusting one API that hasnt been designed with complete accuracy in mind.

akhanubis:
Is the idea of the price IDs to measure project usage in Ethereum or overall? Sushiswap, Badger and Dodo have already expanded to other chains.
I assume Ethereum only, but maybe that can be made explicit somewhere in the definition of the UMIP in case DefiPulse add other chains in the future.

UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+
+There are concerns around using DeFi Pulse data as the primary source for determining price feeds. 
+
+1) DeFi Pulse alone is not decentralized enough
+2) DeFi Pulse is not fully transparent in how they calculate TVL
+3) DeFi Pulse is fairly arbitrarily choosing what platforms to include in their calculation      
+
+If DeFi Pulse suspends their api service a new price feed source will have to be determined.  DeFi Llama offers a free and completely open API with historical values for TVL.  This is an excellent secondary option.
+
+There are likely some security considerations around using a price feed that is available with an hourly granularity. Contract creators should likely specify longer liquidation and withdrawal liveness times to allow for prices to be accurately reported. There could be large jumps in price resulting from the 1 hour lag that could result in unexpected liquidations for token sponsors.
```
akhanubis, line 86: I think the bigger issue here is that the TVL of a protocol can easily jump >+50% between two datapoints, and not because the protocol's real TVL increased by 50% in < 1 hour, but because DefiPulse (or any other source) suddenly added a new address to the list of watched addresses of the protocol.
For example, DP is now reporting $270m for yearn finance (not part of the submitted price ids, but the same can happen with any other protocol) when in reality it should be reporting $1.6b. What would happen to sponsors sitting at a comfortable CR once DP decides to update and suddenly the TVL jumps +492%  in an hour?

UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Motivation
+
+The DVM currently does not support the previously mentioned price indices.
+Adding project specific TVL identifiers enables the creation of synthetic assets with a price that tracks the actual use of a DeFi protocol.  
+
+
+## Technical Specification
+
+### PLACEHOLDER_TVL
```
smb2796, line 32: Looks like the other price identifiers need to be added as well

UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Implementation
+
+The following procedure uses DeFi Pulse Data API and can be utilized to determine the value of any of the price identifiers listed in the title of this UMIP.  To avoid unnecessary repition, "PLACEHOLDER_TVL" and "placeholder" are stand ins for the price id and name of the abovementioned projects. 
+
+API docs can be found at https://docs.defipulse.com/api-docs-by-provider/defi-pulse-data/total-value-locked/total-value-locked
+
+
+1) Sign up for the free Trial account from https://data.defipulse.com/
+2) Call the API with https://data-api.defipulse.com/api/v1/defipulse/api/GetHistory?project=placeholderp&period=1w&api-key=********
```
smb2796, line 59: Can you provide a list of the entries that could be used for `placeholder`?

UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+8) Tornado Cash TVL divided 1,000,000,000
+9) Opium Network TVL divided 1,000,000,000
+10) PieDAO TVL divided 1,000,000,000
+
+
+## Motivation
+
+The DVM currently does not support the previously mentioned price indices.
+Adding project specific TVL identifiers enables the creation of synthetic assets with a price that tracks the actual use of a DeFi protocol.  
+
```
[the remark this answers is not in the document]
bryanjcampbell1: I'm confused by this comment. Would an example of an interatction be a tx hash?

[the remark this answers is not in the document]
mhairimca: more a comment about why someone might want to build a contract using this PI and how they would use it.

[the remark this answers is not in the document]
mhairimca: more a comment about why someone might want to build a contract using this PI and how they would use it.

UMIPs/umip-70.md
```diff
@@ -0,0 +1,86 @@
...
+2) Compound TVL divided by 1,000,000,000
+3) SushiSwap TVL divided by 1,000,000,000
+4) Uniswap TVL divided 1,000,000,000
+5) Badger DAO TVL divided 1,000,000,000
+6) Dodo TVL divided 1,000,000,000
+7) bZx TVL divided 1,000,000,000
+8) Tornado Cash TVL divided 1,000,000,000
+9) Opium Network TVL divided 1,000,000,000
+10) PieDAO TVL divided 1,000,000,000
+
```
mhairimca, line 22: standard notation would make this clearer, but why the processing at this stage rather than results processing?

UMIPs/umip-70.md
bryanjcampbell1: The problem is that DeFi Pulse is the only provider I have found that gives historical values with at hourly increments.  DeFi Llama is awesome but their historical values are daily.

bryanjcampbell1:
More research needs to be done on TVL data sources

akhanubis:
@bryanjcampbell1 would you be open to doing tvl calculation reading onchain data?
Some of the protocols you want to support may be hard to cover, but Uniswap, Sushiswap, Compound should be doable. They have some kind of onchain registry of all the pools/markets (Uniswap factory, Compound list of markets). Compound even has an official API with all the info.

UMIPs/umip-70.md
Hahz48: above mentioned

Hahz48: above mentioned

PR #222, opened 2021-03-30
Add BALUSD and USDBAL as supported price identifiers 

mhairimca:


UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
```diff
@@ -0,0 +1,130 @@
...
+|------------|------------------------------------------------------------------------------------------------------------------------------------------| 
+| UMIP Title | Add BALUSD and USDBAL as supported price identifiers              |
+| Authors    | Mhairi McAlpine (mhairi@UMAproject.org)  |
+| Status     | Draft                                                                                                                     
+| Created    | March 30, 2021   
+| [Discourse Link](https://discourse.umaproject.org/t/add-balusd-usdbal-as-supported-price-identifiers/512) |  
+
+## SUMMARY
+
+The DVM should support price requests for BAL/USD, USD/BAL.
```
pemulis, line 12: I think we should add this to the boilerplate: "The canonical price identifiers on-chain should be `BALUSD` and `USDBAL`."

We can go through and do this for other UMIPs, to avoid confusion. It's tripped us up in proposals before. The strings in the UMIP title are the canonical identifiers, but the format switches up in the summaries for readability.

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
```diff
@@ -0,0 +1,130 @@
...
+
+## MOTIVATION
+
+The DVM currently does not support these price identifiers. BAL is also being proposed as a supported collateral types and using BAL as collateral together with these price identifiers would allow BAL holders to get leverage on their holdings by minting yield-dollar type stablecoins. Adding BAL/USD will also allow for the creation of synthetic BAL, or more bespoke contracts such as covered call options backed by BAL.
```
pemulis, line 16: `collateral type`

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
```diff
@@ -0,0 +1,130 @@
...
+Coinbase Pro: https://api.cryptowat.ch/markets/coinbase-pro/balusd/price
+
+- Update time: Every second for CW. Every block for Balancer. 
+- Historical Price Endpoints:
+
+Binance: https://api.cryptowat.ch/markets/binance/balusdt/ohlc?after=1612880040&before=1612880040&periods=60
+
+Coinbase Pro: https://api.cryptowat.ch/markets/coinbase-pro/balusd/ohlc?after=1612880040&before=1612880040&periods=60
+
+Balancer : ?
```
pemulis, line 44: We should figure out what goes here.

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
```diff
@@ -0,0 +1,130 @@
...
+- Base Currency: USD
+- Quote currency: BAL
+- Scaling Decimals: 18 (1e18)
+- Rounding: Round to nearest 6 decimal places (seventh decimal place digit >= 5 rounds up and < 5 rounds down)
+
+
+## RATIONALE
+
+The markets chosen for pricing BAL/USD were selected on the following basis 
+    - Bal/Eth on Balancer as it is the highest volume maket
```
pemulis, line 107: `market`

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
```diff
@@ -0,0 +1,130 @@
...
+    - Bal/USDT on Binance as it is the highest volume  because they are three of the highest daily trading volume markets available
```
pemulis, line 108: `as it is one of the three highest daily trading volume markets available`

Just guessing here, not sure what this is trying to say.

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
```diff
@@ -0,0 +1,131 @@
...
+- Is an API key required to query these sources? CW has a free tier, but requires an API key beyond that.
+- Is there a cost associated with usage? Yes. Cryptowatch requires a purchase of credits beyond their free tier.
+- If there is a free tier available, how many queries does it allow for?
+
+The free tier is limited to 10 API credits per 24-hours; the cost of querying the market price of a given exchange is 0.005 API credits (i.e. querying the two exchanges will cost 0.010 API credits). This would allow for 1000 free queries per day.
+
+- What would be the cost of sending 15,000 queries? $5.
+
+## PRICE FEED IMPLEMENTATION
+These price identifiers will use price feeds that already exist. Both will use the [Cryptowatch](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js), [Uniswap](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js) and [Expression](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js) price feeds to get the price of BALUSD.
```
Reinis-FRP, line 58: Shouldn't Uniswap be replaced with [Balancer](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js) price feed implementation?

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
```diff
@@ -0,0 +1,131 @@
...
+## IMPLEMENTATION
+
+### BALUSD and USDBAL
+
+1. For the price request timestamp, query for the BALUSD(T) prices on Binance and Coinbase Pro and and the ETHUSD price by following the guidelines of UMIP-6. The open price of the 60-second OHLC period that this price request timestamp falls in should be used.
+2. For the block of the price request timestamp, query for the BALETH price from Balancer.
+3. Multiply the gathered ETHUSD price by BALETH to get the Balancer BALUSD price.
+4. Take the median of these.
+5. Round to 6 decimals to get the BALUSD price.
+6. To get the USDBAL price, voters should just take the inverse of BALUSD from step 5.
```
Reinis-FRP, line 124: USDBAL should also be rounded.
In order to avoid double rounding, I would suggest taking inverse of BALUSD median from step 4

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
akhanubis: There are several of BAL/ETH pools, the one with the most liquidity (and by a huge margin) is 0x59a19d8c652fa0284f44113d0ff9aba70bd46fb4.
I'd leave the UMIP open with something like "the most liquid pool" in case liquidity migrates with Balancer V2, or use a weighted average between the pools.

smb2796: I think this can fall under the category of "differing from broad market consensus" situations where it would be best for UMA voters to choose an alternative methodology. So I think the specific pool should be listed, but this should be noted in implementation along with the more boilerplate broad market consensus language.

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
```diff
@@ -0,0 +1,138 @@
...
+### BALUSD and USDBAL
+
+1. For the price request timestamp, query for the BALUSD(T) prices on Binance and Coinbase Pro and and the ETHUSD price by following the guidelines of UMIP-6. The open price of the 60-second OHLC period that this price request timestamp falls in should be used.
+2. For the block of the price request timestamp, query for the BALETH price from Balancer.
+3. Multiply the gathered ETHUSD price by BALETH to get the Balancer BALUSD price.
+4. Take the median of these.
+5. Round to 6 decimals to get the BALUSD price.
+6. To get the USDBAL price, voters should just take the inverse of the result of Step 4 (unrounded BALUSD price) then round to 6 decimal places.
+
+
```
smb2796, line 133: Can you add the language around "broad market consensus" and relate part of it back to @akhanubis comments around choosing the most valid balancer pool?

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
smb2796: I like this. Another option is that people just write this out fully instead of using the "BAL/USD" format.

"The DVM should support requests for the price of 1 BAL token in USD, or the price of 1 USD in BAL."

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
```diff
@@ -0,0 +1,138 @@
...
+These price identifiers will use price feeds that already exist. Both will use the [Cryptowatch](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js), [Balancer](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js) and [Expression](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js) price feeds to get the price of BALUSD. 
+
+The default price feed config for BALUSD would follow this pattern:
+
+```
+BALUSD: {
+    type: "expression",
+    expression: `
+        SPOT_BALANCER = SPOT_BALANCER_ETH * ETHUSD;
+        median(SPOT_BINANCE, SPOT_COINBASE-PRO, SPOT_BALANCER)
```
chrismaree, line 71: this is a great feed. I really like the use of the expression feed like this to cross over between ETH and BALETH

UMIPs/Add_BALUSD and USDBAL_as_supported_price_identifiers.md
```diff
@@ -0,0 +1,138 @@
...
+    `,
+    lookback: 7200,
+    minTimeBetweenUpdates: 60,
+    customFeeds: {
+      SPOT_BINANCE: { type: "cryptowatch", exchange: "binance", pair: "balusdt" },
+      SPOT_COINBASE-PRO: { type: "cryptowatch", exchange: "coinbase-pro", pair: "balusd" },
+      SPOT_BALANCER: { type: "balancer", address: "0x59a19d8c652fa0284f44113d0ff9aba70bd46fb4" },
```
chrismaree, line 78: has this been tested? I'm pretty sure the balance feed also needs `tokenIn` and `tokenOut` as params.

smb2796: Yep I agree with this. Believe SPOT_BALANCER should look something like:

```
{ 
type: "balancer", 
balancerAddress: "0x59a19d8c652fa0284f44113d0ff9aba70bd46fb4", 
balancerTokenIn: "0xba100000625a3754423978a60c9317c58a424e3D",
balancerTokenOut: "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
}
```

smb2796 (approved):
A few outstanding comments, but I can just add in a few of the comments from my side!

PR #228, opened 2021-04-01
Add ethVIX and iethVIX Price Identifiers

ccashwell:
Adds Price Identifiers for two [dVIX](https://dvix.io) Indexes: ethVIX (long ETH volatility) and iethVIX (short ETH volatility).


UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
+# Headers
+
+| UMIP-72 | |
```
pemulis, line 3: This will get updated based on actual order it passes.

UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
...
+
+## Step 1 - Use historical price feed.
+
+- Users can get 7 days of historical price feeds at https://dvix.io/api/historicalData?asset=ETH. Users need to come to consensus and use the last trusted price.
+
+- If a historical price feed cannot become agreed upon (e.g. it was too far in the past) or the end point api is not trusted then move onto Step 2.
+
+## Step 2 - Calculate ethVIX & iethVIX from raw data.
+
+Note: If users feel that the real-time data is compromised (e.g. a market-making cartel attack on Deribit), this calculation should be done once the data is considered safe again.
```
pemulis, line 161: Pretty involved calculation, but I like the existence of a backup option.

UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
...
+
+1.  Download the [methodology paper](https://dvix.io/ethVIX_methodology.pdf)
+1.  Download the [verification spreadsheet](https://dvix.io/ethVIX_verification.xlsx)
+1. Access the real-time data at https://dvix.io/rawData.
+
+    - Note: Make sure to turn Auto-Refresh off as you will want the snapshot you’re reviewing to be persisted in the interface.
+
+    - Alternatively a user can take a screenshot of the Near and Next Options Expiry on Deribit and get the current lending rate for ETH from [Aave](https://aave.com).
```
pemulis, line 169: Could you elaborate on how Aave's current lending rate would be used here?

UMIPs/umip-72.md
ccashwell: It is indeed fairly complex to do by hand, but we felt that it was important to provide an independent mechanism for validating the data in the interest of transparency and resiliency.

UMIPs/umip-72.md
ccashwell: The Cboe VIX methodology relies on a "risk-free interest rate," which in TradFi is represented by treasury yields.

From the Cboe VIX Methodology paper:
> The risk-free interest rates, R1 and R2, are yields based on U.S. Treasury yield curve rates (commonly
referred to as “Constant Maturity Treasury” rates), to which a cubic spline is applied to derive yields on
the expiration dates of relevant SPX options. As such, the VIX Index value calculation may use different
risk-free interest rates for near- and next-term options. 

We believe that using U.S. Treasury yields for ethVIX is unreasonable for a few reasons, including:

1. Treasury yields are not reflective of typical lending protocol returns for blockchain-based assets.
2. Treasuries are not accessible within the blockchain environment.
3. Treasuries are highly centralized, government-issued instruments which we believe is not appropriate for inclusion within a decentralized protocol.

UMIPs/umip-72.md
ccashwell: Thanks, the instructions were not entirely clear on this so I just went with the next serial value.

UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
...
+| Status | Draft |
+| Created | 03.31.2021 |
+| Link to Discourse | TBD |
+
+
+# Summary
+
+ethVIX is a real-time index representing the expected volatility for ETH over the coming 14 days. iethVIX is the inverse of the ethVIX. When ethVIX moves up, iethVIX moves down by an equivalent percentage, and vice versa.
+
+# Motivation
```
Reinis-FRP, line 16: Could you please add reference for further reading on the subject, e.g. [dVIX](https://dvix.io/) docs

UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
...
+
+    - dVIX Protocol’s ethVIX & iethVIX price feeds
+    - dVIX Protocol’s public website
+    - dVIX Protocol’s IPFS log
+    - Deribit ETH Options data
+
+1. Which specific pairs should be queried from each market?
+
+    - The dVIX API aggregates all options data that is needed for the calculation.
+    - dVIX Protocol’s IPFS log provides a backup copy of this data.
```
Reinis-FRP, line 39: Link?

UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
...
+      - Near Term - The expiry 5-13 days from current time.
+      - Next Term- The expiry 15-23 days from current time.
+
+1. Provide recommended endpoints to query for real-time prices from each market listed.
+
+    These should match the data sources used in "Price Feed Implementation".
+
+    - https://dvix.io/api/latest?asset=ETH
+
+      Note: This endpoint provides a JSON response including the latest ethVIX (field: vix) and iethVIX (field: iVix) price along with the full options chain and other data points used in the calculation.
```
Reinis-FRP, line 50: I do not see `iVix` field in the JSON response. Do you mean `vixInverse` field? is there any processing required on this?

UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
...
+Please provide a link to your price feed pull request.
+https://github.com/UMAprotocol/protocol/pull/2792
+
+# TECHNICAL SPECIFICATIONS
+
+## ethVIX
+
+1. Price Identifier Name - ethVIX
+2. Base Currency - USDC
+3. Quote currency - USDC
```
Reinis-FRP, line 103: why quote currency is the same as base currency?

UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
...
+4. Intended Collateral Currency - USDC
+5. Collateral Decimals - 6 Decimals
+6. Rounding - ethVIX and iethVIX price feeds are rounded to 2 decimal places.
+
+## iethVIX
+
+1. Price Identifier Name - iethVIX
+2. Base Currency - USDC
+3. Quote currency - USDC
```
Reinis-FRP, line 112: why quote currency is the same as base currency?

UMIPs/umip-72.md
ccashwell: Yeah you're right, in the expanded detail endpoint `/api/latest` it's `vixInverse` and it's `iVix` in the historical. I will update the `latest` endpoint to match for posterity.

UMIPs/umip-72.md
ccashwell: It's not `USDC_USDC`, it's actually more like `Convexity_USDC` but of course convexity is a concept, not a currency. We put `USDC` here because the quote price is actually the USDC-denominated price of the base which is not technically a currency at all. Open to suggestions here for sure.

UMIPs/umip-72.md
ccashwell: Same answer as above -- it's not a real currency.

UMIPs/umip-72.md
ccashwell: The Methodology paper is linked already and goes into pretty deep detail, do you think we should duplicate that content here?

UMIPs/umip-72.md
ccashwell: Just to add context on my comment, technically there is no "Quote Currency". ethVIX and iethVIX are an expression of 14-day future volatility expressed as a percentage. Like the Cboe VIX, we multiply that by 100 to change that percentage to something that can be expressed as a dollar (or USDC) amount. Thus, there is no underlying to the ethVIX or iethVIX it is simply a price feed that can, for example, be offered in the form of synthetic derivatives that can be traded.

UMIPs/umip-72.md
ccashwell: We will provide a pinned log for this prior to the first round of synths, but it's not available yet. Shall I remove the IPFS ref and do a future PR to link it once that's done?

UMIPs/umip-72.md
mhairimca: Can that be inserted in as a formula to the quote currency?

UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
...
+# Security considerations
+
+Adding these new price identifiers pose certain risks:
+
+- Currently the options chain data used is from a centralized source (i.e. Deribit).
+- Manipulation can occur on Deribit’s ETH options markets. As outlined previously this is mitigated in two ways:
+  - It will be expensive if the attacker’s orders are filled.
+  - In cases where prices are greatly misaligned with expected market prices the price feeds use the mark price instead of the bid/ask.
+- To keep a 14-day average of volatility, the calculation rolls from one options expiry to another. This can lead to price swings but may also be a vector of attack.
+  - To mitigate this, we are launching weekly-expiring contracts after the next term rollover and having them expire before it occurs again.
```
mhairimca, line 217: Perhaps I've misunderstood but contracts that expire weekly seems very short.

UMIPs/umip-72.md
ccashwell: @mhairimca not sure I understand the question. The formula is quite complex and it's detailed in the white paper. Do you have any examples of what you'd like to see here?

UMIPs/umip-72.md
ccashwell: The response from `/api/latest` now matches the expected format.

UMIPs/umip-72.md
ccashwell: Nope you're correct, we chose weekly expiration because crypto volatility is, eh, quite volatile! Looking at the past two weeks (which have been a period of relatively very low volatility) the open and close are fairly close, but the high and low span a range of nearly 50%. In order to deploy a "safe" instrument under UMA's EMPs, we feel the only reasonable thing to do is limit the upper and lower bounds of convexity by choosing a shorter expiry period.

Volatility is the measure of variance over time. If we extend the time of the contracts to one month the minimal collateral requirements will need to be very high (e.g. 200%-300%). The shorter the duration the lower we can safely set this ratio. Higher collateral requirements, particularly when paired with a fairly high likelihood of liquidation, is a clear disincentive for liquidity providers.

Ultimately this is a trade-off that we think would be best left up to the community as part of a future dVIX Governance Proposal: "Would you rather have longer contracts that require much more capital with potentially higher liquidation risk, or shorter contracts that expire more often but require substantially less collateral?"

UMIPs/umip-72.md
chrismaree: cool. I can see the `iVix` in the link you provided.

UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
...
+Manipulating the bids/asks on Deribit is the main vector of attack for changing the ethVIX price. Two things help to mitigate this attack. First, placing bids/asks that are not aligned with the market can be costly to an attacker if they are filled. Second, in certain cases when bids/asks become extremely skewed from their expected market prices, the price feeds use Deribit’s mark price instead of the bid/ask.
+You can learn more about the approach used to calculate ethVIX in the [methodology paper](https://dvix.io/ethVIX_methodology.pdf).
+
+The settlement price calculation and the standard price request calculation are the same for both the ethVIX and iethVIX. The settlement price calculation is always the value of the price feeds at 06:00 UTC on the day of the contract expiration.
+	
+# IMPLEMENTATION
+
+1. What prices should be queried for and from which markets?
+
+    - dVIX protocol ethVIX & iethVIX price feeds
```
chrismaree, line 131: the ethVix and iethVix price feed are just APIs. I believe that this question in the Implementation section is in reference the the _UMA_ price feeds. To correctly spin up bots to support this financial product we'll require a UMA price feed to be wrapped around the `dvix.io/api/...` endpoint. You can see the existing price feeds here: https://github.com/UMAprotocol/protocol/tree/master/packages/financial-templates-lib/src/price-feed

The implementation of this feed (simply wrapping an API) will likely be very similar to the [CrypoWatchPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js) or the [DeFiPulsePriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefiPulsePriceFeed.js) which simply query external APIs.

ccashwell: I see, this is actually here: https://github.com/UMAprotocol/protocol/pull/2792

UMIPs/umip-72.md
smb2796: Can you update this and the name of the file to the name of your price identifiers?

UMIPs/umip-72.md
```diff
@@ -0,0 +1,217 @@
+# Headers
+
+| UMIP-72 | |
+|---------|-|
+| UMIP Title | ethVIX & iethVIX |
```
smb2796, line 5: Can you update the title to match the format in other price identifiers?

UMIPs/umip-72.md
smb2796: Since there seems to be no base or quote currency, I think it'd be better to just exclude these fields.

UMIPs/umip-72.md
ccashwell: Done ✅

UMIPs/umip-72.md
ccashwell: Done ✅

UMIPs/umip-72.md
ccashwell: I think there is a "quote" but not a "base" currency here. The quoted and collateralized price is indeed USDC. Please confirm your understanding and/or desire and I'll update accordingly.

UMIPs/umip-72.md
pemulis: Cool!

smb2796 (approved):
It looks like the outstanding issues have been resolved and also the price feed is about ready to merge. Since these were the blockers that we identified in the community call, this seems to meet the criteria to get merged in and moved to Last Call. 

Going to approve and merge in, but please feel free to submit additional prs if you wish to edit anything further before the vote starts.

PR #230, opened 2021-04-01
Add "1" Price identifier

rossgalloway:
New Price identifier that always equals 1, to be used when the full financial product library is not needed


nicholaspai:
@rossgalloway thoughts on generalizing this identifier to be a "constants" identifier that we can transform to any constant that we need? For example, the uTVL options require the constant "2" while the call options require the constant "1"

We could take advantage of [ancillary data](https://github.com/UMAprotocol/protocol/pull/2229) to get this functionality. @chrismaree do you think this is possible with ancillary data?

rossgalloway:
@nicholaspai @smb2796 I have royally borked my github flow and want to delete this pull request to fix it. Want to give you guys a heads up before I do. I will make another and we can re-add the comment. Or if you guys have plans or ideas to change this to accept any constant then maybe I can make those changes. Let me know. Otherwise I will delete and resubmit on Apr. 5th.

smb2796:
> @nicholaspai @smb2796 I have royally borked my github flow and want to delete this pull request to fix it. Want to give you guys a heads up before I do. I will make another and we can re-add the comment. Or if you guys have plans or ideas to change this to accept any constant then maybe I can make those changes. Let me know. Otherwise I will delete and resubmit on Apr. 5th.

What is borked in your github flow? Are you referring to the DCO issue?

I also think there is an outstanding question around if this identifier is actually necessary/the best way to build your contract. I was just musing on this in slack, and I think it warrants further thought.

rossgalloway:
> 
> 
> > @nicholaspai @smb2796 I have royally borked my github flow and want to delete this pull request to fix it. Want to give you guys a heads up before I do. I will make another and we can re-add the comment. Or if you guys have plans or ideas to change this to accept any constant then maybe I can make those changes. Let me know. Otherwise I will delete and resubmit on Apr. 5th.
> 
> What is borked in your github flow? Are you referring to the DCO issue?
> 
> I also think there is an outstanding question around if this identifier is actually necessary/the best way to build your contract. I was just musing on this in slack, and I think it warrants further thought.

Not related to the DCO issue. I don't even know what that is.  It is borked because I want to make a separate pull request for YFI as collateral and can't seem to do it without it getting included in this pull request. Hence why you see a created and then deleted "YFI as collateral" file included above.

I have made the current pull request from the master branch of my cloned repository instead of creating separate branches for each pull request. At least I think that is how I should do it. Trying to figure this all out on the fly.

PR #232, opened 2021-04-06
Add RAI Redemption Rate APR as Price Identifiers

ashutoshvarma:


UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+
+This implementation works with [RAI subgraph](https://subgraph.reflexer.finance/subgraphs/name/reflexer-labs/rai).
+To use it with other data sources, modify the `get_rates()` function.
+
+The script below takes `end_timestamp` and calculates the 30 days geometric mean of redemption rate coefficient  APR. UMA token holders should define these as follows.
+```
+end_timestamp = DVM timestamp
+```
+
+```python
```
pemulis, line 308: I don't know if this is a blocker, but would it be possible to get a JavaScript implementation in the [protocol](https://github.com/UMAprotocol/protocol) repo, so that the bots can handle this price feed? It should be an extension of the [PriceFeedInterface](https://github.com/UMAprotocol/protocol/tree/master/packages/financial-templates-lib/src/price-feed/PriceFeedInterface.js) and go in [this directory](https://github.com/UMAprotocol/protocol/tree/master/packages/financial-templates-lib/src/price-feed) (which also has examples).

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+The price resolution method to use will depend on the timestamp the price request was made at.
+
+For a price request made at or after the expiry timestamp, the price will be resolved with the RAI annualized redemption rate
+coefficient calculation defined as per R3 token Technical Specification.
+
+For a price request made before the expiry timestamp, the price will be resolved to a 2-hour TWAP for the Uniswap price of the
+listed synthetic token in RAI.
+
+## R3-APR21/RAI
+The R3 token tied to this price identifier will expire on the cutoff timestamp `1619568000` (28 April 2021 12:00:00 AM).
```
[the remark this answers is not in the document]
ashutoshvarma: Yeah, they are UTC, I will append UTC to clear confusion

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+```
+CUTOFF = 1619568000         # 28 April 2021 12:00:00 AM
```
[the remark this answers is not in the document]
ashutoshvarma: I would like to keep it monthly, i.e expiry at month-end like others (uGas, CAR), for consistency.

We can make it 28th May and 28th June and remove R3-APR21 entirely
- R3-MAY21 - 28 May2021 12:00:00 AM UTC
- R3-JUN21 - 28 June 2021 12:00:00 AM UTC

Will that be fine?

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
ashutoshvarma: This script is only for DVM price requests after the EMP expiry, while synths are live Uniswap price will be used for liquidations, so bots only need to concern with Uniswap pool price, price feed for that is already implemented in UniswapPriceFeed.js.

Also, I was planning to create PR in `protocol` for adding default config for R3 tokens once UMIPs are approved and synths and pool exists

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+### Pre-cutoff
+
+For price requests made before the cutoff, (`1619568000` for `R3-APR21/RAI` and `1622160000` for `R3-MAY21/RAI`),
+use the same two-hour TWAP calculation implementation from UMIP-22.
+
+1. The end TWAP timestamp equals the price request timestamp.
+2. The start TWAP timestamp is defined by the end TWAP timestamp minus the TWAP period (2 hours in this case).
+3. A single Uniswap price is defined for each timestamp as the price that the RAI/R3 pool returns at the end of the latest block whose timestamp is <= the timestamp that is queried for.
+4. The TWAP is an average of the prices for each timestamp between the start and end timestamps. Each price in this average will get equal weight.
+5. As the token price will already implicitly be tracking the R3-MAY21/RAI/RAI or R3-APR21/RAI/RAI price, it should be left as returned without any scaling transformation.
```
Reinis-FRP, line 283: If users are calculating TWAP from [Uniswap formula](https://uniswap.org/docs/v2/core-concepts/oracles/) by taking `price0CumulativeLast` at different blocks from an archive node then the result still has to be adjusted for different pair token `decimals`. Or is it decided that R3 will have the same `decimals` as `RAI`?

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+   However, RAI on Ethereum uses 18 decimal places so voters must submit 1.380000000000000000 RAI.
+
+
+### Post-cutoff
+Since redemption rate can be negative also and EMPs don't support negative indexes as of now. We will use the Redemption Rate Coefficient for the settlement of synths.
+```
+Redemption Rate Coefficient = Redemption Rate + 1
+```
+
+RAI OracleRelayer contact's `redemptionRate()` and subgraphs both returns Redemption Rate Coefficient
```
Reinis-FRP, line 295: Just comparing the output from RAI OracleRelayer contract's `redemptionRate()` and subgraph it seems that the result from `redemptionRate()` call should be divided by `10 ^ 27`

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+
+# MOTIVATION
+
+The DVM currently does not support reporting 30 days medianised RAI redemption rate coefficient or a two-hour TWAP of the R3 token.
+
+1. What are the financial positions enabled by creating this synthetic that does not already exist?
+
+    - The usage of a two-hour TWAP for liquidations, then an aggregated price at expiration is similar to UMIP-22.
+      This setup allows us to create synthetic futures. This will enable RAI holders (those who hold, speculate on or use RAI in other protocols and apps)
+      to hedge against redemption rate volatility which allows them to "lock in" future returns from redeeming or selling RAI.
```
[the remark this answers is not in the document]
ashutoshvarma: Good point! I will add a link along with one-two line summary

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+    Now at the end of the month if average redemption increases then synth will settle for a higher price and he/she will lose some
+    collateral but the increase in RAI market price (due to redemption rate) will cover the cost of lost collateral.
+
+<br> 
+
+# MARKETS & DATA SOURCES
+
+### Pre Cutoff
+This price should be queried from the highest volume Uniswap pool (Whichever one is deemed more legitimate by the UMA token holders)
+on Ethereum where at least one asset in the pair is R3. It's expected that a single Uniswap pool will have 99% of the liquidity and
```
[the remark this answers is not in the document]
ashutoshvarma: You are right, no need to make things complicated unnecessarily.
We should stick with `RAI/R3` pool.

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+
+This price identifier aggregates the value returned by `redemptionRate()` over every block from the 30 days ending at the cutoff/expiration
+timestamp (1619568000 for R3-APR21/RAI and 1622160000 for R3-MAY21/RAI).
+
+It is recommended to gather the raw data from an Ethereum archive node. Alternatively, this annualized value is indexed in the
+[RAI Subgraph]( https://subgraph.reflexer.finance/subgraphs/name/reflexer-labs/rai).
+As of writing this UMIP, the RAI Subgraph is free and will remain so.
+
+All of these endpoints give the redemption rate coefficient per block for historical blocks, however, converting the data from RAI OracleRelayer
+to APR (subgraph already has annualized rate) is required, and 30-day median aggregation is still required for both sources. (Read the implementation section)
```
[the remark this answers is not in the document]
ashutoshvarma: I see, will add bullet points explaining steps explicitly

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
ashutoshvarma: [RAI](https://etherscan.io/token/0x03ab458634910aad20ef5f1c8ee96f1d6ac54919) has 18 decimals and synths minted from EMPs also has 18 decimals. So I think there is no need to scale the result.

Should I change it to something like this ?
> Since both R3 and RAI have the same decimals, it should be left as returned without any scaling transformation.

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
ashutoshvarma: Yes, on-chain it is stored in RAY decimals, i.e 27 decimals, so to convert this we need to divide with `10 ^ 27`

I will update the `Implementation, especially the Post cutoff part.

Looks like, I missed a hell lot of things 😅 
Thanks for the review and pointing them out

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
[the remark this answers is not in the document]
pemulis: Yeah, that would be fine!

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
pemulis: That's true for the liquidator bot, but there are also proposer and disputer bots that would need a script to run at expiry. E.g., the proposer could answer price requests to the optimistic oracle, and the disputer could dispute incorrect settlement prices proposed by someone else.

ashutoshvarma: I see, this is different from PriceFeeds, so where should this script go in `protocol` repo ?

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+| Link to Discourse    | [LINK](https://discourse.umaproject.org/t/add-rai-redemption-rate-apr-as-price-identifiers/624)                                                                               
+
+
+# SUMMARY 
+
+This UMIP will reference a synthetic token to be created with this price identifier.
+This token will be referred to as 'R3' (Rai Redemption Rate) and will represent the token that tracks these identifiers.
+
+The DVM should support price requests for the R3-APR21/RAI and R3-MAY21/RAI price indices.
+
```
mhairimca, line 17: This PI will not be approved and ready for use until 18th April which makes the run of the April token extremely short

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
ashutoshvarma: @pemulis 
https://github.com/UMAprotocol/protocol/blob/175a4dcc5bc98a8fd2ea5a596219e0baece89100/packages/optimistic-oracle/src/proposer.js#L96-L99
Seems the proposer bot ignores price requests from self-referential synths for now, and R3 is self-referential, do we really need a price feed implementation for post expiry?

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+# PRICE FEED IMPLEMENTATION
+
+For the creation of the R3 token, it is desired that the DVM return either the 30-day median annualized rate, or a 2-hour TWAP on the market price of R3.
+The type of price that the DVM will return is dependent on the timestamp the price request is made at.
+This timestamp is the expiry timestamp of the contract that is intended to use this price identifier,
+so the TWAP calculation is used pre-expiry and the closing index value of the R3 calculation is used at expiry.
+
+This is very similar to the uGAS token and this design is outlined in [UMIP 22](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-22.md).
+
+Because the redemptionRate is only used at expiry, it will not be possible for a token sponsor to become undercollateralized based upon its movement.
```
smb2796, line 186: This is slightly different now since the EMP now uses the optimistic oracle to settle by default. It is somewhat necessary to also have a price feed that can get the settlement price.

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
smb2796: Yes, it would be best. The blacklist for those self-referential synths was more of a hack-around to support synths using legacy price identifiers. It creates some difficulty with settlement - prices need to be manually proposed - that isn't optimal.

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+
+**5. Collateral Decimals** - 18
+
+**6. Rounding** - Round to nearest 2 decimal places (third decimal place digit >= 5 rounds up and < 5 rounds down) 
+
+<br>
+
+# RATIONALE
+### Pre-cutoff
+
```
smb2796, line 240: Why is the self-referential model necessary?

smb2796 (commented):
Great submission @ashutoshvarma!

A few comments:

1. Like @pemulis mentioned, it would be best to have a price feed to get the RAI redemption rate.
2. Self-referential price identifiers are somewhat dangerous by design and I wonder if there is another way that we could approach this. Without a large amount of starting liquidity in the referenced synth's pool, the product becomes somewhat unusable because sponsors could be quickly liquidated as a result of small purchases of 'R3'. What was the reasoning for using a self-referential model, and do you think just using the rolling RAI redemption rate would work instead?
3. If the self-referential method is used, the better method to go about doing this would be to propose two price identifiers - one that is the rolling RAI redemption rate, and one that is the self-referential token price. Then using a financial product library, similar to this [one](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol), you could switch the price identifier being used at expiry.

ashutoshvarma:
Thanks, @smb2796 
1.) I'm actually working on it as of now, it will calculate the rolling geometric mean of redemption rate. Which with the use of ExpressionPriceFeed.js can be converted to APR.

2.) Yeah, I'm aware of the instability of self ref synths. I actually discussed this with Evan, he is also concerned regarding this. But still, since these synths are tracking assets that don't have any real-world value, so in case there is a huge disparity between the price feed's price and trading price, the sponsors might intentionally under-collateralize their positions.
For example, if eth shoots up, which means the redemption rate will increase shortly (RAI dampened the eth movement - https://reflexer.finance/faq/), and the R3 doubles in anticipation of the increase in redemption rate.  Now the sponsor will mint the synths and sell for a price higher than price feed, and their profit might be larger than collateral locked which essentially causes intentional liquidations.

3.) That cool! I don't know about this. Actually, I was thinking of a price feed implementation that switches the implementation based on cutoff timestamp
https://discord.com/channels/718590743446290492/718591751543062548/831090769531633704

I'm also not keen on using self ref method, but after reading uSTONKs and uGAS UMIPs I couldn't think of anything better than this. 

If anybody has any suggestion regarding this, do share

ashutoshvarma:
@smb2796 can the risk for intentional liquidations be mitigated without using self ref if we set a high min CR value (2 or 3 instead of 1.25)

Let assume a scenario when we are only using rolling redemption rate as price feed :-
- Redemption Rate (price feed) is $1 and the market price of R3 is also $1 
- ETH shoots up, R3 market price doubles to $2 in anticipation of the redemption rate increase as RAI follow ETH movements
- I go in and mint 100 R3 and withdrawn down to 200% CR (i.e $200 in collateral left) as the price feed is still at $1. Then sell them at the market for $200
- Shortly RAI follows ETH and price feed (redemption rate) increase to $1.5. Now I am under undercapitalized (have $200 but need $300 in collateral)

In this case, profit and collateral are equal if including gas cost, it would still result in a loss if liquidated.
So setting min CR to 2 can somehow prevent till 2x increase. But in order to on the safer side users then have set much higher CRs for their positions.

Am I understanding things right?

pemulis:
@ashutoshvarma do you think you would be able to address the comments in time for the community discussion on Monday? Pretty excited to add this when it's ready, since it's a really cool price feed. Also, can you mark any comments that you fixed in new commits as Resolved, to make it easy to scan the thread and see what's left?

ashutoshvarma:
It's been a pretty busy week for me, but rolling redemption rate price feed implemention is almost complete, just some tests are remaining.

> do you think you would be able to address the comments in time for the community discussion on Monday?

I will try to complete this by tomorrow, so yeah.

> can you mark any comments that you fixed in new commits as Resolved, to make it easy to scan the thread and see what's left?

Sure

Also what are your views on https://github.com/UMAprotocol/UMIPs/pull/232#issuecomment-817920232

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
mhairimca: The UMIP will now not be available until 1st May, which makes the Apr token redundant

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
Reinis-FRP: > [RAI](https://etherscan.io/token/0x03ab458634910aad20ef5f1c8ee96f1d6ac54919) has 18 decimals and synths minted from EMPs also has 18 decimals. So I think there is no need to scale the result.
> 
> Should I change it to something like this ?
> 
> > Since both R3 and RAI have the same decimals, it should be left as returned without any scaling transformation.

OK

smb2796:
> @smb2796 can the risk for intentional liquidations be mitigated without using self ref if we set a high min CR value (2 or 3 instead of 1.25) ?
> 
> Let assume a scenario when we are only using rolling redemption rate as price feed :-
> 
> * Redemption Rate (price feed) is $1 and the market price of R3 is also $1
> * ETH shoots up, R3 market price doubles to $2 in anticipation of the redemption rate increase as RAI follow ETH movements
> * I go in and mint 100 R3 and withdrawn down to 200% CR (i.e $200 in collateral left) as the price feed is still at $1. Then sell them at the market for $200
> * Shortly RAI follows ETH and price feed (redemption rate) increase to $1.5. Now I am under undercapitalized (have $200 but need $300 in collateral)
> 
> In this case, profit and collateral are equal if including gas cost, it would result in a loss if liquidated.
> So setting min CR to 2 can somehow prevent till 2x increase. But in order to on the safer side users then have set much higher CRs for their positions.
> 
> Am I understanding things right?

This is a valid concern, but I think the solution you proposed would mitigate most of this concern. It would be much harder to withdraw to an undercapitalized position size if your min CR was high.

What kind of movement in ETH would it take to justify an expected 100% movement in the 30 day RAI redemption rate median? Do you have any volatility stats for the RAI redemption rate that you can provide?

Another potential solution would be to reduce the length of the period that you are medianizing over. If, for example, you were tracking the 1 week RAI redemption rate, there should be much less "lag" or difference between the R3 spot and the actual RAI redemption rate. Thoughts on this?

pemulis:
> Also what are your views on [#232 (comment)](https://github.com/UMAprotocol/UMIPs/pull/232#issuecomment-817920232)

I agree with @smb2796's notes above and think this could be addressed with a high enough minimum collateralization ratio.

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+The DVM should support requests for a price that resolves to either the median monthly RAI annualized redemptionRate
+coefficient or a 2-hour Time-Weighted Average Price (TWAP) on the highest volume Uniswap RAI/R3 pool.
+The price resolution method to use will depend on the timestamp the price request was made at.
+
+For a price request made at or after the expiry timestamp, the price will be resolved with the RAI annualized redemption rate
+coefficient calculation defined as per R3 token Technical Specification.
+
+For a price request made before the expiry timestamp, the price will be resolved to a 2-hour TWAP for the Uniswap price of the
+listed synthetic token in RAI.
```
therealhexi, line 26: Formatting: Recommend using * and move it directly under the previous paragraph, rather than using two new paragraphs. These paragraphs conceptually belong to the previous paragraph and elaborates line 20

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
therealhexi: +1

Recommend to add some lead time when drafting the UMIP for the first time. Could even more expiry date to June

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+
+
+# MOTIVATION
+
+The DVM currently does not support reporting 30 days medianised RAI redemption rate coefficient or a two-hour TWAP of the R3 token.
+
+1. What are the financial positions enabled by creating this synthetic that does not already exist?
+
+    - The usage of a two-hour TWAP for liquidations, then an aggregated price at expiration is similar to UMIP-22.
+      This setup allows us to create synthetic futures. This will enable RAI holders (those who hold, speculate on or use RAI in other protocols and apps)
```
therealhexi, line 56: Generally would avoid use words `us` in UMIP - does it make sense to use a more subjective pronoun such as RAI stakeholders?

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+
+
+1. What sources should the price be queried from? It is recommended to have at least 3 sources.
+
+    - RAI Official Subgraph -  https://subgraph.reflexer.finance/subgraphs/name/reflexer-labs/rai
+    - APR of RAI OracleRelayer contact's `redemptionRate()`
+
+4. How often is the provided price updated?
+
+    - Every 4 hours (14,400s), (see [`updateRateDelay`](https://docs.reflexer.finance/current-system-parameters#ratesetter))
```
therealhexi, line 149: Does it make any sense to have a 2 hour TWAP if the price feed itself is updated every 4 hours?

UMIPs/umip_add_R3-APR21_and_R3-MAY21_price_identifiers.md
```diff
@@ -0,0 +1,444 @@
...
+
+6.  Do these sources allow for querying up to 74 hours of historical data? 
+
+    - Yes
+
+7.  How often is the provided price updated?
+
+    - Every 4 hours (14,400s), (see [`updateRateDelay`](https://docs.reflexer.finance/current-system-parameters#ratesetter))
```
therealhexi, line 157: This is duplicate as line 147

therealhexi (commented):
Formatting and documentation-based comments. No major blocker.

gitofdeepanshu:
Hey guys,
As we have discussed before, we are no longer using self-ref for collateralization.
R3-10H-TWAP (Pre-cutoff):   10-hr TWAP of redemption rate coefficient APR
R3-30D-GM (Post-cutoff):  30 day Geometric Mean of redemption rate
This still enable us to create synthetic futures for hedging and also address Sean's concern regarding self-ref.

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,301 @@
...
+1. The end TWAP timestamp equals the price request timestamp.
+2. The start TWAP timestamp is defined by the end TWAP timestamp minus the TWAP period (10 hours in this case).
+3. A single redemption rate coefficient APR value is defined for each timestamp as the price that the subgraph
+   returns at the end of the latest block whose timestamp is <= the timestamp that is queried for.
+4. The TWAP is an average of the rates for each timestamp between the start and end timestamps.
+   Each price in this average will get equal weight.
+5. As the values from subgraph are integers, so it should be left as returned without any scaling transformation.
+6. The final price should be submitted with 18 decimals but rounded to 2 decimal places. For example, if the value is
+   1.384827478767976545678765456, then we round to 2 decimal places and convert 1-to-1 to RAI for 1.38 RAI.
+   However, RAI on Ethereum uses 18 decimal places so voters must submit 1.380000000000000000 RAI.
```
mhairimca, line 246: @smb2796 iCan we acceot 1,38 rather thanf force the trailing zeros?

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,301 @@
...
+[Here](https://github.com/ashutoshvarma/protocol/blob/rai-price-feed/packages/financial-templates-lib/src/price-feed/RAIRedemptionRate.js)
+is a reference implementation for an off-chain price feed that can both 10-hr TWAP and 30-day GM of redemption rate coefficient .
+
+<br>
+
+# TECHNICAL SPECIFICATIONS
+
+## R3-10H-TWAP
+
+**1. Price Identifier Name** - R3-10H-TWAP
```
mhairimca, line 139: Underscores are more common in names

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,301 @@
...
+**3. Quote currency** - RAI
+
+**4. Intended Collateral Currency** - RAI
+
+- Does the value of this collateral currency match the standalone value of the listed quote currency?
+
+  - YES
+
+- Is your collateral currency already approved to be used by UMA financial contracts? If no, submit a UMIP to have the desired collateral currency approved for use.
+  - NO, but there is a separate pending UMIP for approving RAI as collateral currency
```
mhairimca, line 174: RAI was approved as a collateral currency in UMIP 77

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,301 @@
...
+updated every 4 hours
+
+### R3-30D-GM (Post-cutoff)
+
+We're using a rolling 30 day period to increase the cost of manipulating the APR price.
+
+We use geometric mean, as opposed to arithmetic mean, to include the effect of rate compounding.
+
+We use an annual percentage rate, as opposed to a 24hr percentage rate or per-second rate, to increase usability.
+
```
mhairimca, line 203: Stylistically it would be better for this to be written in the third person.

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,301 @@
...
+<br>
+
+# PRICE FEED IMPLEMENTATION
+
+For the creation of the R3 token, it is desired that the DVM return either the 30 day geometric mean or a 10-hour TWAP.
+The type of price that the DVM will return is dependent on the timestamp the price request is made at.
+
+[Here](https://github.com/ashutoshvarma/protocol/blob/rai-price-feed/packages/financial-templates-lib/src/price-feed/RAIRedemptionRate.js)
+is a reference implementation for an off-chain price feed that can both 10-hr TWAP and 30-day GM of redemption rate coefficient .
+
```
mhairimca, line 132: @pemulis  - could you check the price feed implementation to make sure it will work with our contracts?

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,301 @@
...
+### R3-30D-GM (Post-cutoff)
+
+We're using a rolling 30 day period to increase the cost of manipulating the APR price.
+
+We use geometric mean, as opposed to arithmetic mean, to include the effect of rate compounding.
+
+We use an annual percentage rate, as opposed to a 24hr percentage rate or per-second rate, to increase usability.
+
+The ground truth data for this is in the RAI OracleRelayer smart contract on the Ethereum blockchain.
+Any differences in UMA governor results for this price identifier should be due to rounding errors that propagate through the calculation (numerical instability)
```
mhairimca, line 205: If rounding is done at the end, this should not be an issue.

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,301 @@
...
+| Link to Discourse | [LINK](https://discourse.umaproject.org/t/add-rai-redemption-rate-apr-as-price-identifiers/624) |
+
+# SUMMARY
+
+This UMIP will reference a synthetic token to be created with this price identifier.
+This token will be referred to as 'R3' (Rai Redemption Rate) and will represent the token that tracks these identifiers.
+
+The DVM should support price requests for the R3-10H-TWAP and R3-30D-GM price indices.
+
+The DVM should support requests for a price that resolves to either the 30 day geometric mean (R3-30D-GM) or
```
mhairimca, line 18: I assume the rate is always positive if you are using a geometric mean - could there ever be a negative rate?

ashutoshvarma: Both price feeds use a redemption rate coefficient which cannot be negative.

UMIPs/umip_add_R3_price_identifiers.md
smb2796: This looks really good! @ashutoshvarma can you submit this as a PR [here](https://github.com/UMAprotocol/protocol/tree/master/packages/financial-templates-lib/src/price-feed)?

UMIPs/umip_add_R3_price_identifiers.md
smb2796: Yes we can. All price identifiers are treated as 18 decimals by default, so it could just be submitted as 1.38 which would then be converted by the voter dapp to `1380000000000000000` before being submitted. @ashutoshvarma this is the desired behavior, correct?

UMIPs/umip_add_R3_price_identifiers.md
ashutoshvarma: Sure

UMIPs/umip_add_R3_price_identifiers.md
ashutoshvarma: Yes

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,302 @@
...
+| Status            | Draft                                                                                           |
+| Created           | April 1, 2021                                                                                   |
+| Link to Discourse | [LINK](https://discourse.umaproject.org/t/add-rai-redemption-rate-apr-as-price-identifiers/624) |
+
+# SUMMARY
+
+This UMIP will reference a synthetic token to be created with this price identifier.
+This token will be referred to as 'R3' (Rai Redemption Rate) and will represent the token that tracks these identifiers.
+
+The DVM should support price requests for the R3_10H_TWAP and R3_30D_GM price indices.
```
smb2796, line 16: Just curious, is the intention to use R3_10H_TWAP pre-expiry and R3_30D_GM at expiry?

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,302 @@
...
+
+The DVM should support requests for a price that resolves to either the 30 day geometric mean (R3_30D_GM) or
+a 10-hour Time-Weighted Average Price (R3_10H_TWAP) of RAI redemption rate coefficient APR
+
+The price resolution method to use will depend on the timestamp the price request was made at.
```
smb2796, line 21: I really like this solution! Since there are now two price identifiers and I'm assuming you all are intending to use the FPL to switch price identifiers at expiry, it doesn't seem like the voters will need to factor the timestamp into which calculation methodology they should use. This is pretty awesome, but also makes this language seem a little misleading.

Wondering if you can soften this language a bit and explain that the intention is to ues the R3_10H_TWAP pre expiry etc, but voters should always return the value based off of which price identifier is being requested.

UMIPs/umip_add_R3_price_identifiers.md
smb2796: Thanks!

UMIPs/umip_add_R3_price_identifiers.md
smb2796: Awesome, I think the 18 decimal place instructions can be removed then.

smb2796: Awesome, I think the 18 decimal place instructions can be removed then.

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,302 @@
...
+
+# SUMMARY
+
+This UMIP will reference a synthetic token to be created with this price identifier.
+This token will be referred to as 'R3' (Rai Redemption Rate) and will represent the token that tracks these identifiers.
+
+The DVM should support price requests for the R3_10H_TWAP and R3_30D_GM price indices.
+
+The DVM should support requests for a price that resolves to either the 30 day geometric mean (R3_30D_GM) or
+a 10-hour Time-Weighted Average Price (R3_10H_TWAP) of RAI redemption rate coefficient APR
```
chrismaree, line 19:
```suggestion
a 10-hour Time-Weighted Average Price (R3_10H_TWAP) of RAI redemption rate coefficient APR.
```

UMIPs/umip_add_R3_price_identifiers.md
ashutoshvarma: Yeah, we will be using FPL to switch between price ids.

> Wondering if you can soften this language a bit 

```suggestion
The price resolution method for the R3 token will depend on the timestamp the price request was made at
but voters should always return the value-based off on which price identifier is being requested.
```
How does this sound?

UMIPs/umip_add_R3_price_identifiers.md
ashutoshvarma: Then, can it be removed?

UMIPs/umip_add_R3_price_identifiers.md
```diff
@@ -0,0 +1,300 @@
...
+### R3_10H_TWAP (Pre-cutoff)
+
+For price requests made before the cutoff, use this 10-hour TWAP calculation implementation.
+
+1. The end TWAP timestamp equals the price request timestamp.
+2. The start TWAP timestamp is defined by the end TWAP timestamp minus the TWAP period (10 hours in this case).
+3. A single redemption rate coefficient APR value is defined for each timestamp as the price that the subgraph
+   returns at the end of the latest block whose timestamp is <= the timestamp that is queried for.
+4. The TWAP is an average of the rates for each timestamp between the start and end timestamps.
+   Each price in this average will get equal weight.
```
Reinis-FRP, line 243: It is bit misleading to call it TWAP (time weighted) then. E.g., if during the 10h period there are 2 APR values that were valid for 4h each and the 3rd APR value has been valid only the last 2h within 10h period - should all 3 APR values get the same weight?

ashutoshvarma: I actually took the TWAP calculation from UMIP-22. That look odd to me also. 

Since my price feed implementation uses the same [function to calculate the TWAP](
https://github.com/UMAprotocol/protocol/blob/b8225c580ea48f58ef44aa308f966fbed5a99cf3/packages/financial-templates-lib/src/price-feed/utils.js#L271-L304) as UniswapPriceFeed, which is used by uGAS, uSTONKS and many other feeds, I assume that it is the accepted definition.
Now looking back to `computeTWAP` it actually calculates the proper TWAP instead of simply average.

@smb2796 is the TWAP calculation from UMIP-22 correct?

UMIPs/umip_add_R3_price_identifiers.md
mhairimca: dont see why not

UMIPs/umip_add_R3_price_identifiers.md
ashutoshvarma: RR = Redemption rate coefficient value
TT = Total time window (10hrs)
TI = Time interval for which RR value persisted

1. The end TWAP timestamp equals the price request timestamp.
2. The start TWAP timestamp is defined by the end TWAP timestamp minus the TWAP period (10 hours in this case).
3. A single redemption rate coefficient APR value (RR) is defined for each timestamp as the price that the subgraph
   returns at the end of the latest block whose timestamp is <= the timestamp that is queried for.
4. The TWAP is a time weighted average of the rates for each timestamp between the start and end timestamps.
   Each price in this average will get weight equal to its time period (TI). TWAP will be calculated as ∑(RR * TI) / TT
5. As the values from subgraph are integers, so it should be left as returned without any scaling transformation.
6. The final price should be submitted with 18 decimals but rounded to 2 decimal places. For example, if the value is
   1.384827478767976545678765456, then round to 2 decimal places and convert 1-to-1 to RAI for 1.38 RAI.
   
   
   is this good ? @Reinis-FRP

smb2796 (approved):
It looks like the TWAP explanation was fixed, and most other major issues were resolved.

One other leftover one from the community call, and something that has been discussed offline, was providing a method for contract deployers to run their own RAI subgraphs. I'm going to merge in because of time constraints, but also add a link the the RAI subgraph code and reference the most current commit hash as the subgraph implementation that voters and bot operators should reference. Please let us know if this differs from your expectations in any way, and we can update the UMIP before voting.

PR #233, opened 2021-04-06
Add RAI as Approved Collateral Currency 

ashutoshvarma:


pemulis:
This should probably be paired with a price identifier UMIP that would approve `RAIUSD` and `USDRAI` price identifiers, which would allow you to value the RAI used as collateral for a synthetic or mint call options for RAI.

ashutoshvarma:
> This should probably be paired with a price identifier UMIP that would approve `RAIUSD` and `USDRAI` price identifiers, which would allow you to value the RAI used as collateral for synthetic or mint call options for RAI.

Yeah, for call options they are indeed required, I will try to write UMIP for this.
Since Monday's call has been announced with objectives, can the new UMIPs be added to it, or they will be reviewed on next call
Both are fine with me.

UMIPs/umip_add_RAI_as_collateral_currency.md
```diff
@@ -0,0 +1,47 @@
...
+- The RAI address, [0x03ab458634910AaD20eF5f1C8ee96F1D6ac54919 ](https://etherscan.io/address/0x03ab458634910AaD20eF5f1C8ee96F1D6ac54919 ),
+  needs to be added to the collateral currency whitelist introduced in UMIP-8.
+- A final fee of 150 needs to be added for RAI in the Store contract.
+
+
+## Rationale
+
+The rationale behind this change is giving deployers more useful collateral currency options.
+
+160 was chosen as the final fee for RAI, because this is approximately ~$500 at current RAI prices whcih is approximately equivalent to other collateral currencies.
```
Reinis-FRP, line 35: 150 is mentioned above in `Technical Specification` - which is correct?

ashutoshvarma: Ohh, My bad
It should be 160 RAI, which was around $500 at the time of writing

pemulis:
> > This should probably be paired with a price identifier UMIP that would approve `RAIUSD` and `USDRAI` price identifiers, which would allow you to value the RAI used as collateral for synthetic or mint call options for RAI.
> 
> Yeah, for call options they are indeed required, I will try to write UMIP for this.
> Since Monday's call has been announced with objectives, can the new UMIPs be added to it, or they will be reviewed on the next call?
> Both are fine with me.

The new RAI/USD UMIP would need to be discussed on the call Monday after next, but I think it's still worth discussing the collateral UMIP this coming Monday, since I believe you're also planning to use RAI collateral for RAI redemption rate options?

UMIPs/umip_add_RAI_as_collateral_currency.md
mhairimca: 160 looks fine.

ashutoshvarma:
Yup, we are also looking into call options

smb2796:
@ashutoshvarma as discussed on the community call, adding RAI as collateral was merged into the other large collateral UMIP for this week to reduce unnecessary votes. You can see this [here](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-77.md#rai-rai-reflex-index). Going to close this PR as a result of this.

PR #239, opened 2021-04-08
Add "1" Price Identifier

rossgalloway:
Recreating this pull request. New price identifier to allow a constant of 1 to be used as a price identifier

### DIFF
--- a/UMIPs/UMIP- Add 1 price identifier.md
+++ b/UMIPs/UMIP- Add 1 price identifier.md
@@ -1,0 +1,153 @@
+## HEADERS
+| UMIP [#]     |                                                                                                                                  |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | 1 Price Identifier                                                                                                |
+| Authors    | Ross - Ross@yam.finance
+| Status     | Draft                                                                                                                                   |
+| Created    | March 31st 2021
+| Link to Discourse   | https://discourse.umaproject.org/t/add-a-1-price-identifier-to-the-dmv/600                                                                             |
+
+<br>
+
+# SUMMARY 
+
+The DMV should support price requests for assets that are priced against themselves and always return a value of 1.
+
+
+# MOTIVATION
+
+1. What are the financial positions enabled by creating this synthetic that do not already exist?
+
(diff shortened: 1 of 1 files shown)


rossgalloway:
Since I re-created this pull request, here is the comment that @nicholaspai posted on the previous one

> @rossgalloway thoughts on generalizing this identifier to be a "constants" identifier that we can transform to any constant that we need? For example, the uTVL options require the constant "2" while the call options require the constant "1"
> 
> We could take advantage of ancillary data to get this functionality. @chrismaree do you think this is possible with ancillary data?

UMIPs/UMIP- Add 1 price identifier.md
```diff
@@ -0,0 +1,153 @@
...
+
+
+**6. Rounding** - 1 decimal place is sufficient
+
+
+<br>
+
+# RATIONALE
+
+- This price identifier is intended to be used in situations where the financial product library is not needed but the price will always return a value of 1 collateral unit.
```
pemulis, line 119: This really feels like something that would be handled in a more straightforward way by a financial product library. It's similar to the covered call library, which always returns a value of `1` prior to expiry, except this library would return `1` both before and after expiry. Our auditor suggested a price feed like this for the covered call library, as well, so that we aren't throwing away oracle prices. However, I think it's actually a lot simpler and easier to reason about to have a financial product library that explicitly throws away the oracle price and returns `1`. It has the same effect and the same number of steps as having a `1` price identifier in practice, but requires less work.

As a comparison, let's say you deployed a financial product library for a 1:1 expires yield token that always expires to the value of 1 token, which is locked up as collateral. You could write it in a generalized way (we did this for the covered call library this week) so that any similar yield token could use the same library. Then, you could set any price identifier as the price to request from the oracle, since it gets thrown out, anyway. The bots are already written to handle existing price feeds, so no extra implementation work there.

Adding a `1` price identifier adds a little bit of extra bot work. Not much, but a little bit. And, imo, it's just a more elaborate way of hardcoding a return value of `1`. The actual issue here is that the EMP isn't really built with this kind of thing in mind, an expiring instrument that has a hardcoded expiry value. You don't need an oracle to tell you that 1=1, but the design of the EMP encourages you to make a price identifier for it.

Here are the features of the EMP that I think you want:
1. Token sponsors can redeem the synth 1:1 with collateral before expiry.
2. Token holders can redeem the synth 1:1 with collateral after expiry.

The oracle call is totally superfluous, but UMA provides a convenient system for deploying contracts that have those two features, so feels like a financial product library thing to me.

smb2796: Great summary and I agree with the approach

rossgalloway: I will close this pull request and focus on using the financial product library. Thanks for the input guys

PR #240, opened 2021-04-08
Add YFI as collateral

rossgalloway:
UMIP to add YFI as a collateral currency


rossgalloway:
looking like this will get covered in the omnibus collateral UMIP that was proposed below this.  Happy to delete and just go with that one.

pemulis:
@rossgalloway I went back and forth on that a minute ago, but think it makes sense to keep it in the omnibus UMIP, since it's already there! Should be easy to consider alongside the other tokens in the collection.

rossgalloway:
@pemulis I will close this pull request then

PR #243, opened 2021-04-12
Create uTVL_KPI_SGT as price identifier

wmuffett:


UMIPs/Add uTVL_KPI_SGT as price identifier
```diff
@@ -0,0 +1,137 @@
+UMIP Create uTVL_KPI_SGT Price Identifier
+UMIP Title: Create KPI Options for SharedStake using SGT as Collateral
```
smb2796, line 2: Can you list the specific price identifiers that are being proposed here (uTVL_KPI_SGT)?

UMIPs/Add uTVL_KPI_SGT as price identifier
```diff
@@ -0,0 +1,137 @@
...
+5.There are a variety of price identifiers for native ERC20 tokens. Calculation of the dollar value should follow the relevant UMIP where it exists, and where there is no Price Identifier UMIP, the three highest volume [ERC20]/USD* markets should be queried, and the median value used.
+•	There is no need for price processing. This is a snapshot based on a particular time; however, it may be useful for TVL Options holders to have oversight of the ongoing TVL and consequently the value of their options on an ongoing basis. There should be a dashboard developed to track the TVL of SharedStake, which will be proposed through their governance site.
+o	The use of dashboards was considered but rejected due to the potential for manipulation and the differences between their calculations of the values of the collateral assets and the value as determined by the relevant UMIPs in previous KPI calculations.
+
+
+IMPLEMENTATION
+•	The contracts to be tracked include
+o	Any EMP contract
```
smb2796, line 105: I'm confused. This looks to be copied over from the uTVL price identifier and, unless I'm mistaken, doesn't apply to this price identifier.

smb2796:
@wmuffett can you change this to a markdown file and adjust the formatting to match other price identifier UMIPs?

UMIPs/Add uTVL_KPI_SGT as price identifier
```diff
@@ -0,0 +1,137 @@
...
+UMIP Title: Create KPI Options for SharedStake using SGT as Collateral
+Authors: Warren Muffett and Chimera
+Create: April 9th, 2021
+
+Summary
+Key Performance Indicator (KPI) options are synthetic tokens that redeem to protocol tokens with the redemption value determined by performance against that indicator. One example of a KPI is Total Value Locked (TVL).
+This UMIP enables the DVM to support price requests based on the TVL of SharedStake.
+A synthetic option is minted against a base collateral, in this case SGT, which expires at 00.00(UTC) December 31th 2021.
+Options are redeemed on the basis of TVL/10^9, with a floor of 0.1 and a ceiling of 2.
+The value locked is calculated by identifying the total values staked within the SGT pool, SGT/ETH LP liquidity mining pool, and vETH2 pool. The dollar value of each of the collateral types is then summed to provide the total value locked.
```
Reinis-FRP, line 11: Is it intended that additional staked assets are added for the calculation of SGT TVL?
If YES, then how DVM voters would know on them - either updates to this UMIP or there is some procedure and source of truth for approved staked assets that should be later included in SGT TVL?
If NO, then this UMIP can be simplified to avoid unnecessary generalizations (e.g. compared to uTVL_KPI_UMA that can have unlimited number of approved collaterals counted as part of TVL)

UMIPs/Add uTVL_KPI_SGT as price identifier
```diff
@@ -0,0 +1,137 @@
...
+2.	Please provide an example of a person interacting with a contract that uses this price identifier.
+•	SharedStake may leverage its community and/or its reputation by minting TVL Options for its token which can be redeemed to a token amount as determined by the TVL of the protocol at the expiry point.
+•	A protocol community member, token holder, voter, or proximal Defi protocol participant may be gifted a TVL option by a protocol as an incentive to build the TVL of the protocol within the option timeframe and redeem at expiry.
+•	Any user may purchase a TVL Option for a protocol that they believe has the potential for growth in TVL prior to expiry.
+3.	The current TVL of SharedStake is approximately $49.4m as at 4:30(UTC) 10th April 2021.
+
+
+Markets and Data Sources
+There are three assets approved as Collateral within SharedStake’s staking contracts. These pools are single sided SGT, SGT/ETH LP, and vETH2 staking. To maintain consistency with existing price identifier UMIPs, it is suggested that TVL is calculated using the deepest Dex pools on Uniswap and Sushiswap.
```
Reinis-FRP, line 25: Where one can check the approved collateral for SharedStake’s staking contracts?

Reinis-FRP, line 25: Does any pool count for TVL calculation? E.g. anyone can setup new SGT pool in Uniswap or other AMM platform. Or there is specific SharedStake’s staking contract that holds staked assets and only this should be used for TVL calculation?

UMIPs/Add uTVL_KPI_SGT as price identifier
wmuffett: Correct, just cleaned this up. Thanks!

UMIPs/Add uTVL_KPI_SGT as price identifier
wmuffett: Need to think about this one. They have added two new staking pools in the past week, so it is possible this could happen again in the future.

UMIPs/Add uTVL_KPI_SGT as price identifier
wmuffett: Added!

wmuffett: This is specific to SharedStake's staking contracts which I clarified in the most recent commit!

UMIPs/Add uTVL_KPI_SGT as price identifier.md
```diff
@@ -0,0 +1,141 @@
...
+5.There are a variety of price identifiers for native ERC20 tokens. Calculation of the dollar value should follow the relevant UMIP where it exists, and where there is no Price Identifier UMIP, the three highest volume [ERC20]/USD* markets should be queried, and the median value used.
+•	There is no need for price processing. This is a snapshot based on a particular time; however, it may be useful for TVL Options holders to have oversight of the ongoing TVL and consequently the value of their options on an ongoing basis. There should be a dashboard developed to track the TVL of SharedStake, which will be proposed through their governance site.
+o	The use of dashboards was considered but rejected due to the potential for manipulation and the differences between their calculations of the values of the collateral assets and the value as determined by the relevant UMIPs in previous KPI calculations.
+
+
+IMPLEMENTATION
+•	The contracts to be tracked include
+•	The total value locked is the dollar value of all three pools on SharedStake.org
```
pemulis, line 112: Others made this point, but you could probably simplify this price identifier quite a lot if you're really going to stick with the three pools and don't plan to accept other types of LP tokens or other assets in the staking pools besides the three you already specified. You could also be more specific in how the price would be calculated, as well, to make settlement easier.

UMIPs/Add uTVL_KPI_SGT as price identifier.md
```diff
@@ -0,0 +1,141 @@
...
+•	The dollar value of each of the contracts should be calculated using the UMIPs and guidance in the Markets and Data section.
+•	These should then be summed to obtain the total value locked (TVL) measured in dollars.
+•	The TVL as measured in dollars should then be divided by 10^9
+1.	What prices should be queried for and from which markets?
+•	The dollar value of each of the contracts should be calculated using the UMIPs and guidance in the Markets and Data section.
+•	Where there is no price identifier UMIP, the price should be queried from the highest volume USD* market
+2.	Pricing interval
+o	This will vary for each value, however, the relevant UMIPs provide a guide to the pricing interval for each
+3.	Input processing
+o	From the list of ExpiringMultiPartyCreator (available at https://github.com/UMAprotocol/protocol/blob/master/packages/affiliates/payouts/devmining-status.json), query all EMP addresses from createExpiringMultiParty() and call pfc() on all of them, repeat same steps for PerpetualCreator and just check the createPerpetual() event, to identify the amount of each collateral type locked in UMA contracts.
```
pemulis, line 122: This needs modifying to reflect your staking pools and not UMA's PfC.

UMIPs/Add uTVL_KPI_SGT as price identifier.md
wmuffett: I know that SharedStake has already added two pools, so I think it might be important to have the ability to add more contracts in the future.

UMIPs/Add uTVL_KPI_SGT as price identifier.md
wmuffett: Perfect! Thank you

UMIPs/Add uTVL_KPI_SGT as price identifier.md
```diff
@@ -0,0 +1,143 @@
+UMIP Title: Create KPI Options for SharedStake using SGT as Collateral; Using Price Identifiers for SGT, SGT/ETH UNI LP, and vETH2
+Status: Draft
+Authors: Warren Muffett and Chimera
```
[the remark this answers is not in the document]
smb2796: Agreed. @wmuffett can you address?

UMIPs/Add uTVL_KPI_SGT as price identifier.md
```diff
@@ -0,0 +1,143 @@
...
+•	The methods used to calculate the dollar value of each of the collateral currencies have been chosen to adhere to previous design decisions in such calculations through UMIPs that have already been approved through our governance procedure however note the following assumptions.
+1.	It is assumed for this purpose that 1vEth2 = 1Eth
+2.	A Liquidity Token (SGT/ETH Uni-V2) is approved to be staked within the SharedStake ecosystem. It is recommended that any additional proposal to add liquidity tokens as approved collateral currencies are evaluated against UMIP 59 and follow those market and data sources to calculate dollar value unless there is good reason for deviance. Note that the calculation of the dollar value of the liquidity token occurs in the penultimate step of the UMIP (Step 5).
+5.There are a variety of price identifiers for native ERC20 tokens. Calculation of the dollar value should follow the relevant UMIP where it exists, and where there is no Price Identifier UMIP, the three highest volume [ERC20]/USD* markets should be queried, and the median value used.
```
smb2796, line 107: A pricing method for SGT needs to be defined.

UMIPs/Add uTVL_KPI_SGT as price identifier.md
```diff
@@ -0,0 +1,143 @@
...
+•	There is no need for price processing. This is a snapshot based on a particular time; however, it may be useful for TVL Options holders to have oversight of the ongoing TVL and consequently the value of their options on an ongoing basis. There should be a dashboard developed to track the TVL of SharedStake, which will be proposed through their governance site.
+o	The use of dashboards was considered but rejected due to the potential for manipulation and the differences between their calculations of the values of the collateral assets and the value as determined by the relevant UMIPs in previous KPI calculations.
+
+
+IMPLEMENTATION
+•	The contracts to be tracked include
+•	The total value locked is the dollar value of all three pools on SharedStake.org
+•	The dollar value of each of the contracts should be calculated using the UMIPs and guidance in the Markets and Data section.
```
smb2796, line 115: This was for UMA's TVL where there were other UMIPs for the collateral types that were being referenced. For this, can you write out a step by step guide that covers:

- Which pools do voters need to reference? How do they get liquidity data for those pools? What assets are part of those pools?
- How do voters get the price of each asset? Something like SGT does not have a SGT/USD price identifier, so getting the SGT/USD price would need to be covered here. Something like ETH can potentially just reference other existing UMIPs.
- What kind of post-processing is needed on the $ value derived from the queried pools?

UMIPs/Add uTVL_KPI_SGT as price identifier.md
```diff
@@ -0,0 +1,143 @@
...
+3.	The current TVL of SharedStake is approximately $49.4m as at 4:30(UTC) 10th April 2021.
+
+
+Markets and Data Sources
+There are three assets approved as Collateral within SharedStake’s staking contracts. These pools are single sided SGT, SGT/ETH LP, and vETH2 staking. To maintain consistency with existing price identifier UMIPs, it is suggested that TVL is calculated using the deepest Dex pools on Uniswap and Sushiswap.
+Additionally, there is a liquidity token that is accepted as collateral at Sharedstake.org, which the TVL can be calculated using the Uniswap liquidity / total circulating supply.
+The contract addresses for the three largest staking pools are as shown below:
+1. SGT: 0xc637dB981e417869814B2Ea2F1bD115d2D993597
+2. SGT/ETH UNIv2 LP: 0x64A1DB33f68695df773924682D2EFb1161B329e8
+3. vETH2: 0xA919D7a5fb7ad4ab6F2aae82b6F39d181A027d35
```
Reinis-FRP, line 31: How do you get number of tokens staked in these contracts?
Do these contracts hold staked tokens or they are delegated somewhere else?
Do these contracts hold future SGT rewards that should not be counted as part of staked assets?

UMIPs/Add uTVL_KPI_SGT as price identifier.md
Reinis-FRP: How can voters verify which staking contracts are added in the future?
Is information on eligible staking contracts available on-chain?
Can some of existing staking contracts be removed in the future?

UMIPs/Add uTVL_KPI_SGT as price identifier.md
```diff
@@ -0,0 +1,141 @@
...
+9.	Is there a cost associated with usage?
+o	See relevant UMIPS
+10.	If there is a free tier available, how many queries does it allow for?
+o	The cost impact of this PI would be negligible.
+11.	What would be the cost of sending 15,000 queries?
+•	There is no need to send 15, 000 queries for this price identifier as it does not require bots.
+
+
+Price Feed Implementation
+As there is no requirement for ongoing monitoring through liquidation or dispute bots, a price feed is not required. The only requirement is a query of the SGT TVL statistic at the timestamp 00.00(UTC) on 31th December according to the data and markets as defined above.
```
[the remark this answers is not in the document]
wmuffett: Would need someone smarter on the technical side to help explain this for me!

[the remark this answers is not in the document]
mrice32: Similar to other TVL options, I think the idea would be that the proposal would be manual (i.e. the bot would throw its hands up and say it can't propose/dispute).

mrice32 (commented):
This looks great!

UMIPs/Add uTVL_KPI_SGT as price identifier.md
```diff
@@ -0,0 +1,146 @@
...
+There are now five assets approved as Collateral within SharedStake’s staking contracts. These pools are single sided SGT, SGT/ETH LP, vETH2 staking, SGT/vETH2 Uniswap LP, and the Saddle vETH2 pool. To maintain consistency with existing price identifier UMIPs, it is suggested that TVL is calculated using the deepest Dex pools on Uniswap and Sushiswap.
+Additionally, there is a liquidity token that is accepted as collateral at Sharedstake.org, which the TVL can be calculated using the Uniswap liquidity / total circulating supply.
+The contract addresses for the five staking pools are as shown below:
+1. SGT: 0xc637dB981e417869814B2Ea2F1bD115d2D993597
+2. SGT/ETH UNIv2 LP: 0x64A1DB33f68695df773924682D2EFb1161B329e8
+3. vETH2: 0xA919D7a5fb7ad4ab6F2aae82b6F39d181A027d35
+4. SGT/vETH2: 0x53dc9d5deb3b7f5cd9a3e4d19a2becda559d57aa
+5. vETH2 Saddle Pool: 0xCF91812631e37C01c443a4fa02DfB59ee2DDbA7c
+
+These pools consist of total tokens stake and the rewards that are to be distributed to stakers. This is easy to navigate with LP tokens by using Etherscan to understand the pools composition and split LPs from SGT rewards.
```
Reinis-FRP, line 35: How one could split staked assets from rewards for SGT staking contract as rewards are in the same currency as staked SGT?
Is it possible that staked tokens can be transferred by mistake without calling the `stake` method? If so, how to distinguish assets that are not recoverable?

UMIPs/Add uTVL_KPI_SGT as price identifier.md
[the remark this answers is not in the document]
pemulis: Fair enough!

UMIPs/Add uTVL_KPI_SGT as price identifier
```diff
@@ -0,0 +1,137 @@
...
+5.There are a variety of price identifiers for native ERC20 tokens. Calculation of the dollar value should follow the relevant UMIP where it exists, and where there is no Price Identifier UMIP, the three highest volume [ERC20]/USD* markets should be queried, and the median value used.
+•	There is no need for price processing. This is a snapshot based on a particular time; however, it may be useful for TVL Options holders to have oversight of the ongoing TVL and consequently the value of their options on an ongoing basis. There should be a dashboard developed to track the TVL of SharedStake, which will be proposed through their governance site.
+o	The use of dashboards was considered but rejected due to the potential for manipulation and the differences between their calculations of the values of the collateral assets and the value as determined by the relevant UMIPs in previous KPI calculations.
```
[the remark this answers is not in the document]
pemulis: If you wrote a Defillama adapter for this with a `tvl` function, that adapter could be referenced, and any new contracts/pools could be added as needed to the adapter. That way, the price identifier is straightforward ("look at the adapter's `tvl` function) while the specific contracts involved are flexible.

Here's ours, for reference: https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/uma/index.js

There are lots of other adapters in there with code you could modify to your needs.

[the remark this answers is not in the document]
pemulis: I don't think you _necessarily_ need to do it this way, but you could also note the current contracts to check for TVL in the UMIP, with a note that the UMIP could be updated with a more robust TVL calculation based on an adapter function in the future.

smb2796:
@wmuffett please fix the formatting to conform to the UMIP template.

UMIPs/Add uTVL_KPI_SGT as price identifier.md
```diff
@@ -0,0 +1,146 @@
...
+
+Price Feed Implementation
+As there is no requirement for ongoing monitoring through liquidation or dispute bots, a price feed is not required. The only requirement is a query of the SGT TVL statistic at the timestamp 00.00(UTC) on 31th December according to the data and markets as defined above.
+
+
+Technical Specifications
+1. Price Identifier Name - uTVL_KPI_SGT
+2. Base Currency - uTVL_SGT.
+3. Quote currency
+If your price identifier is a currency pair, your quote currency will be the denominator of your currency pair. If your price identifier does not have a quote currency, please explain the reasoning behind this.
```
smb2796, line 89: You can remove all notes from the UMIP template.

UMIPs/Add uTVL_KPI_SGT as price identifier.md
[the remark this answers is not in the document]
chrismaree: this is true but we would need to blacklist the price from the proposer/disputer bots to be double sure.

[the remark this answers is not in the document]
smb2796: Well there wouldn't be any price feed because this is fully collateralized, so the proposer/disputer bots wouldn't actually be running for this.

UMIPs/Add uTVL_KPI_SGT as price identifier.md
```diff
@@ -0,0 +1,146 @@
...
+•	A protocol community member, token holder, voter, or proximal Defi protocol participant may be gifted a TVL option by a protocol as an incentive to build the TVL of the protocol within the option timeframe and redeem at expiry.
+•	Any user may purchase a TVL Option for a protocol that they believe has the potential for growth in TVL prior to expiry.
+3.	The current TVL of SharedStake is approximately $37m as at 4:30(UTC) 19th April 2021.
+
+
+Markets and Data Sources
+There are now five assets approved as Collateral within SharedStake’s staking contracts. These pools are single sided SGT, SGT/ETH LP, vETH2 staking, SGT/vETH2 Uniswap LP, and the Saddle vETH2 pool. To maintain consistency with existing price identifier UMIPs, it is suggested that TVL is calculated using the deepest Dex pools on Uniswap and Sushiswap.
+Additionally, there is a liquidity token that is accepted as collateral at Sharedstake.org, which the TVL can be calculated using the Uniswap liquidity / total circulating supply.
+The contract addresses for the five staking pools are as shown below:
```
chimera-defi, line 28: Since the staking pool addresses change at the end of every rewards epoch (currently 3 mos, 1mo for saddle) its best to use something else. I propose using ETH reserves in the UNI rewards pools:
https://etherscan.io/address/0x3d07f6e1627da96b8836190de64c1aed70e3fc55#readContract

We can track the total ETH in the pool as `getReserves()[ _reserve1]` and multiply by the price of ETH x 2 to get TVL. 
e.g. currently this is 329 ETH, reported as `329010667784417474724`. Therefore `329010667784417474724 / 1e18 * 3,000 * 2` returns the TVL in SGT liquidity as roughly ~$2M

UMIPs/Add uTVL_KPI_SGT as price identifier.md
chimera-defi: Please see my comment here and tag me if it doesnt answer your query   
https://github.com/UMAprotocol/UMIPs/pull/243/files#r625338091

UMIPs/Add uTVL_KPI_SGT as price identifier.md
chimera-defi: Please see my comment here and tag me if it doesnt answer your query   
https://github.com/UMAprotocol/UMIPs/pull/243/files#r625338091 

>Is it possible that staked tokens can be transferred by mistake without calling the stake method? If so, how to distinguish assets that are not recoverable?
While this is possible, and there's also the possibility of people depositing other random tokens into the pools as advertising as currently seen on some pools, I dont think this will affect the calc bottom line too much.

PR #246, opened 2021-04-14
Aragon KPI UMIP

nivida:
Because of the time pressure were I forced to fill it out as fast as possible. I'm sure it isn't correct. This is clearly not the way I would have liked to open this PR. But a misunderstanding was resulting in this situation.

### DIFF
--- a/UMIPs/V2migration_KPI_Aragon.md
+++ b/UMIPs/V2migration_KPI_Aragon.md
@@ -1,10 +1,11 @@
 # Headers
 | UMIP-79     |                                                                                                                                          |
 |------------|------------------------------------------------------------------------------------------------------------------------------------------|
-| UMIP Title | V2migration_KPI_Aragon                                                                                         |
+| UMIP Title | V2migration_KPI_Aragon                                                                                      |
 | Authors    | Joel and Sam (sam@aragon.one) |
 | Status     | Draft                                                                                                                                    |
 | Created    | April 14, 2021                                                                                                                           |
+| Forum      | https://forum.aragon.org/t/kpi-options-using-uma-protocol/2633                                                                           |
 
 Note that an UMIP number will be assigned by an editor. When opening a pull request to submit your UMIP, please use an abbreviated title in the filename, umip-draft_title_abbrev.md. The title should be 44 characters or less.
 
@@ -16,18 +17,15 @@ The purpose of this UMIP is to add support for the KPI option price feed for the
 
 Nearly all DAOs created with Aragon are currently on Aragon v1. With the launch of Aragon v2, we would like to incentivise Aragon DAOs to transition from Aragon v1 to Aragon v2. To incentivise this upgrade, we’re proposing to use KPI options as a mechanism to accelerate the transition to Aragon v2.
 
-The options would be distributed to all Aragon v1 DAOs immediately after the option creation. At the option expiry, Aragon v1 DAOs would then be able to redeem their options for the ANT collateral in the KPI option contract. The amount of collateral they can redeem will be dependent upon:
-
-- The DAO’s proportional share of AUM (USD denominated) relative to all Aragon V2 DAOs at 30th June 2021.
-- How quickly the DAO upgrades to Aragon v2 DAO relative to other v2 DAOs
+The options will be distributed to all Aragon v1 DAOs that go through the migration (the amount of options will be proportional to their assets). At the option expiry, DAOs that hold the option will be able to claim/exchange them for the ANT reward. The price for each option will be dependant on the total AUM on v2 DAOs (meaning that if more DAOs migrate, the more the options will be worth it), respecting a top threshold of 0,1 ANT per option. 
 
 # **MARKETS & DATA SOURCES**
 
-No definition of markets & data sources required because this is a UMA KPI option with the ANT token.
+No definition of markets & data sources required because this is a Aragon KPI option with the ANT token.
 
 # **PRICE FEED IMPLEMENTATION**
 
-As there is no requirement for ongoing monitoring through liquidation or dispute bots, a price feed is not required. The only requirement is a query of the UMA TVL statistic at the timestamp 00.00(UTC) on 30th June according to the data and markets as defined above.
+As there is no requirement for ongoing monitoring through liquidation or dispute bots, a price feed is not required. The only requirement is a query of the Aragon AUM statistic at the timestamp 00.00(UTC) on 30th June according to the data and markets as defined above.
 
 # **TECHNICAL SPECIFICATIONS**
 
@@ -37,8 +35,8 @@ As there is no requirement for ongoing monitoring through liquidation or dispute
 
 **3. Quote currency** - NA
 
-- There is no quote currency, the denominator is fixed
-- This price identifier does not have a quote currency as it is designed not to be tied to a currency price metric
+- There is no quote currency, the denominator is fixed to the ANT token.
+- This price identifier does not have a quote currency as it is designed not to be tied to a currency price metric.
 
 **4. Intended Collateral Currency** - ANT
 
@@ -53,23 +51,27 @@ As there is no requirement for ongoing monitoring through liquidation or dispute
 
 # **RATIONALE**
 
-This synthetic is designed as an incentivisation mechanism to push the Aragon DAO's to V2. Since this is percentage and based there will be a natural floor and ceiling for the adoption. 
+This synthetic is designed as an incentivisation mechanism to push the Aragon DAO's to V2. Since it is an incentivisation program Aragon defined a top threshold that we aim to achieve of migrated assets (100MM USD in converted value of all tokens), and this threshold correspondents to maximium payout of the options for the available reward
 
 This contract will only be called once at expiry, there will be no need to run bots, thus this price feed is only applicable at the expiry timestamp.  
 
 # **IMPLEMENTATION**
 
-- The value of an option can get determined simply by the subgraph Aragon will provide at: [https://thegraph.com/explorer/subgraph/aragon/aragon-migration](https://thegraph.com/explorer/subgraph/aragon/aragon-migration)
-- The total value locked is the value within the EMP contract of UMA.
+- The value of an option can get determined simply requesting it at the following end-point provided by the Aragon Association: [datafeed.aragon.org](https://datafeed.aragon.org)
+- The total value locked is the Aragon AUM within the EMP contract of UMA.
+
+To request our data end-point for the Aragon KPI options value can you send a simple GET request to ``datafeed.aragon.org``. 
+
 1. **What prices should be queried for and from which markets?**
     - **Note** - This should match the markets and pairs listed in the `Markets and Data Sources` section.
-    - The current price of the option is based on all funds migrated from V1 DAOs to V2 DAOs and can get requested from our subgraph at: [https://thegraph.com/explorer/subgraph/aragon/aragon-migration](https://thegraph.com/explorer/subgraph/aragon/aragon-migration)
+    - The value of an option is calculated based on all funds migrated from V1 DAOs to V2 DAOs and can get requested from our end-point at: [datafeed.aragon.org](https://datafeed.aragon.org)
 2. **Pricing interval**
     - The pricing interval can be done each day at 00:00.
 3. **Input processing**
-    - To get the latest price of the KPI options is a simple GraphQL query required. Closer details can be seen on the referenced TheGraph Explorer URL.
+    - To get the latest price of the KPI options is a simple GET request required to [datafeed.aragon.org](https://datafeed.aragon.org).
 4. **Result processing**
-    - The result doesn't need any further processing because our subgraph will directly provide the correctly calculated value of each option.
+    - The result doesn't need any further processing because our end-point will directly provide the correctly calculated value of an option.
+    - The following calculation is done: Option price (in ANT) = 100k*(total migrated asset/100MM) / 1MM 
 
 # **Security considerations**
 
--- a/UMIPs/V2migration_KPI_Aragon.md
+++ b/UMIPs/V2migration_KPI_Aragon.md
@@ -63,8 +63,8 @@ This contract will only be called once at expiry, there will be no need to run b
 To request our data end-point for the Aragon KPI options value can you send a simple GET request to ``datafeed.aragon.org``. 
 
 1. **What prices should be queried for and from which markets?**
-    - **Note** - This should match the markets and pairs listed in the `Markets and Data Sources` section.
     - The value of an option is calculated based on all funds migrated from V1 DAOs to V2 DAOs and can get requested from our end-point at: [datafeed.aragon.org](https://datafeed.aragon.org)
+    - The value of those founds are calculated based on the prices provided by [https://www.coingecko.com/](https://www.coingecko.com/) 
 2. **Pricing interval**
     - The pricing interval can be done each day at 00:00.
 3. **Input processing**
--- a/UMIPs/V2migration_KPI_Aragon.md
+++ b/UMIPs/V2migration_KPI_Aragon.md
@@ -7,8 +7,6 @@
 | Created    | April 14, 2021                                                                                                                           |
 | Forum      | https://forum.aragon.org/t/kpi-options-using-uma-protocol/2633                                                                           |
 
-Note that an UMIP number will be assigned by an editor. When opening a pull request to submit your UMIP, please use an abbreviated title in the filename, umip-draft_title_abbrev.md. The title should be 44 characters or less.
-
 # **SUMMARY**
 
 The purpose of this UMIP is to add support for the KPI option price feed for the Aragon KPI option airdrop. The purpose of this UMIP is to support the KPI option price feed for the eventual payout of the options to the recipients of the airdrop.
--- a/UMIPs/V2migration_KPI_Aragon.md
+++ b/UMIPs/V2migration_KPI_Aragon.md
@@ -2,7 +2,7 @@
 | UMIP-79     |                                                                                                                                          |
 |------------|------------------------------------------------------------------------------------------------------------------------------------------|
 | UMIP Title | Add V2migration_KPI_Aragon as a Price Identifier                                                                                      |
-| Authors    | Joel and Sam (sam@aragon.one) |
+| Authors    | Ramon (ramon@aragon.org), Sam (sam@aragon.one) |
 | Status     | Draft                                                                                                                                    |
 | Created    | April 14, 2021                                                                                                                           |
 | Forum      | https://forum.aragon.org/t/kpi-options-using-uma-protocol/2633                                                                           |
@@ -15,7 +15,9 @@ The purpose of this UMIP is to add support for the KPI option price feed for the
 
 Nearly all DAOs created with Aragon are currently on Aragon v1. With the launch of Aragon v2, we would like to incentivise Aragon DAOs to transition from Aragon v1 to Aragon v2. To incentivise this upgrade, we’re proposing to use KPI options as a mechanism to accelerate the transition to Aragon v2.
 
-The options will be distributed to all Aragon v1 DAOs that go through the migration (the amount of options will be proportional to their assets). At the option expiry, DAOs that hold the option will be able to claim/exchange them for the ANT reward. The price for each option will be dependant on the total AUM on v2 DAOs (meaning that if more DAOs migrate, the more the options will be worth it), respecting a top threshold of 0,1 ANT per option. 
+The options will be distributed to Aragon v1 DAOs that go through the migration (the amount of options will be proportional to their assets). At the option expiry, DAOs that hold the option will be able to claim/exchange them for the ANT reward. The price for each option will be dependant on the total AUM on v2 DAOs (meaning that if more DAOs migrate, the more the options will be worth it), respecting a top threshold of 0,1 ANT per option. 
+
+**IMPORTANT**: Only DAOs created before the publish of the reward program are elligible to receive the options. The publish happened on Mar 30, 2021, 0:00 PM UTC, through this Snapshot proposal - https://snapshot.org/#/aragon/proposal/QmXDBG7ZdCfg4fSRDhSwNSsdXggjsLapP9q3ijArysS56C
 
 # **MARKETS & DATA SOURCES**
 
@@ -58,8 +60,7 @@ More informatino about the program can be found on these tow proposals on snapsh
 
 # **IMPLEMENTATION**
 
-- The value of an option can get determined be performing a GET request at the following endpoint provided by the Aragon Association: [datafeed.aragon.org](https://datafeed.aragon.org)
-- The total value locked is the Aragon AUM within the EMP contract of UMA.
+The value of an option can get determined be performing a GET request at the following endpoint provided by the Aragon Association: [datafeed.aragon.org](https://datafeed.aragon.org)
 
 To request our data end-point for the Aragon KPI options value can you send a simple GET request to ``datafeed.aragon.org``.
 This request will return you the following response body:
@@ -82,10 +83,15 @@ This request will return you the following response body:
     - The following calculation is done: Option price (in ANT) = 100k*(Total assets migrated by DAOs to v2 in USD/100MM USD) / 1MM 
     - 100k - amount of ANT collateral allocated to reward program
     - 1MM - amount of options that will be issued
-    - Amount of optinos that will be redeemable by each DAO = (Amount of assets the DAO migrated in USD)/100MM USD * 1MM 
+    - Upperbound for migrated assets (meaning, upperbound the reward program will pay for) - USD 100MM
+    - Amount of options that will be redeemable by each DAO = (Amount of assets the DAO migrated in USD)/100MM USD * 1MM 
+5. **Upperbound exceeded**
+    - Since there is an upperbound of migrated assets that are eligible for this program (100MM USD), if a DAO migrates anything above it, it will not receive the full amount of options. Example: 5 DAOs already migrated 90MM of assets, which means 900M options have already being distributed. The 6th DAO then migrates another 20MM in assets - In this scenario the 6th DAO will receive only 100M options, and not 200M, since we've reached the upperbound. This program works in a first come, first served way.
 
-For the migration of funds of a V1 DAO to a V2 DAO does the user create a proposal to vote about the migration. If the proposal for the migration got accepted will it: 
-1. Call GovernBaseFactory and create a Govern + Queue pair registered by a name in the GovernRegistry.
+6, **Migration mechanics**
+The migration of funds of a V1 DAO to a V2 DAO will be done in the following way:
+1. User create a proposal on v1 DAO to vote about the migration. If the proposal for the migration is accepted it will: 
+2. Call GovernBaseFactory and create a Govern + Queue pair registered by a name in the GovernRegistry.
 3. Optionally register the DAO governance token in the L2 voting system (Aragon Voice).
 4. Transfer the funds from the V1 Vault contract to the V2 Govern contract.
 
@@ -100,8 +106,6 @@ Involved contracts:
     - Voting (``0xfcc089230e47d9376fcbd7177164c095ce8e9f23``): https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol 
 
 # **Security considerations**
-
-**Example questions**
 
 1. Where could manipulation occur?
     - There is little possibly of the DAO's manipulate the system since the metric is specifically for the migration from V1 to V2. Meaning the current state of the Aragon DAO system is finite
--- a/UMIPs/V2migration_KPI_Aragon.md
+++ b/UMIPs/V2migration_KPI_Aragon.md
@@ -88,8 +88,8 @@ This request will return you the following response body:
 5. **Upperbound exceeded**
     - Since there is an upperbound of migrated assets that are eligible for this program (100MM USD), if a DAO migrates anything above it, it will not receive the full amount of options. Example: 5 DAOs already migrated 90MM of assets, which means 900M options have already being distributed. The 6th DAO then migrates another 20MM in assets - In this scenario the 6th DAO will receive only 100M options, and not 200M, since we've reached the upperbound. This program works in a first come, first served way.
 
-6, **Migration mechanics**
-The migration of funds of a V1 DAO to a V2 DAO will be done in the following way:
+6. **Migration mechanics**
+- The migration of funds of a V1 DAO to a V2 DAO will be done in the following way:
 1. User create a proposal on v1 DAO to vote about the migration. If the proposal for the migration is accepted it will: 
 2. Call GovernBaseFactory and create a Govern + Queue pair registered by a name in the GovernRegistry.
 3. Optionally register the DAO governance token in the L2 voting system (Aragon Voice).
--- a/UMIPs/V2migration_KPI_Aragon.md
+++ b/UMIPs/V2migration_KPI_Aragon.md
@@ -60,7 +60,14 @@ This contract will only be called once at expiry, there will be no need to run b
 - The value of an option can get determined simply requesting it at the following end-point provided by the Aragon Association: [datafeed.aragon.org](https://datafeed.aragon.org)
 - The total value locked is the Aragon AUM within the EMP contract of UMA.
 
-To request our data end-point for the Aragon KPI options value can you send a simple GET request to ``datafeed.aragon.org``. 
+To request our data end-point for the Aragon KPI options value can you send a simple GET request to ``datafeed.aragon.org``.
+This request will return you the following response body:
+```
+{
+    value: <Value of a option in ANT>
+    timestamp: <Timestamp to know the exact time of the calculation in our backend>
+} 
+```
 
 1. **What prices should be queried for and from which markets?**
     - The value of an option is calculated based on all funds migrated from V1 DAOs to V2 DAOs and can get requested from our end-point at: [datafeed.aragon.org](https://datafeed.aragon.org)
@@ -72,6 +79,21 @@ To request our data end-point for the Aragon KPI options value can you send a si
 4. **Result processing**
     - The result doesn't need any further processing because our end-point will directly provide the correctly calculated value of an option.
     - The following calculation is done: Option price (in ANT) = 100k*(total migrated asset/100MM) / 1MM 
+
+For the migration of funds of a V1 DAO to a V2 DAO does the user create a proposal to vote about the migration. If the proposal for the migration got accepted will it: 
+1. Call GovernBaseFactory and create a Govern + Queue pair registered by a name in the GovernRegistry.
+3. Optionally register the DAO governance token in the L2 voting system (Aragon Voice).
+4. Transfer the funds from the V1 Vault contract to the V2 Govern contract.
+
+Involved contracts:
+- V2:
+    - GovernBaseFactory (``0xc03710063b0e4435f997A0B1bbdf2680A2f07E13``): https://github.com/aragon/govern/blob/develop/packages/govern-create/contracts/GovernBaseFactory.sol
+    - GovernRegistry (``0x7714e0a2A2DA090C2bbba9199A54B903bB83A73d``): https://github.com/aragon/govern/blob/develop/packages/govern-core/contracts/GovernRegistry.sol 
+    - Govern: https://github.com/aragon/govern/blob/develop/packages/govern-core/contracts/Govern.sol 
+    - Queue: https://github.com/aragon/govern/blob/develop/packages/govern-core/contracts/pipelines/GovernQueue.sol 
+- V1:
+    - Vault (``0xfcc089230e47d9376fcbd7177164c095ce8e9f23``): https://github.com/aragon/aragon-apps/blob/master/apps/vault/contracts/Vault.sol
+    - Voting (``0xfcc089230e47d9376fcbd7177164c095ce8e9f23``): https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol 
 
 # **Security considerations**
 
--- a/UMIPs/V2migration_KPI_Aragon.md
+++ b/UMIPs/V2migration_KPI_Aragon.md
@@ -31,7 +31,7 @@ As there is no requirement for ongoing monitoring through liquidation or dispute
 
 **1. Price Identifier Name** - V2migration_KPI_Aragon
 
-**2. Base Currency** - V2migration_Aragon
+**2. Base Currency** - V2migration_KPI_Aragon
 
 **3. Quote currency** - NA
 
--- a/UMIPs/V2migration_KPI_Aragon.md
+++ b/UMIPs/V2migration_KPI_Aragon.md
@@ -55,7 +55,7 @@ This synthetic is designed as an incentivisation mechanism to push the Aragon DA
 
 This contract will only be called once at expiry, there will be no need to run bots, thus this price feed is only applicable at the expiry timestamp.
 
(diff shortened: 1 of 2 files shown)


UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+Nearly all DAOs created with Aragon are currently on Aragon v1. With the launch of Aragon v2, we would like to incentivise Aragon DAOs to transition from Aragon v1 to Aragon v2. To incentivise this upgrade, we’re proposing to use KPI options as a mechanism to accelerate the transition to Aragon v2.
+
+The options would be distributed to all Aragon v1 DAOs immediately after the option creation. At the option expiry, Aragon v1 DAOs would then be able to redeem their options for the ANT collateral in the KPI option contract. The amount of collateral they can redeem will be dependent upon:
+
+- The DAO’s proportional share of AUM (USD denominated) relative to all Aragon V2 DAOs at 30th June 2021.
+- How quickly the DAO upgrades to Aragon v2 DAO relative to other v2 DAOs
+
+# **MARKETS & DATA SOURCES**
+
+No definition of markets & data sources required because this is a UMA KPI option with the ANT token.
```
ChandlerDeKock, line 26: Should be "Aragon KPI option" 

I would also suggest adding a description of how the subgraph you mention below would work

ChandlerDeKock (commented):
can you rename the file to: V2migration_KPI_Aragon

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
+# Headers
+| UMIP-79     |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | V2migration_KPI_Aragon                                                                                         |
+| Authors    | Joel and Sam (sam@aragon.one) |
+| Status     | Draft                                                                                                                                    |
+| Created    | April 14, 2021                                                                                                                           |
+
```
mhairimca, line 8: Discourse Link

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+# **TECHNICAL SPECIFICATIONS**
+
+**1. Price Identifier Name** - V2migration_KPI_Aragon
+
+**2. Base Currency** - V2migration_Aragon
+
+**3. Quote currency** - NA
+
+- There is no quote currency, the denominator is fixed
```
mhairimca, line 40: How is the denominator defined?

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+    - To get the latest price of the KPI options is a simple GraphQL query required. Closer details can be seen on the referenced TheGraph Explorer URL.
+4. **Result processing**
+    - The result doesn't need any further processing because our subgraph will directly provide the correctly calculated value of each option.
+
+# **Security considerations**
+
+**Example questions**
+
+1. Where could manipulation occur?
+    - There is little possibly of the DAO's manipulate the system since the metric is specifically for the migration from V1 to V2. Meaning the current state of the Aragon DAO system is finite
```
mhairimca, line 79: Can DAOs "back migrate" from V2 to V1?  (Would they want to?)

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+
+# **RATIONALE**
+
+This synthetic is designed as an incentivisation mechanism to push the Aragon DAO's to V2. Since this is percentage and based there will be a natural floor and ceiling for the adoption. 
+
+This contract will only be called once at expiry, there will be no need to run bots, thus this price feed is only applicable at the expiry timestamp.  
+
+# **IMPLEMENTATION**
+
+- The value of an option can get determined simply by the subgraph Aragon will provide at: [https://thegraph.com/explorer/subgraph/aragon/aragon-migration](https://thegraph.com/explorer/subgraph/aragon/aragon-migration)
```
mhairimca, line 62: That link isnt working for me

Option settlement seems based on two aspects - relative weight of the DAO and speed of migration. 
It seems like these will require further results processing, unless it is already defined explicitly in the subgraph.

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+
+The purpose of this UMIP is to add support for the KPI option price feed for the Aragon KPI option airdrop. The purpose of this UMIP is to support the KPI option price feed for the eventual payout of the options to the recipients of the airdrop.
+
+# **MOTIVATION**
+
+Nearly all DAOs created with Aragon are currently on Aragon v1. With the launch of Aragon v2, we would like to incentivise Aragon DAOs to transition from Aragon v1 to Aragon v2. To incentivise this upgrade, we’re proposing to use KPI options as a mechanism to accelerate the transition to Aragon v2.
+
+The options would be distributed to all Aragon v1 DAOs immediately after the option creation. At the option expiry, Aragon v1 DAOs would then be able to redeem their options for the ANT collateral in the KPI option contract. The amount of collateral they can redeem will be dependent upon:
+
+- The DAO’s proportional share of AUM (USD denominated) relative to all Aragon V2 DAOs at 30th June 2021.
```
mhairimca, line 21: AUM = Assets under management?  

How is this defined? (or have I misinterpreted)

UMIPs/umip-79.md
mhairimca, line 79: Could a DAO quickly add additional assets just prior to June 30th and remove once the options expire, giving a false impression of their weight?

githubdoramon: Not really... at least not "automatically"... they could just send assets back to a v1 DAO if they don't like the new thing. But we are assuming we are ok with this if they do

UMIPs/umip-79.md
githubdoramon: Actually, the speed of migration only matters to ensure that we still have options to distribute - it won't make a difference for the actual value of the option. Only the relative weight + total v2 assets migrated that matters

UMIPs/umip-79.md
githubdoramon: Ah.. this is actually outdate... Options will be distributed after a migration occurs, and the amount is the AUM of the DAO (assets under mgmt) / Upper threshold of migration, which is 100MM USD (defined here - https://snapshot.org/#/aragon/proposal/QmXq7KzLPQUeqxQ9cceoHwFR3oxoq7oTHHeNLfhrXNehJ9)

UMIPs/umip-79.md
githubdoramon: no... only the original migration assets are included in the option calculation. But the DAO could remove those assets after migrating. But we are "ok" if this happens, as we assume that the new solution is better and they don't want to leave

UMIPs/umip-79.md
Reinis-FRP, line 26: How one can measure Aragon AUM in USD terms without a market data source?
What type of tokens Aragon DAOs hold?

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+- Is your collateral currency already approved to be used by UMA financial contracts? If no, submit a UMIP to have the desired collateral currency approved for use.
+    - Yes
+
+**5. Collateral Decimals** - 18 decimels
+
+**6. Rounding** - 2 decimal places
+
+# **RATIONALE**
+
+This synthetic is designed as an incentivisation mechanism to push the Aragon DAO's to V2. Since this is percentage and based there will be a natural floor and ceiling for the adoption.
```
Reinis-FRP, line 56: As per @githubdoramon comment in the `MOTIVATION` section there will be USD 100m ceiling for V2 which should be explained here

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+# **RATIONALE**
+
+This synthetic is designed as an incentivisation mechanism to push the Aragon DAO's to V2. Since this is percentage and based there will be a natural floor and ceiling for the adoption. 
+
+This contract will only be called once at expiry, there will be no need to run bots, thus this price feed is only applicable at the expiry timestamp.  
+
+# **IMPLEMENTATION**
+
+- The value of an option can get determined simply by the subgraph Aragon will provide at: [https://thegraph.com/explorer/subgraph/aragon/aragon-migration](https://thegraph.com/explorer/subgraph/aragon/aragon-migration)
+- The total value locked is the value within the EMP contract of UMA.
```
Reinis-FRP, line 63: Update for Aragon AUM

UMIPs/umip-79.md
Reinis-FRP, line 62: What will be the output of Aragon migration subgraph when launched? Will it output USD value or millions of USD for AUM? Or the output will be relative AUM to $100m threshold?

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+1. **What prices should be queried for and from which markets?**
+    - **Note** - This should match the markets and pairs listed in the `Markets and Data Sources` section.
+    - The current price of the option is based on all funds migrated from V1 DAOs to V2 DAOs and can get requested from our subgraph at: [https://thegraph.com/explorer/subgraph/aragon/aragon-migration](https://thegraph.com/explorer/subgraph/aragon/aragon-migration)
+2. **Pricing interval**
```
Reinis-FRP, line 67: Can you provide `curl` example on how to query Aragon migration subpgraph for the requested timestamp parameter?

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+    - The pricing interval can be done each day at 00:00.
+3. **Input processing**
+    - To get the latest price of the KPI options is a simple GraphQL query required. Closer details can be seen on the referenced TheGraph Explorer URL.
```
Reinis-FRP, line 70: This section should discuss how voters shall arrive at correct price in case the primary method (TheGraph Explorer) is not working. How it would be possible to calculate Aragon V2 AUM independently?

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+
+# **SUMMARY**
+
+The purpose of this UMIP is to add support for the KPI option price feed for the Aragon KPI option airdrop. The purpose of this UMIP is to support the KPI option price feed for the eventual payout of the options to the recipients of the airdrop.
+
+# **MOTIVATION**
+
+Nearly all DAOs created with Aragon are currently on Aragon v1. With the launch of Aragon v2, we would like to incentivise Aragon DAOs to transition from Aragon v1 to Aragon v2. To incentivise this upgrade, we’re proposing to use KPI options as a mechanism to accelerate the transition to Aragon v2.
+
+The options would be distributed to all Aragon v1 DAOs immediately after the option creation. At the option expiry, Aragon v1 DAOs would then be able to redeem their options for the ANT collateral in the KPI option contract. The amount of collateral they can redeem will be dependent upon:
```
pemulis, line 19: You can update this section to reflect that the price feed will based on aggregate AUM of all v2 DAOs at expiry, and options will be granted when the DAO migrates.

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+
+- The DAO’s proportional share of AUM (USD denominated) relative to all Aragon V2 DAOs at 30th June 2021.
+- How quickly the DAO upgrades to Aragon v2 DAO relative to other v2 DAOs
```
pemulis, line 22: You can remove this line here since their speed of migration will not affect their option redemption.

UMIPs/umip-79.md
nivida: Thanks! Will do that :-)

UMIPs/umip-79.md
nivida: I will extend this section to clearly define that you get more ANT based on the KPI and that there is no quote currency required.

UMIPs/umip-79.md
nivida: Yes, and we have to handle this case accordingly in our DataFeed for the UMA KPI options. Because if one DAO is just migrating forth and back (V1<->V2) should they don't get any migration reward. Otherwise, the whole concept is gamable.

UMIPs/umip-79.md
nivida: > That link isnt working for me

Yes, the required subgraph is not deployed right now. Is it possible to reference the end-point in a way that we can, later on, change it? For now, we would like to do it simply with a subgraph but I and also the engineer responsible for aren't sure if we reach the limit of possibilities of a subgraph in case we find some other edge cases a DAO could game the system.

I would say we define an end-point as for example ``datafeed.aragon.org`` which provides the GraphQL API as expected and in the best case do we only have to pipe the requests through to our deployed subgraph. @facuspagnuolo wdyt?

UMIPs/umip-79.md
nivida: Does this mean the system can be gamed by definition? @githubdoramon 

I'm not really a big fan of this definition. 😅

UMIPs/umip-79.md
nivida: From my tech perspective was this not required. I thought it would be enough if the DAO knows how much ANT he does get. CC @githubdoramon

UMIPs/umip-79.md
nivida: @githubdoramon Could you provide me a better explanation here? Thanks :-)

UMIPs/umip-79.md
nivida: From my understanding of the business logic behind will our end-point provide the amount of ANT you get per option. CC  @githubdoramon

UMIPs/umip-79.md
nivida: Thanks for asking. I have in advance asked if this is required and was wondering why the answer was it isn't. Will add this asap.

UMIPs/umip-79.md
nivida: Thanks. I will add further details here asap.

UMIPs/umip-79.md
nivida: CC @githubdoramon could you rephrase this explanation? Thanks :-)

UMIPs/umip-79.md
mhairimca: Just to clarify, the only asset here that is considered is ANT 

(are there ANT variants like ANJ that might be relevant?  How would these be treated?)

smb2796: > From my tech perspective was this not required. I thought it would be enough if the DAO knows how much ANT he does get. CC @githubdoramon

UMA voters still need a method to determine what this should be. So this section should cover "where" they get the necessary information. Looks like you list the subgraph in the implementation section.

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+
+This contract will only be called once at expiry, there will be no need to run bots, thus this price feed is only applicable at the expiry timestamp.  
+
+# **IMPLEMENTATION**
+
+- The value of an option can get determined simply by the subgraph Aragon will provide at: [https://thegraph.com/explorer/subgraph/aragon/aragon-migration](https://thegraph.com/explorer/subgraph/aragon/aragon-migration)
+- The total value locked is the value within the EMP contract of UMA.
+1. **What prices should be queried for and from which markets?**
+    - **Note** - This should match the markets and pairs listed in the `Markets and Data Sources` section.
+    - The current price of the option is based on all funds migrated from V1 DAOs to V2 DAOs and can get requested from our subgraph at: [https://thegraph.com/explorer/subgraph/aragon/aragon-migration](https://thegraph.com/explorer/subgraph/aragon/aragon-migration)
```
smb2796, line 66: Is this the correct link? Might just be a problem with the graph's website, but this link isn't working for me.

Can you list the specific query and result field that voters should use here?

UMIPs/umip-79.md
githubdoramon: If they do that, they would just keep paying deployment costs for new DAOs... since we are going to look for the first migration only

UMIPs/umip-79.md
githubdoramon: If we hit the upper threshold (100MM of UAM), all the 1MM options will be issued, and their total value is 100k ANT, so it is not 1:1

- Total minted options: 1MM
- Max price for options: 100k*(total AUM/Threshold) / 1MM - 0,1ANT -> where total AUM/Threshold = 1

2 scenarios to exemplify:

Final AUM - 30MM
Options value = 100k*(30/100)/1MM = 0,03ANT


Final AUM - 90MM
Options value = 100k*(90/100)/1MM = 0,09ANT

UMIPs/umip-79.md
githubdoramon: @nivida We could create extra safety constraints (for example, you need to keep the assets for X months), but this would not be feasible in the short term. But I don't see the current scenario as gamed, because if people remove the assets from the DAO, just means they regretted their decision to migrate, which we should accept (since we believe that v2 is indeed better)

UMIPs/umip-79.md
githubdoramon: Is this more clear? @Reinis-FRP @nivida
"This synthetic is designed as an incentivisation mechanism to push the Aragon DAO's to V2. Since it is an incentivisation program Aragon defined a top threshold that we aim to achieve of migrated assets (100MM USD in converted value of all tokens), and this threshold correspondents to maximium payout of the options for the available reward"

UMIPs/umip-79.md
githubdoramon: Yes... basically the ANT value of one option

UMIPs/umip-79.md
githubdoramon: @nivida
"The options will be distributed to all Aragon v1 DAOs that go through the migration (the amount of options will be proportional to their assets). At the option expiry, DAOs that hold the option will be able to claim/exchange them for the ANT reward. The price for each option will be dependant on the total AUM on v2 DAOs (meaning that if more DAOs migrate, the more the options will be worth it), respecting a top threshold of 0,1 ANT per option"

UMIPs/umip-79.md
githubdoramon: @Reinis-FRP , @mhairimca and @smb2796 
Is the question here
"UMA voters have to know how much one ANT is worth", or
"How do we know that value, in USD, of the assets a v2 DAO holds"
?

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+2. **Pricing interval**
+    - The pricing interval can be done each day at 00:00.
+3. **Input processing**
+    - To get the latest price of the KPI options is a simple GraphQL query required. Closer details can be seen on the referenced TheGraph Explorer URL.
+4. **Result processing**
+    - The result doesn't need any further processing because our subgraph will directly provide the correctly calculated value of each option.
+
```
mhairimca, line 73: What if it goes haywire?

I guess at one level, as Aragon is providing the collateral and providing the info for the distribution of options, if it goes wrong its an Aragon risk to bear, but it seems untransparent.

UMIPs/umip-79.md
```diff
@@ -0,0 +1,87 @@
...
+# **Security considerations**
+
+**Example questions**
+
+1. Where could manipulation occur?
+    - There is little possibly of the DAO's manipulate the system since the metric is specifically for the migration from V1 to V2. Meaning the current state of the Aragon DAO system is finite
+2. How could this price ID be exploited?
+    - Little chance of an exploitation on the priceID
```
mhairimca, line 81: It seems very dependent on the subgraph.

UMIPs/umip-79.md
Reinis-FRP: YES, Thank you!

UMIPs/umip-79.md
githubdoramon: Ah... ok... @Reinis-FRP, let me know if like this it makes more sense (similar to the other section I changed below)

"
The options will be distributed to all Aragon v1 DAOs that go through the migration (the amount of options will be proportional to their assets). At the option expiry, DAOs that hold the option will be able to claim/exchange them for the ANT reward. The price for each option will be dependant on the total AUM on v2 DAOs (meaning that if more DAOs migrate, the more the options will be worth it), respecting a top threshold of 0,1 ANT per option. At expiry, if top threshold is reached, all options will have been distributed, and all locked ANT will be redeemable"
"

Reinis-FRP: Yes, this sounds better, but we would need to fix that maximum 0.1 ANT represents 100MM USD AUM threshold, e.g. (AUM in USD)/10^9 with the cap of 0.1.
Also, please note that expected payout depends on keeping 1:0.1 ratio for [synths minted]:[ANT locked], hence, it is worth mentioning it here.

UMIPs/umip-79.md
nivida: Thanks :)

UMIPs/umip-79.md
nivida: I have added the calculation we will do to the document and it will be visible here on the next commit.

UMIPs/umip-79.md
nivida: We have discarded the subgraph idea we had because it would have been a really quick and dirty solution. We will now provide a simple end-point at ``datafeed.aragon.org``. On calling this URL will you get the value of an option in ANT as reponse.

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,93 @@
...
+    - The following calculation is done: Option price (in ANT) = 100k*(Total assets migrated by DAOs to v2 in USD/100MM USD) / 1MM 
+    - 100k - amount of ANT collateral allocated to reward program
+    - 1MM - amount of options that will be issued
+    - Amount of optinos that will be redeemable by each DAO = (Amount of assets the DAO migrated in USD)/100MM USD * 1MM 
+
+# **Security considerations**
+
+**Example questions**
+
+1. Where could manipulation occur?
```
mrice32, line 84: Could I create a v1 DAO with $100MM just before expiry and migrate it, let the contract expire and withdraw my 100MM? 

I guess this speaks to a broader ambiguity around how you compute the dollar amount of migrated v1 DAOs.

- Is this only measuring the total market cap of v2 DAOs at expiry (i.e. no "migration" is necessary)?
- If a v1 DAO increases their TVL before migration, does that affect the total?
- If a v1 DAO increases their TVL after migration, but before expiration, does that affect the total?

mrice32 (commented):
Looks great -- a few questions

UMIPs/V2migration_KPI_Aragon.md
githubdoramon: Ah... well pointed... only v1 DAOs created before the first vote on Snapshot (https://snapshot.org/#/aragon/proposal/QmXDBG7ZdCfg4fSRDhSwNSsdXggjsLapP9q3ijArysS56C) are eligible!!!! Forgot to mention this on the doc.

Replying to your questions
- Is this only measuring the total market cap of v2 DAOs at expiry (i.e. no "migration" is necessary)? It is measuring the market cap OF the migrated DAOs... not new DAOs on v2 that were not on v1 for example.
- If a v1 DAO increases their TVL before migration, does that affect the total? If they increase it, and they migrate this new total, it impacts... BUT, the migration is a one time action... It is not about sending assets from one DAO to the other and random times
- If a v1 DAO increases their TVL after migration, but before expiration, does that affect the total? It does not impact at all

Basically... the migratino action will happen once for a DAO, and this is when we evaluate the mkt value of that DAO in particular

Reinis-FRP: > Ah... well pointed... only v1 DAOs created before the first vote on Snapshot (https://snapshot.org/#/aragon/proposal/QmXDBG7ZdCfg4fSRDhSwNSsdXggjsLapP9q3ijArysS56C) are eligible!!!! Forgot to mention this on the doc.

Just to clarify what you mean by eligible:
1. v1DAOs that receive KPI options, or
2. v1DAOs that are counted as part of v2 migration for USD value calculation?

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,93 @@
...
+This synthetic is designed as an incentivisation mechanism to push the Aragon DAO's to V2. Since it is an incentivisation program Aragon defined a top threshold that we aim to achieve of migrated assets (100MM USD in converted value of all tokens), and this threshold correspondents to maximium payout of the options for the available reward
+
+This contract will only be called once at expiry, there will be no need to run bots, thus this price feed is only applicable at the expiry timestamp.
+
+More informatino about the program can be found on these tow proposals on snapshot: [Proposal 1 - The reward program and ANT collateral](https://snapshot.org/#/aragon/proposal/QmXDBG7ZdCfg4fSRDhSwNSsdXggjsLapP9q3ijArysS56C) and [Proposal 2 - Upper AUM threshold for the program](https://snapshot.org/#/aragon/proposal/QmXq7KzLPQUeqxQ9cceoHwFR3oxoq7oTHHeNLfhrXNehJ9)
+
+
+# **IMPLEMENTATION**
+
+- The value of an option can get determined be performing a GET request at the following endpoint provided by the Aragon Association: [datafeed.aragon.org](https://datafeed.aragon.org)
```
Reinis-FRP, line 63: Since the API endpoint is not yet available, can you provide an expected response format here?

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,93 @@
...
+- The total value locked is the Aragon AUM within the EMP contract of UMA.
+
+1. **What prices should be queried for and from which markets?**
+    - The value of an option is calculated based on all funds migrated from V1 DAOs to V2 DAOs and can get requested from our endpoint at: [datafeed.aragon.org](https://datafeed.aragon.org)
+    - The funds that were migrated will have its USD value calculated based on the prices provided by [https://www.coingecko.com/](https://www.coingecko.com/) 
+2. **Pricing interval**
+    - The pricing interval are updated each day at 00:00.
+3. **Input processing**
+    - To get the latest price of the KPI options perform a simple GET request to [datafeed.aragon.org](https://datafeed.aragon.org).
```
Reinis-FRP, line 72: Is it considered that human intervention could be necessary in case the datafeed is not available or was expected to output different result if one were to do manual calculation of USD value migrated? In that case it would be helpful to include guidance for voters on how to verify USD value migrated manually.

UMIPs/V2migration_KPI_Aragon.md
githubdoramon: An eligible DAO (meaning it can be part of the program), is any DAO that existed prior to Mar 30, 2021, 0:00 PM UTC (So, before the first proposal was released on Snapshot). If one of those DAOs migrate to v2, it can claim the options.

If a DAO was created after that on v1, and migrate, it will not be entitled to receive the options.

UMIPs/V2migration_KPI_Aragon.md
githubdoramon: It should return real-time value of an option.

Expected response should be a number (representing the value option in ANT), for example 0.001

@nivida please confirm both information.

UMIPs/V2migration_KPI_Aragon.md
githubdoramon: @Reinis-FRP - will this be enough?

"- Migrated assets can have its value calculated manually by listing all tokens that were migrated, and fetching its prices (in USD) from coingecko at the moment of migration, multiplying tokens by prices, and summing it all"

Reinis-FRP: Great, if possible, could you also add a migration contract address and specific events that one would need to observe on-chain in order to manually verify the amount of migration?

githubdoramon: This is under development... @nivida can you also shine a light here?

pemulis:
Answering the remaining questions will help clarify manual calculation, if necessary, but this is looking good!

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,93 @@
...
+
+This contract will only be called once at expiry, there will be no need to run bots, thus this price feed is only applicable at the expiry timestamp.
+
+More informatino about the program can be found on these tow proposals on snapshot: [Proposal 1 - The reward program and ANT collateral](https://snapshot.org/#/aragon/proposal/QmXDBG7ZdCfg4fSRDhSwNSsdXggjsLapP9q3ijArysS56C) and [Proposal 2 - Upper AUM threshold for the program](https://snapshot.org/#/aragon/proposal/QmXq7KzLPQUeqxQ9cceoHwFR3oxoq7oTHHeNLfhrXNehJ9)
+
+
+# **IMPLEMENTATION**
+
+- The value of an option can get determined be performing a GET request at the following endpoint provided by the Aragon Association: [datafeed.aragon.org](https://datafeed.aragon.org)
+- The total value locked is the Aragon AUM within the EMP contract of UMA.
```
akhanubis, line 64: What does that mean? Isnt the VL locked in the EMP fixed at 100k ANT?

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,93 @@
...
+2. **Pricing interval**
+    - The pricing interval are updated each day at 00:00.
+3. **Input processing**
+    - To get the latest price of the KPI options perform a simple GET request to [datafeed.aragon.org](https://datafeed.aragon.org).
+4. **Result processing**
+    - The result doesn't need any further processing because our endpoint will directly provide the correctly calculated value of an option.
+    - The following calculation is done: Option price (in ANT) = 100k*(Total assets migrated by DAOs to v2 in USD/100MM USD) / 1MM 
+    - 100k - amount of ANT collateral allocated to reward program
+    - 1MM - amount of options that will be issued
+    - Amount of optinos that will be redeemable by each DAO = (Amount of assets the DAO migrated in USD)/100MM USD * 1MM
```
akhanubis, line 78: Can you add the upper bounds to the calculations? Otherwise numbers go above the limits.

Can you also add an example scenario with >= 2 DAOs migrating more than 100MM in total?

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,93 @@
...
+
+
+# **IMPLEMENTATION**
+
+- The value of an option can get determined be performing a GET request at the following endpoint provided by the Aragon Association: [datafeed.aragon.org](https://datafeed.aragon.org)
+- The total value locked is the Aragon AUM within the EMP contract of UMA.
+
+1. **What prices should be queried for and from which markets?**
+    - The value of an option is calculated based on all funds migrated from V1 DAOs to V2 DAOs and can get requested from our endpoint at: [datafeed.aragon.org](https://datafeed.aragon.org)
+    - The funds that were migrated will have its USD value calculated based on the prices provided by [https://www.coingecko.com/](https://www.coingecko.com/)
```
therealhexi, line 68: Is there a more specific link on Goingecko, pointing to ANT v2 at the expiration timestamp?

therealhexi (commented):
Minor suggestions

UMIPs/V2migration_KPI_Aragon.md
nivida: It will provide real-time value before expiration. I will add an example of the response body. We will return the timestamp when the value was calculated plus the value of an option as a number.

UMIPs/V2migration_KPI_Aragon.md
nivida: Can you elaborate closer on this? If I understood you correctly would I have to provide a list of coins managed by v1 DAOs?

UMIPs/V2migration_KPI_Aragon.md
nivida: Thanks, I will add the contract details.

UMIPs/V2migration_KPI_Aragon.md
githubdoramon: @therealhexi - If you mean the price of ANT token, is this one (https://www.coingecko.com/en/coins/aragon), but this is not the price that matters... From Sam's comment, it will depend on the assets the DAO holds, which can be any token, and wouldn't make sense to have a list of all existing tokens here right?

UMIPs/V2migration_KPI_Aragon.md
githubdoramon: Yep.... this line is just confusing... removing it

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,119 @@
...
+# **PRICE FEED IMPLEMENTATION**
+
+As there is no requirement for ongoing monitoring through liquidation or dispute bots, a price feed is not required. The only requirement is a query of the Aragon AUM statistic at the timestamp 00.00(UTC) on 30th June according to the data and markets as defined above.
+
+# **TECHNICAL SPECIFICATIONS**
+
+**1. Price Identifier Name** - V2migration_KPI_Aragon
+
+**2. Base Currency** - V2migration_Aragon
+
```
mhairimca, line 35: should be V2migration_KPI_Aragon?

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,119 @@
...
+4. **Result processing**
+    - The result doesn't need any further processing because our endpoint will directly provide the correctly calculated value of an option.
+    - The following calculation is done: Option price (in ANT) = 100k*(Total assets migrated by DAOs to v2 in USD/100MM USD) / 1MM 
+    - 100k - amount of ANT collateral allocated to reward program
+    - 1MM - amount of options that will be issued
+    - Upperbound for migrated assets (meaning, upperbound the reward program will pay for) - USD 100MM
+    - Amount of options that will be redeemable by each DAO = (Amount of assets the DAO migrated in USD)/100MM USD * 1MM 
+5. **Upperbound exceeded**
+    - Since there is an upperbound of migrated assets that are eligible for this program (100MM USD), if a DAO migrates anything above it, it will not receive the full amount of options. Example: 5 DAOs already migrated 90MM of assets, which means 900M options have already being distributed. The 6th DAO then migrates another 20MM in assets - In this scenario the 6th DAO will receive only 100M options, and not 200M, since we've reached the upperbound. This program works in a first come, first served way.
+
```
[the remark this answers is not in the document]
githubdoramon: @mhairimca this is not implemented per se... we will issue 1MM options.... and distribute them airdropping to the DAOs when they claim the option. All options will be in an Aragon wallet, and we will distribute them until they are over.

Do this answer your question?

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,119 @@
...
+This synthetic is designed as an incentivisation mechanism to push the Aragon DAO's to V2. Since it is an incentivisation program Aragon defined a top threshold that we aim to achieve of migrated assets (100MM USD in converted value of all tokens), and this threshold correspondents to maximium payout of the options for the available reward
+
+This contract will only be called once at expiry, there will be no need to run bots, thus this price feed is only applicable at the expiry timestamp.
+
+More informatino about the program can be found on these tow proposals on snapshot: [Proposal 1 - The reward program and ANT collateral](https://snapshot.org/#/aragon/proposal/QmXDBG7ZdCfg4fSRDhSwNSsdXggjsLapP9q3ijArysS56C) and [Proposal 2 - Upper AUM threshold for the program](https://snapshot.org/#/aragon/proposal/QmXq7KzLPQUeqxQ9cceoHwFR3oxoq7oTHHeNLfhrXNehJ9)
+
+
+# **IMPLEMENTATION**
+
+The value of an option can get determined be performing a GET request at the following endpoint provided by the Aragon Association: [datafeed.aragon.org](https://datafeed.aragon.org)
```
smb2796, line 63: Is this WIP? The link is down for me

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,119 @@
...
+
+**6. Rounding** - 2 decimal places
+
+# **RATIONALE**
+
+This synthetic is designed as an incentivisation mechanism to push the Aragon DAO's to V2. Since it is an incentivisation program Aragon defined a top threshold that we aim to achieve of migrated assets (100MM USD in converted value of all tokens), and this threshold correspondents to maximium payout of the options for the available reward
+
+This contract will only be called once at expiry, there will be no need to run bots, thus this price feed is only applicable at the expiry timestamp.
+
+More informatino about the program can be found on these tow proposals on snapshot: [Proposal 1 - The reward program and ANT collateral](https://snapshot.org/#/aragon/proposal/QmXDBG7ZdCfg4fSRDhSwNSsdXggjsLapP9q3ijArysS56C) and [Proposal 2 - Upper AUM threshold for the program](https://snapshot.org/#/aragon/proposal/QmXq7KzLPQUeqxQ9cceoHwFR3oxoq7oTHHeNLfhrXNehJ9)
```
smb2796, line 58:
```suggestion
More information about the program can be found on these two proposals on snapshot: [Proposal 1 - The reward program and ANT collateral](https://snapshot.org/#/aragon/proposal/QmXDBG7ZdCfg4fSRDhSwNSsdXggjsLapP9q3ijArysS56C) and [Proposal 2 - Upper AUM threshold for the program](https://snapshot.org/#/aragon/proposal/QmXq7KzLPQUeqxQ9cceoHwFR3oxoq7oTHHeNLfhrXNehJ9)
```

UMIPs/V2migration_KPI_Aragon.md
smb2796: Just to clarify - would the ineligible DAO's migrated assets still count towards `Total migrated assets` in the price calculation?

UMIPs/V2migration_KPI_Aragon.md
```diff
@@ -0,0 +1,119 @@
...
+4. **Result processing**
+    - The result doesn't need any further processing because our endpoint will directly provide the correctly calculated value of an option.
+    - The following calculation is done: Option price (in ANT) = 100k*(Total assets migrated by DAOs to v2 in USD/100MM USD) / 1MM 
+    - 100k - amount of ANT collateral allocated to reward program
+    - 1MM - amount of options that will be issued
+    - Upperbound for migrated assets (meaning, upperbound the reward program will pay for) - USD 100MM
+    - Amount of options that will be redeemable by each DAO = (Amount of assets the DAO migrated in USD)/100MM USD * 1MM 
+5. **Upperbound exceeded**
+    - Since there is an upperbound of migrated assets that are eligible for this program (100MM USD), if a DAO migrates anything above it, it will not receive the full amount of options. Example: 5 DAOs already migrated 90MM of assets, which means 900M options have already being distributed. The 6th DAO then migrates another 20MM in assets - In this scenario the 6th DAO will receive only 100M options, and not 200M, since we've reached the upperbound. This program works in a first come, first served way.
```
smb2796, line 89: Is this cap on a per DAO basis or a cap on the total migrated assets? 

Also if I'm understanding this correctly, a cap of 100M would imply a price ceiling of 1 ANT. Is this correct?

UMIPs/V2migration_KPI_Aragon.md
githubdoramon: No @smb2796 ... only the eligible ones

UMIPs/V2migration_KPI_Aragon.md
githubdoramon: Yes @smb2796 ... the endpoint is not yet available

UMIPs/V2migration_KPI_Aragon.md
githubdoramon: @smb2796 The CAP is for total migrated assets... And the cap is actually 0.1 ANT, since the collateral for the reward is 100k ANT, for 1MM options.

UMIPs/V2migration_KPI_Aragon.md
githubdoramon: Updated

smb2796 (approved):
Thanks for making the updates to clarify the calculation methodology @githubdoramon! Since this was the main blocker identified on the community call, I'm going to merge in and move this UMIP to `Last Call` status.

PR #247, opened 2021-04-15
Create uStonks_JUN21 as a price identifier

rossgalloway:


rossgalloway:
I still need to update the google doc spreadsheet that is linked in the UMIP. The file is correct but it's just a copy of the doc from the last UMIP. I will try to get to that tomorrow.

UMIPs/umip_add_ustonks_jun21_price_ident.md
```diff
@@ -0,0 +1,130 @@
...
+| Created    | April 14, 2021   
+| Discourse Link | https://discourse.umaproject.org/t/add-ustonks-jun21-price-identifier/878 |  
+
+## SUMMARY
+
+The DVM should support price requests for a basket of stocks that represent the top ten tickers with the most comment volume on r/wallstreetbets. These ten stocks will be equally weighted to create an index named uSTONKS. 
+
+## MOTIVATION
+
+The DVM currently does not support a current price identifier for the creation of indices like uSTONKS.  Financial markets should be universally accessible — unrestrained by censorship, pre-existing financial and social capital, or nationality.  With the uSTONKS price identifier synthetic tokens can be created which will allow anybody to obtain exposure to the basket of stocks in a decentralized way.  This will allow unrestricted trading 24 hours a day, 7 days a week.
```
Reinis-FRP, line 16: `uSTONKS_APR21` is already supported in UMIP-47

UMIPs/umip_add_ustonks_jun21_price_ident.md
```diff
@@ -0,0 +1,130 @@
...
+| SQ | 258.40 |
+| MVIS | 12.34 |
+| RBLX | 75.35 |
+| MARA | 42.97 |
+
+Each stock also is assigned an index value base. The index is equally weighted, so each stock is assigned a base of 10. This means that on 04/14/2021 at close, the uSTONKS index would have been worth 100 with every component making up 1/10th of that index.
+ 
+To calculate the uSTONKS price, an UMA voter should:
+1. Query for the close price of one component on June 30, 2021. It is recommended that voters use the Google Sheets GOOGLEFINANCE function.
+2. Divide the June 31 price by the base price and multiply by the index value base (10) to get today’s index value.
```
Reinis-FRP, line 99: June 30

UMIPs/umip_add_ustonks_jun21_price_ident.md
```diff
@@ -0,0 +1,132 @@
...
+
+Each stock also is assigned an index value base. The index is equally weighted, so each stock is assigned a base of 10. This means that on 04/14/2021 at close, the uSTONKS index would have been worth 100 with every component making up 1/10th of that index.
+ 
+To calculate the uSTONKS price, an UMA voter should:
+1. Query for the close price of one component on June 30, 2021. It is recommended that voters use the Google Sheets GOOGLEFINANCE function.
+2. Divide the June 30 price by the base price and multiply by the index value base (10) to get today’s index value. 
+3. Perform this function for each component in the index and sum all of the results together.
+4. Round the result of step three to six decimal places
+
+An example of this calculation is shown in [this](https://docs.google.com/spreadsheets/d/1kccQQJ8FqSnaDbeK4ljXERIg9lKkErHXtvGSbPcDTAw/edit?usp=sharing) Google Sheet.
```
smb2796, line 105: Looks like this sheet needs to be updated for the new basket.

rossgalloway: updated!

smb2796: Thanks! Looks like there are some reference issues in the sheet. Can you fix these?

rossgalloway: The REF! issues comes for values on dates that are in the future.  Is there a good excel command for those to return something else (like "-") if the value doesn't exist? Or should I just remove the rows and assume people using it can populate more rows when they want?

smb2796: Oh I see. No I don't think the rows should be removed. It makes sense, but it does look like a couple of days are missing.

rossgalloway: I think the days that are missing are weekends. As far as I can tell from the function, it reads the data from googlefinance() for the most recent day that is included in their data. They don't populate days the market is closed, so the spreadsheet skips those days. The next day that populates will be today once the market closes and data is in. I checked this by changing the index date to a week earlier and it skips the weekends but continues again with weekdays.

mhairimca (approved):

PR #249, opened 2021-04-16
Add uCRSPTMT_SEP21 price identifier

javipus:
The DVM should support price requests for the [Vanguard Total Stock Market ETF](https://investor.vanguard.com/etf/profile/overview/vti) (VTI), a highly diversified exchange-traded fund that aims to track the US stock market as a whole.

### DIFF
--- a/UMIPs/umip-draft_vti_priceID.md
+++ b/UMIPs/umip-draft_vti_priceID.md
@@ -1,0 +1,109 @@
+## HEADERS
+| UMIP [#]     |                                                                                                                                  |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add uVTI_MAY2021 as a price identifier                                                                                                  |
+| Authors    | Javier Prieto, [@javipus](github.com/javipus)
+| Status      | Draft                                                                                                                                   |
+| Created    | Mar 23, 2021 
+| Link to Discourse    |                                                                               
+
+<br>
+
+# SUMMARY 
+
+The DVM should support price requests for the [Vanguard Total Stock Market ETF](https://investor.vanguard.com/etf/profile/overview/vti) (VTI), a highly diversified exchange-traded fund that aims to track the US stock market as a whole.
+
+<br>
+
+# MOTIVATION
+
+UMA already counts with many price identifiers enabling the creation of derivatives on crypto assets and prices, as well as some identifiers for traditional, off-chain financial products like gold or stocks. However, access to highly diversified indices tracking large sectors of the economy are sorely lacking. Having exposure to index funds could prove beneficial to people outside the US, as these products provide excellent risk-adjusted returns. In particular, they provide a relatively low risk strategy to beat USD inflation.
+
+<br>
+
+# MARKETS & DATA SOURCES
+
+The contract will expire at 16:00 EST on 2021/05/21. The corresponding POSIX timestamp is `1621627200`. To determine the price, the DVM will choose between the two protocols described below depending on the timestamp of the oracle request, `request_timestamp`.
+
+- If `request_timestamp > 1621627200`, use the "After expiry" protocol.
+- Otherwise, use the "Before expiry" protocol.
+
+## After expiry
+
+After expiry, the contract should be settled by querying the expiry date (2021/05/21) close price from the NYSE, the only exchange where VTI shares are traded. The price will be quoted in USD. The price will be the arithmetic mean of the prices returned by the Google and Yahoo APIs. Here are example queries using the [yfinance](https://github.com/ranaroussi/yfinance) package and the Google Sheets function `GOOGLEFINANCE`:
+
+```python
+from yfinance import Ticker
+from pandas import to_datetime
+vti = Ticker("VTI")
+price = vti.history().loc[to_datetime("2021/05/21"), "Close"]
+```
+
+```
+=GOOGLEFINANCE("VTI", "close", DATE(2021, 5, 21))
+```
+
+Both queries can be done without need for API keys and at no cost to the user, as they are below the free tier limit of both services. Although querying the APIs directly is preferred, non-technical users can access the price using web interfaces provided by [Yahoo Finance](https://finance.yahoo.com/quote/VTI) and [Google Search](https://www.google.com/search?hl=en&q=%24vti).
+
+## Before expiry
+
+Before expiry, the price will be determined by the on-chain secondary market. At any time, the price will be the 2h Time-Weighted Average Price (TWAP) of the Uniswap uVTI_21MAY/USDC pool with the largest volume.
+
+<br>
+
+# PRICE FEED IMPLEMENTATION
+
+The liquidator and disputer bots should use the reference [Uniswap price feed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js) implementation.
+
+<br>
+
+# TECHNICAL SPECIFICATIONS
+
+**1. Price Identifier Name** - uVTI_MAY21
+
+**2. Base Currency** - uVTI_MAY21
+
+**3. Quote currency** - USDC // TODO: USD oracle or divide by USD/USDC price
+
+**4. Intended Collateral Currency** - USDC
+
+**5. Collateral Decimals** - 6
+
+**6. Rounding** - Down
+
+<br>
+
+# RATIONALE
+
+A self-referential price before expiry avoids price drift when NYSE is not trading (nights and weekends). This requires token sponsors to set initial price; it is recommended they consult the Yahoo/Google APIs as they may be arbitraged otherwise.
+
+Short-lived price manipulation of an AMM is relatively cheap, but becomes expensive over time as the position can be arbitraged. This motivates the introduction of TWAPs, in line with Uniswap's recommendation for building price oracles.
+
+<br>
+
+# IMPLEMENTATION
+
+1. **What prices should be queried for and from which markets?**
+
+  - After expiry: VTI close price on 2021/05/21 from Yahoo Finance and Google Finance
+  - Before expiry: Uniswap 2h TWAP
+
+2. **Pricing interval**
+
+  - After expiry: NA
+  - Before expiry: Every block (~15 seconds)
+
+3. **Processing**
+  
+  - After expiry: `mean(yahoo_vti_price, google_vti_price)`
+  - Before expiry: NA
+
+<br>
+
+# Security considerations
+
+1. Where could manipulation occur?
+If trading is not very liquid, the price could be temporarily manipulated before expiry by executing large trades on Uniswap. However, we note that sustaining the attack for the whole duration of a 2h TWAP would be too costly and probably not worth it.
+
+2. What are current or future concern possibilities with the way the price identifier is defined?
+After expiry, the price ID relies on only two data sources. If both were down for whatever reason, this could delay contract settlement.
--- a/UMIPs/umip-draft_vti_priceID.md
+++ b/UMIPs/umip-draft_vti_priceID.md
@@ -64,7 +64,7 @@ The liquidator and disputer bots should use the reference [Uniswap price feed](h
 
 **5. Collateral Decimals** - 6
 
-**6. Rounding** - Down
+**6. Rounding** - The price shall be rounded to 6 decimal places using [round half down](https://en.wikipedia.org/wiki/Rounding#Round_half_down).
 
 <br>
 
--- a/UMIPs/umip-draft_vti_priceID.md
+++ b/UMIPs/umip-draft_vti_priceID.md
@@ -1,53 +1,48 @@
 ## HEADERS
 | UMIP [#]     |                                                                                                                                  |
 |------------|------------------------------------------------------------------------------------------------------------------------------------------|
-| UMIP Title | Add uVTI_MAY2021 as a price identifier                                                                                                  |
+| UMIP Title | Add uCRSPTMT_SEP21 as a price identifier                                                                                                  |
 | Authors    | Javier Prieto, [@javipus](github.com/javipus)
-| Status      | Draft                                                                                                                                   |
-| Created    | Mar 23, 2021 
-| Link to Discourse    |                                                                               
+| Status      | Draft                                                                                                                             |
+| Created    | Mar 23, 2021
+| [Link to Discourse](https://discourse.umaproject.org/t/add-vti-may21-price-identifier/895)    |                                                                              
 
 <br>
 
 # SUMMARY 
 
-The DVM should support price requests for the [Vanguard Total Stock Market ETF](https://investor.vanguard.com/etf/profile/overview/vti) (VTI), a highly diversified exchange-traded fund that aims to track the US stock market as a whole.
+The DVM should support price requests for the [CRSP US Total Market TR Index](http://www.crsp.org/products/investment-products/crsp-us-total-market-index) (CRSPTMT), a highly diversified index that aims to track the US stock market as a whole.
 
 <br>
 
 # MOTIVATION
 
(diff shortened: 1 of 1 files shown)


Hahz48 (dismissed):

UMIPs/umip-draft_vti_priceID.md
```diff
@@ -0,0 +1,109 @@
...
+| Status      | Draft                                                                                                                                   |
+| Created    | Mar 23, 2021 
+| Link to Discourse    |                                                                               
+
+<br>
+
+# SUMMARY 
+
+The DVM should support price requests for the [Vanguard Total Stock Market ETF](https://investor.vanguard.com/etf/profile/overview/vti) (VTI), a highly diversified exchange-traded fund that aims to track the US stock market as a whole.
+
```
mhairimca, line 15: Would it perhaps be better to track the underlying index (MSCI US Broad Market) rather than the ETF itself.

UMIPs/umip-draft_vti_priceID.md
```diff
@@ -0,0 +1,109 @@
...
+# MOTIVATION
+
+UMA already counts with many price identifiers enabling the creation of derivatives on crypto assets and prices, as well as some identifiers for traditional, off-chain financial products like gold or stocks. However, access to highly diversified indices tracking large sectors of the economy are sorely lacking. Having exposure to index funds could prove beneficial to people outside the US, as these products provide excellent risk-adjusted returns. In particular, they provide a relatively low risk strategy to beat USD inflation.
+
+<br>
+
+# MARKETS & DATA SOURCES
+
+The contract will expire at 16:00 EST on 2021/05/21. The corresponding POSIX timestamp is `1621627200`. To determine the price, the DVM will choose between the two protocols described below depending on the timestamp of the oracle request, `request_timestamp`.
+
```
mhairimca, line 27: This timeline is quite short.

UMIPs/umip-draft_vti_priceID.md
```diff
@@ -0,0 +1,109 @@
...
+**2. Base Currency** - uVTI_MAY21
+
+**3. Quote currency** - USDC
+
+**4. Intended Collateral Currency** - USDC
+
+**5. Collateral Decimals** - 6
+
+**6. Rounding** - Down
```
Reinis-FRP, line 72: How many decimals should be left after rounding?

UMIPs/umip-draft_vti_priceID.md
javipus: I agree. We could push it to June 21.

UMIPs/umip-draft_vti_priceID.md
javipus: Is there a ticker for that?

UMIPs/umip-draft_vti_priceID.md
javipus: 6 decimals, since that's the precision of USDC. Do you think I should expand on this and provide an example?

Reinis-FRP: I believe it is sufficient just to state that resolved price is rounded 6 decimals

pemulis (commented):
Sorry this very small comment was pending since last week, haha. I think this is a cool price identifier.

Hahz48:
Lol sorry about thatSent from my Galaxy
-------- Original message --------From: John Shutt ***@***.***> Date: 4/23/21  3:57 PM  (GMT-06:00) To: UMAprotocol/UMIPs ***@***.***> Cc: Hahz48 ***@***.***>, Comment ***@***.***> Subject: Re: [UMAprotocol/UMIPs] Add VTI_MAY21 price identifier (#249) 
@pemulis commented on this pull request.

Sorry this very small comment was pending since last week, haha. I think this is a cool price identifier.

In UMIPs/umip-draft_vti_priceID.md:
> @@ -0,0 +1,109 @@
+## HEADERS
+| UMIP [#]     |                                                                                                                                  |
+| UMIP Title | Add uVTI_MAY2021 as a price identifier                                                                                                  |
+| Authors    | Javier Prieto, ***@***.***(github.com/javipus)
+| Status      | Draft                                                                                                                                   |
+| Created    | Mar 23, 2021 
+| Link to Discourse    |                                                                               
+
+<br>
+
+# SUMMARY 
+
+The DVM should support price requests for the [Vanguard Total Stock Market ETF](https://investor.vanguard.com/etf/profile/overview/vti) (VTI), a highly diversified exchange-traded fund that aims to track the US stock market as a whole.

Very cool!

UMIPs/umip-draft_vti_priceID.md
javipus:
```suggestion
**6. Rounding** - The price shall be rounded to 6 decimal places using [round half down](https://en.wikipedia.org/wiki/Rounding#Round_half_down).
```

UMIPs/umip-draft_vti_priceID.md
```diff
@@ -0,0 +1,109 @@
...
+# MARKETS & DATA SOURCES
+
+The contract will expire at 16:00 EST on 2021/05/21. The corresponding POSIX timestamp is `1621627200`. To determine the price, the DVM will choose between the two protocols described below depending on the timestamp of the oracle request, `request_timestamp`.
+
+- If `request_timestamp > 1621627200`, use the "After expiry" protocol.
+- Otherwise, use the "Before expiry" protocol.
+
+## After expiry
+
+After expiry, the contract should be settled by querying the expiry date (2021/05/21) close price from the NYSE, the only exchange where VTI shares are traded. The price will be quoted in USD. The price will be the arithmetic mean of the prices returned by the Google and Yahoo APIs. Here are example queries using the [yfinance](https://github.com/ranaroussi/yfinance) package and the Google Sheets function `GOOGLEFINANCE`:
```
therealhexi, line 33: Curious: Does NYSE provide any type of programmatic price feed?

UMIPs/umip-draft_vti_priceID.md
```diff
@@ -0,0 +1,109 @@
...
+
+<br>
+
+# Security considerations
+
+1. Where could manipulation occur?
+If trading is not very liquid, the price could be temporarily manipulated before expiry by executing large trades on Uniswap. However, we note that sustaining the attack for the whole duration of a 2h TWAP would be too costly and probably not worth it.
+
+2. What are current or future concern possibilities with the way the price identifier is defined?
+After expiry, the price ID relies on only two data sources. If both were down for whatever reason, this could delay contract settlement.
```
therealhexi, line 109: What's the dispute liveliness period? You can call out that period is historically much longer than the downtime of either Yahoo or Google

therealhexi (commented):
Very well written draft. General comments - I'm learning myself.

UMIPs/umip-draft_vti_priceID.md
javipus: Apparently they do but the pricing starts at $7,500 if I'm reading [this pdf](https://www.nyse.com/publicdocs/nyse/data/NYSE_Market_Data_Pricing.pdf) correclty.

UMIPs/umip-draft_vti_priceID.md
javipus: That's a good point. Votes need to be committed in 24h. If both APIs happen to be down simultaneously for that long, maybe not being able to settle the contract will be the least of our problems.

UMIPs/umip-draft_vti_priceID.md
```diff
@@ -0,0 +1,109 @@
...
+
+UMA already counts with many price identifiers enabling the creation of derivatives on crypto assets and prices, as well as some identifiers for traditional, off-chain financial products like gold or stocks. However, access to highly diversified indices tracking large sectors of the economy are sorely lacking. Having exposure to index funds could prove beneficial to people outside the US, as these products provide excellent risk-adjusted returns. In particular, they provide a relatively low risk strategy to beat USD inflation.
+
+<br>
+
+# MARKETS & DATA SOURCES
+
+The contract will expire at 16:00 EST on 2021/05/21. The corresponding POSIX timestamp is `1621627200`. To determine the price, the DVM will choose between the two protocols described below depending on the timestamp of the oracle request, `request_timestamp`.
+
+- If `request_timestamp > 1621627200`, use the "After expiry" protocol.
```
smb2796, line 28: Should this be `>=`?

UMIPs/umip-draft_vti_priceID.md
smb2796: There are other cheaper providers so I don't think there is any need to use a NYSE api directly

UMIPs/umip-draft_vti_priceID.md
```diff
@@ -0,0 +1,109 @@
...
+
+# RATIONALE
+
+A self-referential price before expiry avoids price drift when NYSE is not trading (nights and weekends). This requires token sponsors to set initial price; it is recommended they consult the Yahoo/Google APIs as they may be arbitraged otherwise.
+
+Short-lived price manipulation of an AMM is relatively cheap, but becomes expensive over time as the position can be arbitraged. This motivates the introduction of TWAPs, in line with Uniswap's recommendation for building price oracles.
+
+<br>
+
+# IMPLEMENTATION
```
smb2796, line 84: This section should be formatted as a step by step guide that voters should follow to arrive at the correct price. Can you explain it in this way?

smb2796:
@javipus thoughts on separating this out into two price identifiers? One would just be the VTI price, and the other would be the self-referential synth price. You could then use [this](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol) financial product library to switch the identifiers at expiry.

This is the pattern that we are trying to follow for self-referential price identifiers moving forward, and gives you a few advantages.

1. Means that you wouldn't need to propose another VTI price identifier moving forward.
2. Allows for proposal bots to actually propose the correct expiry price.

This would need a new price feed to get the VTI price, but is the best way to structure this imo.

UMIPs/umip-draft_vti_priceID.md
javipus: I don't think it makes much of a difference, but I don't mind changing it

UMIPs/umip-draft_vti_priceID.md
javipus: I assume you're referring to the Implementation section, correct?

javipus:
@smb2796 that makes a lot of sense. I was mostly pattern-matching to prior UMIPs like [uStonks](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-47.md), but I agree that separating price feed from synth implementation is cleaner.

Should I just rewrite this PR to turn it into a VTI price feed whitelisting UMIP or should I submit a new one?

smb2796:
> @smb2796 that makes a lot of sense. I was mostly pattern-matching to prior UMIPs like [uStonks](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-47.md), but I agree that separating price feed from synth implementation is cleaner.
> 
> Should I just rewrite this PR to turn it into a VTI price feed whitelisting UMIP or should I submit a new one?

Yep makes sense! uSTONKS was just originally created before the FPL functionality was available.

You could rewrite the UMIP to have both price identifiers. One for VTI (or the underlying index) and the other for the self-referential token price. 

Also I agree with the point @mhairimca brought up. I think in a lot of ways it would be cleaner to track the underlying index (CRSP US Total Market Index I believe for VTI) rather than VTI itself. Getting data for this might be slightly more challenging, but I think we should look into it. Thoughts?

UMIPs/umip-draft_vti_priceID.md
```diff
@@ -0,0 +1,109 @@
+## HEADERS
+| UMIP [#]     |                                                                                                                                  |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add uVTI_MAY2021 as a price identifier                                                                                                  |
+| Authors    | Javier Prieto, [@javipus](github.com/javipus)
+| Status      | Draft                                                                                                                                   |
+| Created    | Mar 23, 2021 
+| Link to Discourse    |                                                                               
+
```
mhairimca, line 9: link to discourse

UMIPs/umip-draft_vti_priceID.md
akhanubis: Yes, that's two weeks away and the price ID hasnt been approved yet. It should be set to a later date

UMIPs/umip-draft_vti_priceID.md
```diff
@@ -0,0 +1,109 @@
...
+## Before expiry
+
+Before expiry, the price will be determined by the on-chain secondary market. At any time, the price will be the 2h Time-Weighted Average Price (TWAP) of the Uniswap uVTI_21MAY/USDC pool with the largest volume.
+
+<br>
+
+# PRICE FEED IMPLEMENTATION
+
+The liquidator and disputer bots should use the reference [Uniswap price feed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js) implementation.
```
akhanubis, line 56: Can you include the config for the UniswapPricefeed (with a placeholder address)?

UMIPs/umip-draft_vti_priceID.md
akhanubis: Yes, for both the after and before expiry

claytonroche:
Why only the "MAY" one? It's not possible to have one that works forever?

mhairimca (commented):
Still waiting on design changes

javipus:
@claytonroche we can't make a self-referential contract perpetual because it would have no peg to the off-chain asset, and we need to make this contract self-referential because we want to use CRSPTMT instead of VTI and there are no good APIs the bots can read that price from.

javipus:
I think the only thing left is a step by step guide to calculate price before expiry. I would be very grateful if anyone could suggest UMIPs I could use as template.

smb2796:
> I think the only thing left is a step by step guide to calculate price before expiry. I would be very grateful if anyone could suggest UMIPs I could use as template.

The uSTONKS umip should be a good template to follow: https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-47.md#implementation

UMIPs/umip-draft_vti_priceID.md
```diff
@@ -0,0 +1,110 @@
...
+
+The [Center for Research in Security Prices](http://www.crsp.org/products/documentation/crsp-indexes) (CRSP) is a data provider affiliated to the University of Chicago that maintains several such indices. Of all of them, the [US Total Market Total Return Index](http://www.crsp.org/products/investment-products/crsp-us-total-market-index) (CRSPTMT), is the most comprehensive and realistic metric of total US stock market performance, as it includes nearly all publicly traded US companies and it accounts for dividends.
+
+<br>
+
+# MARKETS & DATA SOURCES
+
+The contract will expire at 16:00 EST on 2021/09/21. The corresponding POSIX timestamp is `1632240000`. To determine the price, the DVM will choose between the two protocols described below depending on the timestamp of the oracle request, `request_timestamp`.
```
smb2796, line 28: Wonder if it would make the most sense to have this timestamp be after market hours?

javipus: We can push it to 17h on expiry day if you're worried about overlap between the two time periods

smb2796: Yeah that would be my concern. Think it is easier to calculate if we are always querying for a market close price, since those are widely communicated and easily verifiable. Mind updating?

smb2796 (approved):
So a few updates that were discussed on the UMIP community call and in the comments that still need to make their way into this UMIP.

1. Adjusting timestamp to not fall in market hours
2. Adding implementation section steps
3. Adding a placeholder Uniswap price feed configuration

@javipus I'm going to merge this in, add these corrections in on my side and tag you in the PR. Please let me know if you have any issues with the updates.

UMIPs/umip-draft_vti_priceID.md
smb2796: Sorry this was a miscommunication on my side. I see that you listed 16:00 EST, but 1632240000 is actually 12:00 EST. This is where I got confused.

PR #250, opened 2021-04-18
Add punk-basic as collateral

gitofdeepanshu:


UMIPs/punk-basic-collateral.md
```diff
@@ -0,0 +1,40 @@
...
+| UMIP Title | Add PUNK-BASIC as approved collateral currency          |
+| Authors    | John Shutt (john@umaproject.org), Deepanshu Hooda (deepanshuhooda2000@gmail.com) |
+| Status     | Draft                                                                                                                                    |
+| Created    | April 16, 2021                                                                                                                           |
+| [Discourse Link](https://discourse.umaproject.org/t/add-punk-basic-as-collateral/912)    |       
+# Summary (2-5 sentences)
+This UMIP will add PUNK-BASIC as approved collateral currency. This will involve adding PUNK-BASIC to the whitelist and adding flat final fees to charge per-request.
+# Motivation
+
+Punk-Basic is a single fund token, which is backed 1:1 by any CryptoPunks NFT. Because this fund allows any CryptoPunk to be deposited as collateral, it acts as a floor price tracker for Ethereum’s original digital collectibles.
```
Reinis-FRP, line 15: Do I understand correctly that one can invest / redeem one's CryptoPunks NFT in exchange for PUNK-BASIC? If NFTs each have differing market value based on unique features, then each individual minting/redeeming should affect underlying value of fund token, so in theory market price of PUNK-BASIC should sharply react whenever someone mints or burns NFT that significantly differs from average pooled NFT value.
Is this observed in practice and what kind of volatility it implies for the collateral token?

gitofdeepanshu: Punk-Basic is a class of NFT all of which have same value. The reason for same value is that when a NFT holders redeem their NFT, they get a random NFT from the pool which makes all the NFTs equivalent.
Hope it answers your question.
There are only 104 of these out of which 100 have been minted for PUNK-BASIC token.

Reinis-FRP: Oh, interesting! Just looking at mint/redeem transactions I see that NFTs are deposited/redeemed from [this address](https://etherscan.io/token/0xb7f7f6c52f2e2fdb1963eab30438024864c313f6?a=0xaf93fcce0548d3124a5fc3045adaf1dde4e8bf7e), but I see now 158 WPUNKS held there. Does this contract hold WPUNKS also for other NFT funds and how one could verify what is the actual amount and value of NFTs backing PUNK-BASIC token? I believe this information might be important since the actual PUNK-BASIC trading volume is not that high and in case there is suspicion of PUNK-BASIC token price manipulation, then one could fall-back to checking the underlying NFT value.

pemulis: I think that WPUNKS are used for other CryptoPunk funds, as well. For instance, $PUNK-FEMALE, $PUNK-ATTR-4, etc. $PUNK-BASIC accepts any CryptoPunk as a deposit, but the other specialized funds have specific requirements. https://blog.nftx.org/cryptopunks-single-index-funds-now-also-on-sushiswap/

smb2796:
Is there a particular reason why this was kept separate from your omnibus UMIP?

smb2796:
Closing because this was added to the [Omnibus UMIP](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-82.md)

PR #257, opened 2021-04-21
Update Price Identifier Template

smb2796:
NOTE: This template update cannot be voted on.

The UMA guidelines for new price identifiers have always been intended to elicit answers to the following questions.
- What price is being proposed?
- How should voters calculate that price?
- Is this the best calculation methodology to use?
- Why should this price identifier be added?
- Are there any security concerns associated with adding or using this price identifier?

As long as these general questions are being answered, it is necessary that the templates themselves evolve in order to best meet the needs of authors and reviewers. 

Previously we have seen two distinct types of templates. The original barebones template provided only section headers and broad instructions. This created a lot of flexibility for UMIP authors, but also made UMIP review and authorship challenging as UMIP content was not very standardized. 

The newer [template](https://github.com/UMAprotocol/UMIPs/blob/master/price-identifier-template.md), currently used today, was intended to be more explicit in its instructions to make authorship and review easier. In some aspects I believe that has been successful, but has also created a format that is difficult to read and oftentimes contains unnecessary information.

This new price identifier template attempts to find a happy medium between the two previous templates.

Intended improvements and updates:
- Improves formatting.
- Removes question/answer format formerly present in some sections.
- Provides more explicit instructions on what should information be filled in and what should be removed.
- Removes fields that are no longer necessary (Intended Collateral Currency).
- Removes redundancies between sections.
- Renames `Markets & Data Sources` to `Data Specifications`.
- Asks for an example calculation to accompany the voting implementation.
- Removes the accompanying "price-identifier-guide" document, which was generally redundant and only encouraged authors to copy/paste the examples rather than think through and author their own UMIP.


akhanubis:
Does it make sense to add a price feed similar to ExpressionPriceFeed that can be used to describe IDs that use different implementations based on the timestamp (before and at settlement)? Example: before settlement use UniswapPriceFeed with X config, at settlement use MyNewPriceFeed with Y config.

smb2796:
> Does it make sense to add a price feed similar to ExpressionPriceFeed that can be used to describe IDs that use different implementations based on the timestamp (before and at settlement)? Example: before settlement use UniswapPriceFeed with X config, at settlement use MyNewPriceFeed with Y config.

It's a good thing to discuss. I think this makes more sense to do with a Financial Product Library (FPL) and can already be done today. An FPL can allow a contract to switch price identifiers depending on the price request timestamp. This would mean that there doesn't need to be any sort of switching logic in the price feeds, because the two price identifiers would be handled separately. [Here](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol) is an example of an FPL that does this.

price-identifier-template.md
```diff
@@ -1,228 +1,83 @@
...
-| Created    | [DATE] 
-| Link to Discourse    | [LINK]                                                                               
-<br>
-Note that an UMIP number will be assigned by an editor. When opening a pull request to submit your UMIP, please use an abbreviated title in the filename, umip-draft_title_abbrev.md.
-The title should be 44 characters or less.
+# Summary 
 
-<br>
-<br>
+The DVM should support price requests for **Price Identifier Name**. **Price Identifier name** reflects the **Summary of Price Identifier**.
```
pemulis, line 15: Nit: Capitalize the `n` in `name` in the second bolded portion.

price-identifier-template.md
```diff
@@ -1,228 +1,83 @@
...
-"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the issue.
+# Motivation
 
-- **Example** 
+*Please explain why you want to add this price identifier. What types of synthetics are you intending to create with this?*
 
-    - The DVM should support price requests for the [Base currency / Quote currency] price index.
+# Data Specifications
 
+*How should voters access the data necessary to calculate the value of this price identifier? What specific markets should be referenced?*
```
pemulis, line 24: `What specific markets or other data sources should be referenced?` -- in the case that the price identifier doesn't have a market-based data source, like sports prediction markets, insurance, etc.

price-identifier-template.md
```diff
@@ -1,228 +1,83 @@
...
 
-# MOTIVATION
+*If proposing multiple price identifiers, please add markets for each.*
```
pemulis, line 26: `please add markets or other data sources`

price-identifier-template.md
```diff
@@ -1,228 +1,83 @@
...
 
-This section should clearly explain the types of financial products that will be created and the mechanics of an example financial product using this price identifier. Please answer the following questions:
+-----------------------------------------
+- Price identifier name: **First Price ID Name** 
+- Markets & Pairs: **Markets & Pairs** - *Example: Binance ETH/USDT, Coinbase Pro ETH/USD. This might not apply to all price identifiers*
+- Example price providers: **Provider to use** - *Cryptowatch, TraderMade, Quandl, the Graph*
```
pemulis, line 31: `Example price or data providers`

price-identifier-template.md
```diff
@@ -1,228 +1,83 @@
...
+- Cost to use: **Explanation or link to provider pricing plan**
+- Real-time price update frequency: **Frequency** - *60 seconds*
```
pemulis, line 33: `data update`

price-identifier-template.md
```diff
@@ -1,228 +1,83 @@
...
+- Historical price update frequency: **Frequency** - *5 minutes*
```
pemulis, line 34: `data update`

price-identifier-template.md
```diff
@@ -1,228 +1,83 @@
...
-
-2. How could this price ID be exploited?
-
-3. Do the instructions for determining the price provide people with enough certainty?
-
-4. What are current or future concern possibilities with the way the price identifier is defined?
-
-5. Are there any concerns around if the price identifier implementation is deterministic?
+Some optional questions to consider: (*Please remove before submission*)
+- How could price manipulation occur?
```
pemulis, line 79: `price manipulation or data feed manipulation`

pemulis (commented):
Really like how this simplifies things. I suggested some changes for price identifiers that rely on data feeds that are not, strictly speaking, price feeds. For example, the score of a sports game may determine the price of a synthetic token that pays out based on the results, but the score itself is not a price. Similar for insurance and a bunch of other possible use cases for the optimistic oracle.

price-identifier-template.md
```diff
@@ -1,228 +1,83 @@
...
-| Created    | [DATE] 
-| Link to Discourse    | [LINK]                                                                               
-<br>
-Note that an UMIP number will be assigned by an editor. When opening a pull request to submit your UMIP, please use an abbreviated title in the filename, umip-draft_title_abbrev.md.
-The title should be 44 characters or less.
+# Summary 
 
-<br>
-<br>
+The DVM should support price requests for **Price Identifier Name**. **Price Identifier Name** reflects the **Summary of Price Identifier**.
```
mhairimca, line 15: Great improvement to simplify for UMIP authors

pemulis (approved):

PR #261, opened 2021-04-21
WIP: Add cryptopunk index umip

cc7768:


UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,151 @@
...
+The post-expiry price implementation can be found at [PENDING]()
+
+
+# Technical Specifications
+
+-----------------------------------------
+- Price identifier name: `CRYPTOPUNKX-30d`
+- Base Currency: **ETH**
+- Quote Currency: **ETH**
+- Rounding: Round to 6 decimal places (third decimal place digit >= 5 rounds up and < 5 rounds down)
```
smb2796, line 107:
```suggestion
- Rounding: Round to 6 decimal places (seventh decimal place digit >= 5 rounds up and < 5 rounds down)
```

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,151 @@
+## Headers
+
+| UMIP [#]            |                                                                           |
+| ------------------- | ------------------------------------------------------------------------- |
+| UMIP Title          | [Add CRYPTOPUNKX-30d as DVM price identifier]                             |
```
smb2796, line 5: Can remove all of the square brackets here.

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,151 @@
...
+# Data Specifications
+
+All relevant price data is computed using information that can be found on the blockchain.
+
+* Pre-expiry, the token is valued according to the price that it trades at in the highest volume Uniswap pool.
+* Post-expiry, the token is valued according to a function of the `PunkBought` events published by the CryptoPunks Market contract
+
+-----------------------------------------
+- Price identifier name: `CRYPTOPUNKX-30d`
+- Markets & Pairs: **Markets & Pairs** - Uniswap `CRYPTOPUNKX-30d/ETH`, CryptoPunk Market contract `PunkBought` events
```
smb2796, line 47: The bolded headers should be removed. They are intended to be replaced by your answers.

UMIPs/umip-add-cryptopunk.md
cc7768: :pray:

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+The post-expiry price implementation can be found at [PENDING]()
+
+
+# Technical Specifications
+
+-----------------------------------------
+- Price identifier name: `CRYPTOPUNKX-30d`
+- Base Currency: **ETH**
```
smb2796, line 105: Base currency should really be "median Cryptopunk trade" and quote currency should be ETH. Perhaps "Currency" should be removed from these headers.

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,151 @@
...
+2. Retrieve 2 hour TWAP from the given pool which can be done using the Uniswap Price Feed from UMA's financial templates library
+
+
+**Post-expiry**
+
+When a price request is made post expiration, the following process should be followed:
+
+1. Retrieve all CryptoPunk `PunkBought` events from the 30 days prior to expiration
+2. Identify the last price that each CryptoPunk was traded at using the event data
+3. Take the median of these pricese
```
smb2796, line 137:
```suggestion
3. Take the median of these prices
```

smb2796:
Thoughts on separating this out into two price identifiers? One that refers to the self-referential price of the synth that is yet to be created, and one that is referring only to the 30 day cryptopunks median?

This would allow us to use a FPL to switch the price identifiers at expiry, rather than having the switching logic housed in the price id. In the future this would mean that we could reuse the cryptopunks median for other products.

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,151 @@
...
+cryptopunks = []
+
+# Create mapping to store most recent price
+cryptopunk_blockprice = {}
+
+# Find the last PunkBought event for each CryptoPunk
+for event in events:
+    if event.cryptopunk not in cryptopunks:
+        push(event.cryptopunk, cryptopunks)
+        cryptopunkprice[event.cryptopunk] = {"block": event.block, "value": event.value}
```
Reinis-FRP, line 88: Or `cryptopunk_blockprice[event.cryptopunk] = {"block": event.block, "value": event.value}`?

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,151 @@
...
+# Data Specifications
+
+All relevant price data is computed using information that can be found on the blockchain.
+
+* Pre-expiry, the token is valued according to the price that it trades at in the highest volume Uniswap pool.
+* Post-expiry, the token is valued according to a function of the `PunkBought` events published by the CryptoPunks Market contract
+
+-----------------------------------------
+- Price identifier name: `uPUNK`
+- Markets & Pairs: Uniswap `uPUNK/ETH`, CryptoPunk Market contract `PunkBought` events
```
Reinis-FRP, line 47: Is it possible to include the address of CryptoPunk Market contract here?

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,151 @@
+## Headers
+
+| UMIP [#]            |                                                                           |
+| ------------------- | ------------------------------------------------------------------------- |
+| UMIP Title          | Add uPUNK as DVM price identifier                                         |
```
smb2796, line 5:
```suggestion
| UMIP Title          | Add uPUNK as a price identifier                                         |
```

UMIPs/umip-add-cryptopunk.md
cc7768: Yep! :grinning:

UMIPs/umip-add-cryptopunk.md
cc7768: This is a great idea -- Saves other people the need to look it up.

UMIPs/umip-add-cryptopunk.md
cc7768: :+1: -- I have some other changes in title so I'm going to do manually, but, otherwise, I would have accepted suggestion.

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,203 @@
...
+# Data Specifications
+
+All relevant price data is computed using information that can be found directly on the blockchain.
+
+-----------------------------------------
+
+The `PUNKETH-TWAP` price identifier depends on prices drawn from the `PunkBought` events of the CryptoPunk market contract
+
+- Price identifier name: `PUNKETH`
+- Markets & Pairs: CryptoPunk Market contract `PunkBought` events. The CryptoPunk contract address is `0xb47e3cd837dDF8e4c57F05d70Ab865de6e193BBB` which you can see at https://etherscan.io/address/0xb47e3cd837ddf8e4c57f05d70ab865de6e193bbb
```
pemulis, line 47: How does this market relate to Wrapped CryptoPunks, if at all? Would all relevant CryptoPunk sales be captured by this market contract?

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,203 @@
...
+- Markets & Pairs: Uniswap `uPUNK/ETH`
+- Example price providers: The Uniswap price data can be obtained directly from the blockchain
+- Cost to use: [Infura](https://infura.io/) supports up to 100,000 requests per day for free. This information should also available on [The Graph](https://thegraph.com/)
+- Real-time price update frequency: Updated every block
+- Historical price update frequency: Updated every block
+
+
+# Price Feed Implementation
+
+The `PUNKETH` price would require the creation of a new price feed. The pseudo-code for such an identifier is below:
```
pemulis, line 67: Note to other reviewers: I have this new price feed on my plate for next week.

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,203 @@
...
+- Estimated current value of price identifier: 22.430000 (as of 21 April 2021 19:34 UTC)
+
+
+
+# Rationale
+
+The `PUNKETH` price identifier had a few decisions that we believe were important to the design:
+
+* _CryptoPunks_: As mentioned earlier in this document, we chose to build an index using CryptoPunks because they were the original NFT. This originality has lead to them being highly valued and having consistent enough trade volume.
+* _30 day average_: The 30 day average allows for the index to reflect common trading prices a cross many CryptoPunks rather than to respond to particular transactions
```
pemulis, line 135: `trading prices across`

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,203 @@
...
+
+The other main concern is if there were just insufficient CryptoPunk trades being made. If there were only 1-2 trades happening every 30 days, this index becomes much less useful because there's less information contained in its price.
+
+### PUNKETH-TWAP
+
+The main concerns of the TWAP price are:
+
+1. **Token price manipulation**: If the Uniswap pool is not sufficiently liquid, then attackers could try to drive down the Uniswap price and withdraw more collateral than intended. Most DeFi attacks have been done using flash loans, but flash loans would be ineffective since the price is measured at the end of each block. Collateralizaton based on the TWAP price would make it more capital intensive (and thus risky) to target the token price in this way.
+2. **TWAP mismatch**: If the price of the token rises quickly then there would become a mismatch between the market price and the TWAP price. This might allow sponsers to mint tokens with less collateral than what they could seel them for on the market. Reasonable levels of collateralizaton requirements and the 2 hour "liveness period" help combat this concern.
```
pemulis, line 201: `sell them for`

UMIPs/umip-add-cryptopunk.md
therealhexi: Curious: would this price be along the line of `InfuraPriceFeed.js`?

therealhexi (commented):
For my curiosity

UMIPs/umip-add-cryptopunk.md
smb2796, line 135: In other locations you refer to the median. Does this 30 day average apply?

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,203 @@
...
+### PUNKETH
+
+When a price request is made, the following process should be followed:
+
+1. Retrieve all CryptoPunk `PunkBought` events from the 30 days prior to expiration
+2. Identify the last price that each CryptoPunk was traded at using the event data -- Only one price should be produced per CryptoPunk even if they had traded multiple times
+3. Take the median of these prices - If there is an even number of prices, take the mean of the two values closest to the median.
+
+
+**Example**
```
smb2796, line 155: +100 for examples. I think it makes the implementation section much more useful.

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,203 @@
...
+
+If the timestamp requested was `1619222400` then:
+
+* We would need to identify all `PunkBought` events from `1619222400 - 30 days -> 1616630400` to `1619222400`
+* Imagine that we had 5 `PunkBought` events with (`ts`, `punk`, `eth`) pairs of `[(1616630450, 1000, 20), (1616631450, 5000, 30), (1616631550, 5000, 35), (1618631550, 6000, 22), (1618632550, 9999, 15)]` then the prices we would use to compute the median would be `[20, 35, 22, 15]`
```
smb2796, line 160: What are the 'punk' values here? Are these punk identifiers?

UMIPs/umip-add-cryptopunk.md
mhairimca: If there is active manipulation at the same time as a (genuine) TWAP mismatch, how could voters determine what was legit and what was manipulation to come to an accurate price.

cc7768: ^Thanks @pemulis 

@mhairimca. I think this is an important question. I don't actually know how the voters would determine what a legit price would be -- In some ways, I don't know that we "need" to know though because I think the "right" answer would be case specific.

One solution for a shorter term attack might be to use the last price voters thought was realized prior to price manipulation.

UMIPs/umip-add-cryptopunk.md
cc7768: Yep -- I'm going to change this label from `punk` to `punk_id` to make that more clear.

UMIPs/umip-add-cryptopunk.md
pemulis: I'm using the `UniswapPriceFeed` as a jumping-off point, since this is also dealing with on-chain data and events.

smb2796 (approved):
Looks good! Going to approve and move to `Last Call` status. The one outstanding item from the UMIP community call is adding in a PR for the expiry price feed. Since this should be up in the next day or so before voting starts, I'm going to merge this in and we can add the PR link into the UMIP in a subsequent PR.

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,203 @@
...
+2. Identify the last price that each CryptoPunk was traded at using the event data -- Only one price should be produced per CryptoPunk even if they had traded multiple times
+3. Take the median of these prices - If there is an even number of prices, take the mean of the two values closest to the median.
+
+
+**Example**
+
+If the timestamp requested was `1619222400` then:
+
+* We would need to identify all `PunkBought` events from `1619222400 - 30 days -> 1616630400` to `1619222400`
+* Imagine that we had 5 `PunkBought` events with (`ts`, `punk_id`, `eth`) pairs of `[(1616630450, 1000, 20), (1616631450, 5000, 30), (1616631550, 5000, 35), (1618631550, 6000, 22), (1618632550, 9999, 15)]` then the prices we would use to compute the median would be `[20, 35, 22, 15]`
```
krzysu, line 160: @smb2796 @pemulis @therealhexi CryptoPunksMarket contract has a bug, that when bid accepted, PunkBought event is triggered with price equal to 0. I remember scratching my head around it for a while. So that makes getting correct sale prices more complicated.
If you figure it out, I would be interested in using some Punk prices feed/api :)

pemulis: Are you sure this is a bug in `acceptBidForPunk`? It doesn't look like the `bid.value` is overwritten. My suspicion is that the zero-value `PunkBought` events are triggered by the `buyPunk` method (https://github.com/larvalabs/cryptopunks/blob/master/contracts/CryptoPunksMarket.sol#L156), where an address can be specified as an approved buyer. In those cases, I assume payment is happening elsewhere, not through the market contract, and the actual price reported in the event (`0`) is not correct. In order to get the actual price, you would need to figure out where it sold.

pemulis: Hmm, @cc7768 pointed out this transaction that reported a `0` value on a call to `acceptBidForPunk`, so you may be right. https://etherscan.io/tx/0xf9db9e3fa5a3aa7975ab89bb35529b1274500d0b4ce54db8f68dc7b686685ca3

pemulis: Current thinking: Take the most recent `PunkBidEntered` event prior to the `PunkBought` event, since only one bid exists at a time (and is guaranteed to be the highest bid). You can be certain that the `value` in `PunkBidEntered` is the sale price.

pemulis: You were totally right about the nature of the bug.

krzysu: always happy to help! :)

UMIPs/umip-add-cryptopunk.md
```diff
@@ -0,0 +1,203 @@
...
+  - An investor who believes that the median trade price of CryptoPunks will decrease could mint `uPUNK` and sell the minted tokens.
+
+
+# Data Specifications
+
+All relevant price data is computed using information that can be found directly on the blockchain.
+
+-----------------------------------------
+
+The `PUNKETH-TWAP` price identifier depends on prices drawn from the `PunkBought` events of the CryptoPunk market contract
```
pakokrew, line 44: The `PUNKETH` price

PR #268, opened 2021-04-28
create umip_add_Digg_Positive_Rebases

jon-tompkins:


umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,67 @@
...
+min(if(r<5,0,((r-5)/(30-5))^1.5),1)
+https://docs.google.com/spreadsheets/d/1Kb58KUiaCFClfL9hkf0OCXJzXHC-9lDwnrxQ3eEobg4/edit?usp=sharing
+
+
+
+
+# Technical Specifications
+
+Price identifier name: DIGG_Positive_Rebases
+Base Currency: NA
```
mhairimca, line 46: The base currency would be the token that you are planning to airdrop

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,67 @@
...
+
+We would like this identifier added so that we can distriute positive rebase options as a mechanism to help support the stability of DIGG.  More details on the campaign can be found in this post:
+https://medium.com/badgerdao/badgerdao-x-uma-introducing-rebase-mining-3c663a5abdce
+
+# Data Specifications
+
+-----------------------------------------
+- Price identifier name: **DIGG_Positive_Rebases** 
+- Markets & Pairs: **N/A**
+- Example data providers: **Available on-chain**
```
mhairimca, line 25: Needs to identify contract addresses to be queried.

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,67 @@
...
+- Price identifier name: **DIGG_Positive_Rebases** 
+- Markets & Pairs: **N/A**
+- Example data providers: **Available on-chain** 
+- Cost to use: **None**
+- Real-time data update frequency: **Daily** 
+- Historical data update frequency: **Daily** 
+
+# Price Feed Implementation
+
+At 12:00 UTC every day compare the total supply of DIGG (https://etherscan.io/token/0x798d1be841a82a273720ce31c822c61a67a601c3) to its supply from the previous day.  If it is greater than the previous day then the price feed should iterate by 1.  The exact amount of DIGG in the redemption pool has not been settled yet.  the below formula and accompanying spreadsheet illustrates how to calculate the percentage of the redemption pool that each option will be redeemable for.  Once the size of the pool is known (in bDIGG) then the formula can be updated by multiplying by the total redemption pool size.
```
mhairimca, line 32: Clarify start/end times

mhairimca: Although there is no need to determine this in the PI, the exact max amount of digg would have to be identified prior to deployment so that it could be locked as collateral.

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,67 @@
...
+# Implementation
+
+I think comparing total supply one day to the other is the easiest but am open to other ways to implement this.
+
+1. Using the timestamp that falls on 12:00 UTC but is closest and earlier than the price request timestamp (D2), read totalSupply from the DIGG token contract (D2_Supply).
+2. Query for totalSupply at 12:00 UTC on the day proceeding the day's (D1_Supply).
+3. If D2_Supply > D1_Supply  (if supply is equal do not iterate), increment DIGG_Positive_Rebases by 1. If D2_Supply <= D1_Supply, the DIGG_Positive_Rebases value should remain constant.
+4. Steps 2 and 3 should be repeated until the D1_Supply timestamp is earlier than the block timestamp that the contract using this price identifier was launched in OR the days since the deployment = to 30.
+5. Voters should return the value of DIGG_Positive_Rebases once one of the conditions of step 4 is met. This value should be returned as is with no rounding.
+
```
mhairimca, line 60: This simply gives the number of positive rebases, it doesnt specifiy the min or max values or how to calculate the metric in between these values. 

In broad chunks you would need something like 
 - Step 1 - Calculate the number of days of positive rebases
 - Step 2 - If positive rebases is less than or equal to 5, value =0
  - Step 3 - if positive rebases are between 5 and 30 then calculate according to [formula]
  - Step 4 - if positive rebases are greater than 30 then value = [max value]
  
  More flesh would have to be put on these steps to make it deterministic and identify exactly how the calculation should be done.

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,67 @@
...
+# Rationale
+
+This price feed is very specific to these options and needs to align to start at the date they are issued and stop running at maturity (30 days post launch).  We may look ot use this mechanism in the future and are open to suggestions on how to restructure the price feed to make it more flexible for future use.
```
mhairimca, line 63: Once a price feed is approved it is available for anyone to use (although as its so specific, its unlikely that people would want to.)

The contract should be deployed on the date the options start with an expiry date of 30 days hence, this isnt something that needs to go in the UMIP tho, as these are contract deployment issues.

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,67 @@
+
+| UMIP                |                                                               |
+| ------------------- | ------------------------------------------------------------- |
+| UMIP Title          | Add DIGG DIGG_Positive_Rebases as a supported price identifier|
+| Authors             | Jon (BadgerDAO)                                               |
+| Status              | Draft                                                         |
+| Created             | 4/27/2021                                                     |
+| Discourse Link      | **Create a post in [UMA's Discourse](https://discourse.umaproject.org/c/umips/18) and link here**            |
```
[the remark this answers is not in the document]
jon-tompkins: Added

umip_add_Digg_Positive_Rebases.md
jon-tompkins: The asset we are trying to airdrop is ultimately DIGG but will likely be managed as the staked version bDIGG because it is a simple erc20 and doesnt rebase.  For the purposes of these options is it beneficial for us to take that bDIGG and put them into a vault and mint a new token (ubDIGG) of which 100% of the supply will be used for this program and then have that be what is referenced for the price?

umip_add_Digg_Positive_Rebases.md
jon-tompkins: Added token contract is that what you were looking for?

umip_add_Digg_Positive_Rebases.md
jon-tompkins: Yes, this will be decided early next week and will be denominated in bDIGG

umip_add_Digg_Positive_Rebases.md
jon-tompkins: The formula was in price feed implementation section, i have copied down to this area.

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,76 @@
...
+4. Steps 2 and 3 should be repeated until the D1_Supply timestamp is earlier than the block timestamp that the contract using this price identifier was launched in OR the days since the deployment = to 30.
+5. Voters should return the value of DIGG_Positive_Rebases once one of the conditions of step 4 is met. This value should be returned as is with no rounding.
+
+This forumla can be used to determine the percentage of the redemption pool options will be eligible to redeem (redemption pool size to be determined through badger governance)
+r = total rebases (using iterative calculation above)
+5 = minmum rebases
+30 = max rebases
+min(if(r<5,0,((r-5)/(30-5))^1.5),1)
+https://docs.google.com/spreadsheets/d/1Kb58KUiaCFClfL9hkf0OCXJzXHC-9lDwnrxQ3eEobg4/edit?usp=sharing
+
```
Reinis-FRP, line 67: Since spreadsheet is formatted as percentage with 2 decimals that implies effective 4 decimals for 0-1 range while only 2 decimals are set in the `Technical Specifications`

umip_add_Digg_Positive_Rebases.md
mhairimca: Ah, yes, it would have to be bDIGG that was locked as collateral and distributed at the end of the contract....unless you had some kind of customised "unwrapper" that would change the bDIGG into DIGG at expiry.

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,76 @@
...
+Price identifier name: DIGG_Positive_Rebases
+Base Currency: NA
+Quote Currency: NA
+Rounding: Round to 2 decimal places (third decimal place digit >= 5 rounds up and < 5 rounds down)
+Estimated current value of price identifier: 0
+
+# Implementation
+
+I think comparing total supply one day to the other is the easiest but am open to other ways to implement this.
```
therealhexi, line 53: Can you provide an easy way to fetch the total supply amount at a particular timestamp (12:00 UTC every day)?

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. Using the timestamp that falls on 12:00 UTC but is closest and earlier than the price request timestamp (D2), read totalSupply from the DIGG token contract (D2_Supply).
```
therealhexi, line 55: Does `the most recent 12:00 UTC preceding the request timestamp` sound neater?

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,76 @@
...
+2. Query for totalSupply at 12:00 UTC on the day proceeding the day's (D1_Supply).
+3. If D2_Supply > D1_Supply  (if supply is equal do not iterate), increment DIGG_Positive_Rebases by 1. If D2_Supply <= D1_Supply, the DIGG_Positive_Rebases value should remain constant.
```
therealhexi, line 57: increment DIGG_Positive_Rebases by `the formula min(if(r<5,0,((r-5)/(30-5))^1.5),1)`?

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,76 @@
...
+4. Steps 2 and 3 should be repeated until the D1_Supply timestamp is earlier than the block timestamp that the contract using this price identifier was launched in OR the days since the deployment = to 30.
```
therealhexi, line 58: Do you mean "Steps 2 and 3 should be repeated until the D1_Supply timestamp is `after` the block timestamp"?

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,76 @@
...
+5. Voters should return the value of DIGG_Positive_Rebases once one of the conditions of step 4 is met. This value should be returned as is with no rounding.
```
therealhexi, line 59: Earlier you mentioned to round to 2 decimals. Should we keep things consistent?

umip_add_Digg_Positive_Rebases.md
therealhexi: Would be good if you give an example to help readers understand.

therealhexi (commented):
Initital review

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Technical Specifications
+
+Price identifier name: DIGG_Positive_Rebases
+Base Currency: NA
+Quote Currency: NA
+Rounding: Round to 2 decimal places (third decimal place digit >= 5 rounds up and < 5 rounds down)
+Estimated current value of price identifier: 0
+
+# Implementation
```
akhanubis, line 51: I think a much better implementation would be to use the `LogRebase` event to fetch all the changes to the total supply in one RPC call. Current implementation requires either persistently storing the value of the price ID over time from the beginning, or an archive node to read totalSupply values in the past.

You can use `getPastLogs` to get all `LogRebase` events for the DIGG address since the beginning of the rebase mining program and then compare each event's totalSupply with the previous one to count how many positives rebases had been so far. 

There are no liquidation bots so I guess it doesnt matter that much, but still.

umip_add_Digg_Positive_Rebases.md
smb2796: Also the base currency doesn't actually apply to the price identifier, so I think it works as NA.

I don't think it is necessary to use ubDIGG. I don't see any downsides in just using bDIGG, and ubDIGG would need to be approved as a collateral to be used. So definitely recommend just sticking with bDIGG.

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,76 @@
...
+This forumla can be used to determine the percentage of the redemption pool options will be eligible to redeem (redemption pool size to be determined through badger governance)
+r = total rebases (using iterative calculation above)
+5 = minmum rebases
+30 = max rebases
+min(if(r<5,0,((r-5)/(30-5))^1.5),1)
+https://docs.google.com/spreadsheets/d/1Kb58KUiaCFClfL9hkf0OCXJzXHC-9lDwnrxQ3eEobg4/edit?usp=sharing
+
+
+
+# Rationale
```
smb2796, line 70: The Rationale section should be moved above the `Implementation` section to follow the format of other UMIPs.

umip_add_Digg_Positive_Rebases.md
Hahz48: line 63 2nd sentence: Should say "we may look to"/not "we may look ot"

umip_add_Digg_Positive_Rebases.md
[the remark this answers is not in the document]
mhairimca: I dont see a link

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,76 @@
...
+- Cost to use: **None**
+- Real-time data update frequency: **Daily** 
+- Historical data update frequency: **Daily** 
+
+# Price Feed Implementation
+
+At 12:00 UTC every day compare the total supply of DIGG (https://etherscan.io/token/0x798d1be841a82a273720ce31c822c61a67a601c3) to its supply from the previous day.  If it is greater than the previous day then the price feed should iterate by 1.  The exact amount of DIGG in the redemption pool has not been settled yet.  the below formula and accompanying spreadsheet illustrates how to calculate the percentage of the redemption pool that each option will be redeemable for.  Once the size of the pool is known (in bDIGG) then the formula can be updated by multiplying by the total redemption pool size.
+
+r = total rebases (using iterative calculation above)
+5 = minmum rebases
```
pemulis, line 35: `minimum rebases`

umip_add_Digg_Positive_Rebases.md
pemulis: Just checking: Do we only count one rebase per day towards the total? I agree checking the `LogRebase` events seems like the way to go, and if we're only counting one per day, we can keep the timestamp calculation logic in the implementation section to see which day each event occurred on.

smb2796: I believe there is only 1 rebase per day. @jon-tompkins thoughts on using this implementation instead?

umip_add_Digg_Positive_Rebases.md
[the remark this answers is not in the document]
jon-tompkins: I see it there https://discourse.umaproject.org/t/umip-81-add-digg-positive-rebases-price-identifier/1026

umip_add_Digg_Positive_Rebases.md
jon-tompkins: that works for me.  the flow is going to be to deposit the approved DIGG into bDIGG once governance has approved and  then use that for the pool when ready

umip_add_Digg_Positive_Rebases.md
jon-tompkins: Added Examples

umip_add_Digg_Positive_Rebases.md
jon-tompkins: I am open to this approach as long as it can be determined if, when the function is called, that the result is a positive vs negative rebase.

umip_add_Digg_Positive_Rebases.md
```diff
@@ -0,0 +1,81 @@
+
+| UMIP                |                                                               |
+| ------------------- | ------------------------------------------------------------- |
+| UMIP Title          | Add DIGG DIGG_Positive_Rebases as a supported price identifier|
```
smb2796, line 4:
```suggestion
| UMIP Title          | Add DIGG_Positive_Rebases as a supported price identifier|
```

umip_add_Digg_Positive_Rebases.md
akhanubis: > I am open to this approach as long as it can be determined if, when the function is called, that the result is a positive vs negative rebase.

You will get a list of logs with timestamp and the new total supply so you can sort them and then compare. Whenever the total supply of a log is lower than the previous log total supply, count 1 positive rebase.

umip_add_Digg_Positive_Rebases.md
jon-tompkins: thanks

umip_add_Digg_Positive_Rebases.md
jon-tompkins: great, sounds good

umip_add_Digg_Positive_Rebases.md
jon-tompkins: So, this may come down to nomenclature.  We need to iterate the positive rebases  but the number of positive rebases is in input into the formula which then results in the % of the pool the options should be able to settle for.  r in the forumla needs to increment, which will make the output go up, but not linearly.

umip_add_Digg_Positive_Rebases.md
jon-tompkins: no, the before component is so it doesnt start counting too early.

smb2796 (approved):
Looks good overall. If BadgerDAO assesses the LogRebase method and decides to use it, the implementation in the UMIP can be updated at a later date.

Going to approve and move to Last Call as most other issues seem to be minor.

umip_add_Digg_Positive_Rebases.md
jon-tompkins: i have adjusted the decimals to be more consistent

umip_add_Digg_Positive_Rebases.md
jon-tompkins: spreadsheet has been updated and decimals adjusted in the post

PR #269, opened 2021-04-28
Update uPUNK price implementation

cc7768:
This is a solution to the concern raised in https://github.com/UMAprotocol/UMIPs/pull/261#discussion_r622282931


smb2796 (approved):
Looks good!

PR #272, opened 2021-04-30
Update CNYUSD UMIP rounding spec

smb2796:
The CNYUSD price from Tradermade is only ever returned to 5 decimals. There was an oversight in the UMIP where rounding was specified to 6 decimals. Although this would have no affect on the actual returned price, it seems best to update this UMIP to avoid any confusion.


ChandlerDeKock (approved):
approved - thanks

PR #276, opened 2021-05-05
Add ifarm price ID

gruadus:
Trying to add seperate iFARM UMIP


smb2796 (approved):
Approving to merge in and move to Last Call

PR #280, opened 2021-05-05
[UMIP-90] Add NGNUSD, PHPUSD, CADUSD, ZARUSD, KRWUSD and JPYUSD as price identifiers

SayWhatDev:


UMIPs/umip-90.md
```diff
@@ -0,0 +1,152 @@
+## Headers
+| UMIP-90   |                                                          |
+|------------|----------------------------------------------------------|
+| UMIP Title | Add NGNUSD, PHPUSD, CADUSD, ZARUSD, KRWUSD and JPYUSD as price identifiers                           |
+| Authors    | Vsevolod Potorocha (vsevolod@jarvis.network)                 |
+| Status     | Draft                                                 |
+| Created    | 30/04/2021                                   |
+
```
akhanubis, line 8: Missing discourse link for the thread related to this UMIP

https://discourse.umaproject.org/c/umips/18

akhanubis (commented):
Looks good overall.

Have you considered using a self referencing price during non market  working hours like some of the other IDs using forex or stock markets?

smb2796:
Agreed, this looks really good! I think the self-referential model could potentially work for these use cases - but I believe the intention is to use these with a perpetual synthetic, which would make the self-referential model a little clunky. Also in the past we have not used the self-referential model for fx prices, given that they are low risk for moving more than 1-2% in off-hours. But if these pairs are more volatile than that, perhaps we should reconsider.

UMIPs/umip-90.md
SayWhatDev: > Missing discourse link for the thread related to this UMIP
> 
> https://discourse.umaproject.org/c/umips/18

Link added in the last commit :)

smb2796:
> > Agreed, this looks really good! I think the self-referential model could potentially work for these use cases - but I believe the intention is to use these with a perpetual synthetic, which would make the self-referential model a little clunky. Also in the past we have not used the self-referential model for fx prices, given that they are low risk for moving more than 1-2% in off-hours. But if these pairs are more volatile than that, perhaps we should reconsider.
> 
> @smb2796 Thanks for that input. Yes the price feeds are used for a perpetual synthetic. However some of those pairs have higher volatility than as example the majors (EURUSD,GBPUSD etc). I am not entirely sure what you refer to as a `self-referential model`, but I'll look into it :)

A good example would be uSTONKS_JUN21, which prices based off of its own synthetic price pre-expiry, and then switches to calculating the value of an index of stocks at expiry. For your case with a perpetual, it isn't as clear how this would be done - given that you don't have a set expiry timestamp and that the FX rates would probably only be used in the case of an `emergencyShutdown`.

PR #281, opened 2021-05-09
[UMIP-91] Add UST as whitelisted collateral currency

SayWhatDev:


therealhexi:
This file is already included in #283 there is no need to keep this PR open

PR #282, opened 2021-05-09
[UMIP-92] Add yvUSDC as whitelisted collateral currency

SayWhatDev:


therealhexi (commented):
This file is already included in #283 there is no need to keep this PR open

PR #290, opened 2021-05-12
Add `SelfMintingDerivativeFactory` as `Creator` in DVM registry

SayWhatDev:


UMIPs/umip-94.md
```diff
@@ -0,0 +1,90 @@
...
+
+| UMIP-XX  |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Add `SelfMintingDerivativeFactory` as `Creator` in DVM registry                                                                                                |
+| Authors    | Pascal Tallarida (pascal@jarvis.network)                 |
+| Status     | Draft                                                 |
+| Created    | 12/05/2021   
+| Discourse link    | https://discourse.umaproject.org/t/add-selfmintingderivativefactory-as-creator-new-uma-fork/1079                                   |
+
+# Summary
```
Reinis-FRP, line 11: I found the [http://bit.ly/Synthereum_Manifesto](http://bit.ly/Synthereum_Manifesto) really helpful in trying to understand how this UMIP fits into Jarvis network design and roadmap. So you should definitely link it here

UMIPs/umip-94.md
```diff
@@ -0,0 +1,90 @@
...
+
+Last year, we have forked the perpetual contracts implementation of UMA and proposed the [UMIP-34](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-34.md) to integrate it within the DVM. We called this contract PerpetualPoolParty. This fork allows our synthetic assets to be multi-collateralized (multiple collateral linked to multiple Derivatives contract can back the same synthetic asset), perpetual (but without stability fee, funding rate etc.), and permissioned (only our whitelisted Smart Contracts can mint them).
```
Reinis-FRP, line 13: Have you also merged in any updates in UMA protocol contracts since the initial fork?

UMIPs/umip-94.md
```diff
@@ -0,0 +1,90 @@
...
+
+The UMA’s governance approved this fork earlier this year, allowing us to launch our mainnet.
+
+The PerpetualPoolParty was a “broker contract” allowing the end-users to trade against a smart contract (the broker's liquidity pool) which is the sole Token Sponsor: when a user wants to buy a synthetic asset, it triggers a mint transaction where the liquidity pool self-mints a synthetic fiat with a collateral in USDC deposited by liquidity providers, and sell this synthetic fiat for USDC to the end-user, at the Chainlink price, within the same transaction.
+
+Now we are about to launch our “bank contract” which allows anyone to be a Token Sponsor, pretty much like the EMP. Yet, we could not use the EMP due to some specification of our protocols (multi-collateral synthetic assets, perpetual, set minting limits, etc.). So we forked PerpetualPoolParty.
```
Reinis-FRP, line 19: Is the “bank contract” effectively the same idea that was proposed in [http://bit.ly/Synthereum_Manifesto](http://bit.ly/Synthereum_Manifesto) p11 as a "borrowing pool (self-minting)" to allow the liquidator to mint synths?

If yes, does it really solves the problem for liquidators being able to perform liquidation on undercollaterized liquidity pool positions?  E.g. I would assume a liquidator could self-mint and liquidate, but in order to get back their full collateral locked in self-minting they would first need to buy synthetic tokens from "broker contract" before redeeming them for their full collateral - so they could still be stuck as "broker contract" could not sell them synthetic tokens due to pool being depleted.

UMIPs/umip-94.md
```diff
@@ -0,0 +1,90 @@
...
+Under certain conditions, limiting the cap has a downside: if there is not enough synthetic asset on the secondary market to buy them to liquidate a position, it is possible that the limit prevents minting enough synthetic to do so.
+
+4. Make the contract permissionless : the PerpertualPoolParty contract was permissioned (only whitelisted pools can be a Sponsor). We changed this so anyone can become a Sponsor, like it is now in any UMA derivatives of course.
+
+5. Remove the liquidity pool logic : the current PerpetualPoolParty contract works with liquidity pools.
+
+6. The derivatives parameters (CapDepositLimit, CapMintLimit and Fee) are now stored in a controller contract (`SelfMintingController.sol`): this allows us to change those parameters while keeping the same derivatives. 
+
+7. In the factory smart contract, we can only link the derivative contract to an existing synthetic asset, which have been previously deployed.
```
Reinis-FRP, line 59: Just to clarify, all derivative contracts (either created through `SynthereumDerivativeFactory.sol` or `SelfMintingDerivativeFactory.sol`) will hold the minter role for the linked synthetic assets (including already deployed jEUR, jGBP and jCHF)?

UMIPs/umip-94.md
Pascaltallarida: No the UMA team wanted for us to have a clear separation.

UMIPs/umip-94.md
Pascaltallarida: Yes, we changed the name because there is no interest fee (so borrowing was confusing) and we updated also "liquidity pools" for "broker". 

Yes it does solve this issue since it unlock liquidity and creates the issue you are mentioning; this is why we are creating secondary market (for now on Sushiswap, and most likely then on Uniswap v3, Kyber v3 and/or Balancer v2) as well as money markets from where to borrow synths.

But it is the same as any other protocol: if you mint yUSD with UMA collateral token to liquidate someone, the yUSD are burnt so you would need to buy them back off the market or mint new.

UMIPs/umip-94.md
Pascaltallarida: Yes, this is correct.

UMIPs/umip-94.md
Pascaltallarida: Ok, will add :)

UMIPs/umip-94.md
rollodex: How does the secondary market fix the issue described above? How is it different from creating a new token on UMA and listing on a secondary marketplace? Please clarify how it solves the original issue. Thanks

SayWhatDev: > How does the secondary market fix the issue described above? How is it different from creating a new token on UMA and listing on a secondary marketplace? Please clarify how it solves the original issue. Thanks

Having a secondary market would allow to avoid a situation where a liquidator has self-minted synthetic assets in order to perform a liquidation, but is unable to mint synthetic assets from the Broker contract (due to pool being depleted), since the liquidator would be able to purchase said synthetic asset directly from a secondary market (as ex. Uniswap, Kyber etc.) and burn those synthetic assets in order to unlock his collateral.

smb2796:
Last blocker before merging is adding the deployed contract addresses. Will wait on those and then merge!

smb2796 (approved):
Thanks for adding the contracts and congratulations on deploying! Going to merge in and move to `Last Call`.

PR #291, opened 2021-05-12
[UMIP-95] Add `DerivativeFactory` as `Creator` in DVM registry

SayWhatDev:


UMIPs/umip-95.md
```diff
@@ -0,0 +1,80 @@
...
+# List of deployed contracts:
+
+
+# Implementation and code base
+
+The `DerivativeFactory.sol` contract as well as all supporting contracts for the `PerpetualPoolParty.sol` contract (including `PerpetualPoolParty.sol` contract) can be found here are available for a review by anyone [here](https://gitlab.com/jarvis-network/apps/exchange/mono-repo/-/tree/feature/derivative-v2/libs/contracts/contracts/contracts/derivative/v2).
+
+The core contracts of the protocol like `Manager.sol` and `Deployer.sol` can be found and reviewed by anyone [here](https://gitlab.com/jarvis-network/apps/exchange/mono-repo/-/tree/feature/derivative-v2/libs/contracts/contracts/contracts/core).
+
+# Security considerations
```
akhanubis, line 69: Can a faulty PerpetualPoolParty deployed by Jarvis have any kind of repercusion on the UMA DVM? If I understand correctly, it can't, but it might be helpful to add a comment regarding the scope of a possible vulnerability in a Jarvis synth.

UMIPs/umip-95.md
```diff
@@ -0,0 +1,80 @@
...
+
+Reminder:  is a protocol to issue multi-collateralized synthetic fiat assets against liquidity pools. Liquidity pools hold USD-stablecoin such as USDC and are the sole Token Sponsor: a mint is a transaction where the liquidity pool self-mints a synthetic fiat with a collateral in USDC, and sell this synthetic fiat for USDC to the end-user, at the Chainlink price.
+
+The [UMIP-34](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-34.md) have been approved by the governance, allowing us to launch our synthetic assets collateralized by USDC on the mainnet ($1M TVL, $700k of synthetic assets minted; we have capped the amout of synthetic assets that can be minted until more audits are conducted; the protocol has received one full audit from Halborn, and a second one is undergoing with Ubik).
+
+In order to increase the liquidity, security and scalability of our synthetic assets we have decided to deploy a new `DerivativeFactory.sol` contract and re-deploy our existing Derivatives/Pools (jEUR/USDC, jGBP/USDC, jCHF/USDC) in order to lower some of the required parameters like starting GCR and Liquidation treshold.
```
akhanubis, line 19: typo: ~~treshold~~ threshold (3 occurences)

akhanubis (commented):
Looks good, and given that it has been discussed already with the UMA team, I dont see any reason not to give DerivativeFactory the Creator role.

UMIPs/umip-95.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Here is a breakdown on the whole deployment process of a new derivative:
+
+1. Our DAO address (which is currently a team address until the DAO is set up) calls `deployPoolAndDerivative` or `deployOnlyDerivative` of `Deployer.sol`.
+2. `Deployer.sol` calls `createPerpetual` function of `DerivativeFactory.sol` and this function will call `createPerpetual` function of the base contract `PerpetualPoolPartyCreator.sol` that will deploy the new `PerpetualPoolParty.sol`.
+
+### Modifications done: 
+
+- Removal of role management from the `PoolOnChainPriceFeed.sol` and `PerpetualPoolParty.sol`. This is an unnecessary feature since the role management is handled by the `Manager.sol` contract. 
+- Removal of `emergencyShutdown` functionality from the `PoolOnChainPriceFeed.sol` contract. Initially the `emergencyShutdown` can be called by each `PoolOnChainPriceFeed.sol`. In order to improve security of the protocol we have moved the `emergencyShutdown` functionality to be called only from the `Manager.sol` contract and only by authorized addresses - in our case the UMA DAO and Jarvis DAO.
```
Reinis-FRP, line 45: Are these modifications relative to the version approved in UMIP-34?
I noticed that contract naming now is different (in UMIP-34 all names were prepended with `Synthereum`)

UMIPs/umip-95.md
```diff
@@ -0,0 +1,80 @@
...
+
+Giving `Creator` role to `DerivativeFactory.sol` would allow more scalable and liquid synthetic assets to be backed by the UMA liquidation/dispute system, since every newly deployed derivative (`PerpetualPoolParty.sol`) is registered with the DVM.
+
+# Technical Specification
+
+### Here is a breakdown on the whole deployment process of a new derivative:
+
+1. Our DAO address (which is currently a team address until the DAO is set up) calls `deployPoolAndDerivative` or `deployOnlyDerivative` of `Deployer.sol`.
+2. `Deployer.sol` calls `createPerpetual` function of `DerivativeFactory.sol` and this function will call `createPerpetual` function of the base contract `PerpetualPoolPartyCreator.sol` that will deploy the new `PerpetualPoolParty.sol`.
+
```
Reinis-FRP, line 41: UMIP-34 also included calls to `createPool` function and linking the pool to `SynthereumPerpetualPoolParty.sol` - does it mean existing pools will not be redeployed? But how would it then be possible to launch new pools without this functionality?

UMIPs/umip-95.md
Pascaltallarida: It cannot. A vulnerability will only affect Jarvis, as everything is siloed. We will add that.

UMIPs/umip-95.md
SayWhatDev: > Are these modifications relative to the version approved in UMIP-34?
> I noticed that contract naming now is different (in UMIP-34 all names were prepended with `Synthereum`)

Hello, the `DerivativeFactory` approved in UMIP-34 has no logical differences from the current `DerivativeFactory` proposed in this UMIP. 

The derivative contracts being deployed by the factory has the only change that the role handling is now passed to the RoleManager contract instead of the pool contract and the emergencyShutdown can not be called by any linked pool, but instead only through the RoleManager. The whitelisted addresses that can call the emergencyShutdown on a derivative are the UMA DAO and Jarvis DAO.

The naming has changed in order to keep our versioning clear in the repository. Initially in V2 of the protocol we've used an off-chain price feed to agree on a price between the LP and the user on which a synthetic asset will be minted through a derivative. Although this version was not deployed on Mainnet, we've kept the code base in our repo for future reference. The V3 which was deployed using the DerivativeFactory approved in UMIP-34 uses an on-chain price feed to achieve the agreement between the LP and the user, thus we've changed the names of the contracts to correspond to that version.

UMIPs/umip-95.md
SayWhatDev: We still have the functionality addressed in UMIP-34 in particular the call to the `createPool`. Existing pools will be redeployed as we want to change some of the params set in them like the initial GCR and the liquidation threshold, however we've removed the explanation on pool creation as this logic is irrelevant to the DVM and the `DerivativeFactory`.

UMIPs/umip-95.md
Reinis-FRP: Thanks, just for clarity, can you please then put it more explicitly the version against which the modifications listed here are done

smb2796:
Same comment as #290. Waiting on the contract addresses and then will merge!

smb2796 (approved):
Moving to `Last Call`!

PR #292, opened 2021-05-13
Correct details in upunks and digg price id

smb2796:


cc7768 (approved):
This looks good to me -- You divided by 1,000 correctly :smile: 

The updated text also all makes sense.

PR #300, opened 2021-05-21
Update USDOCEAN price identifier based upon lessons learned in voting

smb2796:
This update does 3 things:

1. Removes Bittrex OCEAN/USDT in favor of Kraken OCEAN/USD
2. Updates the rounding spec of USDOCEAN from 18 decimals to 6 decimals
3. Clarifies that rounding should not happen in intermediate steps for USDOCEAN

See the discussion in the UMA Discord `voting` channel for more details


nicholaspai (approved):
LGTM, the 18 decimal precision was definitely an oversight

PR #301, opened 2021-05-21
Set up mainnet cross-chain oracle bridge

nicholaspai:


UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,56 @@
...
+- The `Bridge` passes the price request data to a "beacon" oracle called a `SourceOracle` which makes note of the L2 that submitted the request, before submitting a price request to the DVM.
+- Once the DVM resolves the price request, anyone can "publish" the result to the `SourceOracle` and specify which L2 it should communicate the resolved price to. This call simply queries the DVM for the price resolved and copies it to the `SourceOracle`, and therefore it cannot publish a different price than the one resolved.
+- On this same publish call, the `SourceOracle` submits a "deposit" to the `Bridge` and will emit a "Deposit" event containing price resolution metadata if the deposit is successful. Again, a trusted off-chain "relayer" will detect the "Deposit" event and submit a deposit to the `Bridge` contract on L2 with the resolved price metadata.
+- The `Bridge` on L2 will finally publish the resolved price to the `SinkOracle`, which means that the `OptimisticOracle` on L2 can now fetch a DVM-resolved price.
+
+## Technical Specification
+To accomplish this upgrade, one governance action must be taken:
```
mrice32, line 32: I totally forgot -- we also need to have the `Bridge` added to the Finder.

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Implementation
+We have implemented and deployed a [trusted bridge system](https://chainbridge.chainsafe.io/) on other EVM networks so that registered contracts, like `OptimisticOracles`, can submit price requests to "beacon" oracles that will ultimately relay the requests to mainnet via off-chain relayers. The bridge contract system was conceived by Chainbridge and has been cloned into the `UMAprotocol/protocol` repository [here](https://github.com/UMAprotocol/protocol/tree/master/packages/core/contracts/chainbridge).
+
+There are no DVMs (`Voting.sol` contracts) deployed to L2; instead we deploy a "beacon" contract called a [`SinkOracle`](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/chainbridge/SinkOracle.sol). The purpose of the `SinkOracle` is to receive price resolution data from the DVM on L1, and it can also send price requests to the DVM. A corresponding `SourceOracle` is deployed on L1 that parses price request information before communicating directly with the DVM. Generally, price resolution data flows from `SourceOracle` to `SinkOracle`, while price requests are bubbled up from the `SinkOracle` to the `SourceOracle`.
```
mrice32, line 18: tiny phrasing nit: I would slightly rephrase from:
```
The purpose of the `SinkOracle` is to receive price resolution data from the DVM on L1, and it can also send price requests to the DVM.
```

to 
```
The purpose of the `SinkOracle` is to send price requests to the L1 DVM and receive corresponding price resolution data from the L1 DVM.
```

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,56 @@
...
+- We anticipate that there will be 1 `SourceOracle` deployed to mainnet, and 1 `SinkOracle` deployed to each L2 network that needs to securely obtain prices from L1.
+- Each `SinkOracle` will have a unique `chainId` that it will submit with price requests to the `SourceOracle`. This effectively enables a unique communication channel between each `SinkOracle` and the 1 `SourceOracle`.
+
+## Technical Example: bridging a price request from L2 to L1: 
+- If an `OptimisticOracle` on L2 makes a price request, it will send it to the `SinkOracle`, which will submit a "deposit" to the `Bridge` contract on L2 that will emit a "Deposit" event containing the price request metadata if the deposit is successful. A trusted off-chain "relayer" will detect this "Deposit" event and submit a "deposit" to the `Bridge` contract on L1 with the same price request metadata. The L1 `Bridge` now has the price request information it needs to pass on to the DVM to submit a normal price request.
+- The `Bridge` passes the price request data to a "beacon" oracle called a `SourceOracle` which makes note of the L2 that submitted the request, before submitting a price request to the DVM.
```
mrice32, line 26: nit: are they both considered "beacon" oracles (they both inherit from the beacon oracle contract)?

mrice32 (commented):
Looks great!

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Summary (2-5 sentences)
+This UMIP will have the effect of introducing a new "beacon" oracle contract that will allow communication of price request data across EVM networks. There are two types of beacon oracles: the `SinkOracle` and the `SourceOracle`, deployed on L2 and L1 respectively.
+
+## Motivation
+The security of the DVM depends on UMA token holders voting on mainnet, rendering any price resolutions on mainnet as the final "source of truth". Therefore, any non-mainnet EVM network that wants to either submit or resolve price requests must be able to communicate with the mainnet DVM. 
+
+## Implementation
+We have implemented and deployed a [trusted bridge system](https://chainbridge.chainsafe.io/) on other EVM networks so that registered contracts, like `OptimisticOracles`, can submit price requests to "beacon" oracles that will ultimately relay the requests to mainnet via off-chain relayers. The bridge contract system was conceived by Chainbridge and has been cloned into the `UMAprotocol/protocol` repository [here](https://github.com/UMAprotocol/protocol/tree/master/packages/core/contracts/chainbridge).
```
pemulis, line 16: Are price requests only relayed from L2 to L1 if there is a dispute in the L2 optimistic oracle?

nicholaspai: Yea the idea is for the optimistic oracle on L2 to bubble requests up to the sink oracle on L2 for disputes. Most of the time we can resolve requests optimistically

UMIPs/beacon-oracle.md
nicholaspai: Ah yes and GenericHandler

UMIPs/beacon-oracle.md
nicholaspai: Yeah, this is admittedly a bit confusing so I removed all reference in this section to "beacon" oracles.

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,56 @@
...
+## Security considerations
+Please see the individual PRs below for details on how each affects the security of the UMA ecosystem. This repo is undergoing audit, and the final audit report will be added here once complete. We are comfortable deploying and using these contracts before the audit is complete because the off-chain relayer is trusted.
+
+The main security risk introduced is that the off-chain relayer has the power to:
+- modify price request data before sending from L2 to L1
+- modify price resolution data before sending from L1 to L2
+- submit spam price requests to the DVM on L1 by calling `Bridge.deposit`
+
+The secondary security risk is that anyone can trigger spam "Deposit" events and induce unneccessary relayer actions by:
+- calling `SourceOracle.publishPrice` repeatedly. This method can be called an unlimited amount of times so that if a publish process fails for any reason on L2, then the L1 caller can restart the process. We are still considering constraining the amount of times this method can be called.
```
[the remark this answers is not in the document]
nicholaspai: Removed this section.

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,56 @@
...
+- calling `SinkOracle.requestPrice` repeatedly. This method will revert only after a price is published from L1 to the `SinkOracle`.
+- The reason why we enable this right now is so as a fallback measure in case a cross-network communication fails on the second leg for some reason, and we need to restart the process beginning from emitting a "Deposit" event.
```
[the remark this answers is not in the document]
nicholaspai: Actually on further thought, `requestPrice` can only be called by a `RegisteredContract` so this is not an issue. I removed this whole "secondary issue" section. Yet another case of our past selves solved a problem that our current selves forgot about.

UMIPs/beacon-oracle.md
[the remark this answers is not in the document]
mrice32: This probably isn't super relevant to this UMIP, but can't you still spam this by calling deposit directly?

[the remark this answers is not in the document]
nicholaspai: Yep, fixing in this [PR](https://github.com/UMAprotocol/protocol/pull/3013)

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,54 @@
+## Headers
+| UMIP-XX     |                                                                                                                                          |
+|------------|------------------------------------------------------------------------------------------------------------------------------------------|
+| UMIP Title | Configure Cross Chain Oracle Bridge              |
+| Authors    | Nick Pai (nick@umaproject.org) |
+| Status     | Draft                                                                                                                                    |
+| Created    | May 21, 2021                                                                                                                           |
+
+## Summary (2-5 sentences)
```
mrice32, line 9: Should we add details to this document about how the governance integration works?

mrice32 (commented):
Just one question

UMIPs/beacon-oracle.md
nicholaspai: Yeah good call, just added in [this commit](https://github.com/UMAprotocol/UMIPs/pull/301/commits/d4ed475526dfe6cecffd8d449e4610b60d5e1b5e) curious what you think of the description

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,54 @@
...
+This UMIP will have the effect of introducing a new "beacon" oracle contract that will allow communication of price request data across EVM networks. There are two types of beacon oracles: the `SinkOracle` and the `SourceOracle`, deployed on L2 and L1 respectively.
+
+## Motivation
+The security of the DVM depends on UMA token holders voting on mainnet, rendering any price resolutions on mainnet as the final "source of truth". Therefore, any non-mainnet EVM network that wants to either submit or resolve price requests must be able to communicate with the mainnet DVM. 
+
+## Implementation
+We have implemented and deployed a [trusted bridge system](https://chainbridge.chainsafe.io/) on other EVM networks so that registered contracts, like `OptimisticOracles`, can submit price requests to "beacon" oracles that will ultimately relay the requests to mainnet via off-chain relayers. The bridge contract system was conceived by Chainbridge and has been cloned into the `UMAprotocol/protocol` repository [here](https://github.com/UMAprotocol/protocol/tree/master/packages/core/contracts/chainbridge).
```
Reinis-FRP, line 20: Better to be more explicit on which entity the "We" in the beginning of the statement refers to, especially, as this is a trusted model and the security of correct price resolution on L2 depends on the trusted bridging entity.

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,54 @@
...
+- We anticipate that there will be 1 `SourceOracle` deployed to mainnet, and 1 `SinkOracle` deployed to each L2 network that needs to securely obtain prices from L1.
+- Each `SinkOracle` will have a unique `chainId` that it will submit with price requests to the `SourceOracle`. This effectively enables a unique communication channel between each `SinkOracle` and the 1 `SourceOracle`.
+
+## Technical Example: bridging a price request from L2 to L1: 
+- If an `OptimisticOracle` on L2 makes a price request, it will send it to the `SinkOracle`, which will submit a "deposit" to the `Bridge` contract on L2 that will emit a "Deposit" event containing the price request metadata if the deposit is successful. A trusted off-chain "relayer" will detect this "Deposit" event and submit a "deposit" to the `Bridge` contract on L1 with the same price request metadata. The L1 `Bridge` now has the price request information it needs to pass on to the DVM to submit a normal price request.
+- The `Bridge` passes the price request data to the `SourceOracle` which makes note of the L2 that submitted the request, before submitting a price request to the DVM.
```
Reinis-FRP, line 30: Is chain ID and the requesting L2 contract address also stamped on to the DVM as ancillary data so that voters can identify the ultimate request source and judge whether or not this constitutes a spam attack?
Or instead the spam protection currently relies only on the L1 trusted relayer?

UMIPs/beacon-oracle.md
```diff
@@ -7,7 +7,11 @@
...
+This UMIP will have the effect of introducing a new "beacon" oracle contract that will allow communication of price request data across EVM-compatible networks. There are two types of beacon oracles: the `SinkOracle` and the `SourceOracle`, deployed on L2 and L1 respectively.
+
+The oracle contracts use Chainbridge to send messages between different chains. In particular, DVM price requests are passed from a `SinkOracle` on L2 to the `SourceOracle` on L1, which communicates with the DVM and resolves the price request via the standard mechanism. Once resolved, the price is pushed back to the `SinkOracle` chain where it can be utilized. This allows UMA financial contracts to be deployed in multiple different chains but still use the Ethereum mainnet DVM to resolve prices as required.
+
+The cross-chain infrastructure also includes contracts that pass governance proposals from the `Governor` contract on L1 to a `SinkGovernor` on L2. Governance proposals are similarly sent through the Chainbridge message bridge.
```
mrice32, line 14: We should probably mention the SourceGovernor too. Otherwise, this LGTM!

mrice32 (approved):

UMIPs/beacon-oracle.md
Reinis-FRP, line 30: Do I understand correctly that any fees paid stays in respective L2 `Store` contracts? How L1 DVM voters are expected to observe the fees paid in L2?

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Technical Specification
+To accomplish this upgrade, three governance actions must be taken:
+- The `SourceOracle` will need to be registered with the `Registry` so that it can make price requests to the DVM. The currently deployed `SourceOracle` can be found [here](https://etherscan.io/address/0x9eD925c04a6e391039f3362976F958e6934Ad0F4).
+- The `Bridge` contract address [here](https://etherscan.io/address/0xBA26bC014c4c889431826C123492861e886408b9) will need to be added to the `Finder` under the name "Bridge". This is how the `SourceOracle` will know which contract to call in order to relay messages to L2.
+- The `GenericHandler` contract address [here](https://etherscan.io/address/0x60E6140330F8FE31e785190F39C1B5e5e833c2a9) will need to be added to the `Finder` under the name "GenericHandler". This contract is a required middle layer for use by the `Bridge` to relay messages from L2 to the `SourceOracle`.
+
+## Rationale
+As described in the "Technical Example" section, the `SourceOracle` needs to be able to submit price requests to the DVM and only registered contracts in the `Registry` can do this.
```
Reinis-FRP, line 42: Is it possible to discuss any alternative bridging mechanisms considered?

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,54 @@
...
+## Security considerations
+Please see the individual PRs below for details on how each affects the security of the UMA ecosystem. This repo is undergoing audit, and the final audit report will be added here once complete. We are comfortable deploying and using these contracts before the audit is complete because the off-chain relayer is trusted.
+
+The main security risk introduced is that the off-chain relayer has the power to:
+- modify price request data before sending from L2 to L1
+- modify price resolution data before sending from L1 to L2
+- submit spam price requests to the DVM on L1 by calling `Bridge.deposit`
+- submit false price resolutions to L2 that did not originate from an L1 price resolution.
+
+Notably, the ChainBridge system enables the relayer system to eventually grow into a more decentralized federation of trusted relayers, but we anticipate the relayer set to be small to bootstrap the system.
```
Reinis-FRP, line 54: Still, this sounds to be a reputation based system, even when decentralized. Alternatively L2 protocols could rely on the same set of trusted relayers to provide price request data directly - why do they need UMA voters for this?
Or this is just a bootstrapping solution and in the long term it would still be possible to switch to a solution where relayers have some value at stake and consensus mechanism to correctly translate L1 price resolution to L2 networks?

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,54 @@
...
+
+There are no DVMs (`Voting.sol` contracts) deployed to L2; instead we deploy a "beacon" contract called a [`SinkOracle`](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/chainbridge/SinkOracle.sol). The purpose of the `SinkOracle` is to send price requests to the L1 DVM and receive corresponding price resolution data from the L1 DVM. A corresponding `SourceOracle` is deployed on L1 that parses price request information before communicating directly with the DVM. Generally, price resolution data flows from `SourceOracle` to `SinkOracle`, while price requests are bubbled up from the `SinkOracle` to the `SourceOracle`.
+
+The relationship between `SinkOracle` and `SourceOracle` is "N-to-1":
+- We anticipate that there will be 1 `SourceOracle` deployed to mainnet, and 1 `SinkOracle` deployed to each L2 network that needs to securely obtain prices from L1.
+- Each `SinkOracle` will have a unique `chainId` that it will submit with price requests to the `SourceOracle`. This effectively enables a unique communication channel between each `SinkOracle` and the 1 `SourceOracle`.
+
+## Technical Example: bridging a price request from L2 to L1:
```
Reinis-FRP, line 28: I found the [visual representation](https://docs.google.com/presentation/d/1sGvKMknFAhDf-tfd6EkEOijGMrXCieC2hIycIObAVGs/edit#slide=id.gd5fbfb2d7f_0_21) of proposed bridging architecture quite helpful. Maybe could link it here as well.

UMIPs/beacon-oracle.md
nicholaspai: Hm this is a really good point, maybe we should pass in the requester? The downside is passing an extra variable between contracts and through the Bridge, but allowing voters to know which requester sent the message could be very useful. @mrice32 thoughts?

nicholaspai: L1 DVM voters will know the canonical `Store` contract address on L2 so they can determine if a request paid the fees correctly. We could even implement voting dApps that show big WARNINGS if a price request seemingly did NOT pay fees on L2. WDYT Reinis?

UMIPs/beacon-oracle.md
nicholaspai: I'll add this to the UMIP, but worth raising a discussion here. In an ideal world, we would have used a canonical arbitrary message bridge maintained by the Polygon team and similar to the one described [here in the polygon docs](https://docs.matic.network/docs/develop/l1-l2-communication/state-transfer). However, it does not appear that the deployed tunnel/bridge contracts are maintained [here](https://github.com/jdkanani/fx-portal). So, for the sake of speed, we wanted to implement our own solution that could bridge DVM price requests to any EVM without an arbitrary message bridge.

So, this solution with a trusted relayer is suboptimal admittedly, but it will work as a proof of concept. The DVM can always remove a relayer if it gets compromised for any reason, or DVM voters can refuse to serve price requests sent by a malicious relayer. So in the worst case situation, the DVM can be secured pretty easily.

@mrice32 I didn't realize that matic does support a canonical tunnel system (see first link), but it doesn't look like they are supported. Thoughts?

nicholaspai: I am a bit surprised reading the latest matic docs because this [`L2 <> L1: State Transfer`](https://docs.matic.network/docs/develop/l1-l2-communication/state-transfer) section definitely wasn't there 1 month ago...

UMIPs/beacon-oracle.md
nicholaspai: This is definitely more of a bootstrapping solution. In the future we could pretty easily switch to a solution where the relayer network is a large decentralized one who must post bonds. If Polygon were to support an arbitrary message Bridge secured by their own set of validators then we'd definitely want to use that one.

nicholaspai: Do you think its worth changing the language a bit to emphasize that this is very much a proof of concept? Or do you think we should delay this until we have a more decentralized set of relayers, or identify an arbitrary message bridge.

UMIPs/beacon-oracle.md
nicholaspai: > Is chain ID and the requesting L2 contract address also stamped on to the DVM as ancillary data so that voters can identify the ultimate request source and judge whether or not this constitutes a spam attack?
Or instead the spam protection currently relies only on the L1 trusted relayer?

To answer your original question, the requesting chain ID is stored in the `Bridge` contract on L1, so this can be found within the same transaction that later sends the `requestPrice` call. This metadata can also be used to identify the L2 requester.

UMIPs/beacon-oracle.md
pemulis: If possible, I would love to see some more detail on the path to decentralization using ChainBridge, maybe a short summary and link to some ChainBridge docs that get into this.

UMIPs/beacon-oracle.md
Reinis-FRP: Looking at the current `Voting.sol` contract I now do realize that it does not log ancillary data in the `PriceRequestAdded` event even though it is passed as argument in the `requestPrice` function. So currently there is no added benefit to stamp additional information on chain ID to ancillary data as `Voting.sol` does not log it.

Voters should still be able to extract information on chain ID and requester when `SourceOracle` emits `PriceRequestAdded` event in the same transaction.

Still, I would assume this makes it more challenging to develop a Voting dApp as number of registered contracts allowed to make `requestPrice` calls to `Voting.sol` will increase over time. This will become more complicated as Voting dApp has to gather additional information by checking and interpreting logs emitted by calling contracts.

If there is a planned upgrade on `Voting.sol` in the future, maybe it could include ancillary data in the `PriceRequestAdded` event so that it is easier for voting dApp to interpret it? Though it also requires approving some standards on ancillary data stamp formatting for it to be practically useful.

Reinis-FRP: As currently price request bridging would operate under trusted relayer model I do not see the added benefit to include fee information in L1 price request as voters should still verify it manually on L2 `Store` contract.

However, if the bridging model becomes more decentralized in the future, it might actually make sense to pass on paid fee currency and value information by stamping it by L2 OO to the ancillary data. But this would require some standardization on the ancillary data stamping.

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,61 @@
...
+| Authors    | Nick Pai (nick@umaproject.org) |
+| Status     | Draft                                                                                                                                    |
+| Created    | May 21, 2021                                                                                                                           |
+
+## Summary (2-5 sentences)
+This UMIP will have the effect of introducing a new "beacon" oracle contract that will allow communication of price request data across EVM-compatible networks. There are two types of beacon oracles: the `SinkOracle` and the `SourceOracle`, deployed on L2 and L1 respectively.
+
+The oracle contracts use Chainbridge to send messages between different chains. In particular, DVM price requests are passed from a `SinkOracle` on L2 to the `SourceOracle` on L1, which communicates with the DVM and resolves the price request via the standard mechanism. Once resolved, the price is pushed back to the `SinkOracle` chain where it can be utilized. This allows UMA financial contracts to be deployed in multiple different chains but still use the Ethereum mainnet DVM to resolve prices as required.
```
daywiss, line 12: would be nice to have a link to more info on chainbridge in the first sentence

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Implementation
+We have implemented and deployed a [trusted bridge system](https://chainbridge.chainsafe.io/) on other EVM networks so that registered contracts, like `OptimisticOracles`, can submit price requests to "beacon" oracles that will ultimately relay the requests to mainnet via off-chain relayers. The bridge contract system was conceived by Chainbridge and has been cloned into the `UMAprotocol/protocol` repository [here](https://github.com/UMAprotocol/protocol/tree/master/packages/core/contracts/chainbridge).
+
+There are no DVMs (`Voting.sol` contracts) deployed to L2; instead we deploy a "beacon" contract called a [`SinkOracle`](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/chainbridge/SinkOracle.sol). The purpose of the `SinkOracle` is to send price requests to the L1 DVM and receive corresponding price resolution data from the L1 DVM. A corresponding `SourceOracle` is deployed on L1 that parses price request information before communicating directly with the DVM. Generally, price resolution data flows from `SourceOracle` to `SinkOracle`, while price requests are bubbled up from the `SinkOracle` to the `SourceOracle`.
```
daywiss, line 22: im confused as to what a `beacon` contract is. can this be defined more clearly and why is this terminology important?

UMIPs/beacon-oracle.md
```diff
@@ -0,0 +1,61 @@
...
+- The `Bridge` on L2 will finally publish the resolved price to the `SinkOracle`, which means that the `OptimisticOracle` on L2 can now fetch a DVM-resolved price.
+
+## Technical Specification
+To accomplish this upgrade, three governance actions must be taken:
+- The `SourceOracle` will need to be registered with the `Registry` so that it can make price requests to the DVM. The currently deployed `SourceOracle` can be found [here](https://etherscan.io/address/0xa37001be797637bb27f88f079607db2fdf264c24).
+- The `Bridge` contract address [here](https://etherscan.io/address/0xBA26bC014c4c889431826C123492861e886408b9) will need to be added to the `Finder` under the name "Bridge". This is how the `SourceOracle` will know which contract to call in order to relay messages to L2.
+- The `GenericHandler` contract address [here](https://etherscan.io/address/0x60E6140330F8FE31e785190F39C1B5e5e833c2a9) will need to be added to the `Finder` under the name "GenericHandler". This contract is a required middle layer for use by the `Bridge` to relay messages from L2 to the `SourceOracle`.
```
daywiss, line 39: the `generichandler` isnt described anywhere previously on what this does or why its needed

daywiss (approved):

nicholaspai:
Closing this PR in favor of a soon-to-be implemented contract that integrates with Polygon's [recently deployed arbitrary message bridge](https://docs.matic.network/docs/develop/l1-l2-communication/state-transfer/).

The `OptimisticOracle` remains live on Polygon (mainnet and testnet) and can be tested by integrators, but prices will be served by a trusted relayer still. Therefore we recommend that no one attempts to secure real value with the Polygon contracts yet.

PR #313, opened 2021-06-02
Create BASK Basket DAOs KPI options

Hahz48:
https://discourse.umaproject.org/u/JJ.C


BASK Basket DAOs KPI options
```diff
@@ -0,0 +1 @@
+
```
therealhexi, line 1: Is this PR submitted by mistake?

Hahz48:
Yes I believe another was submitted the right waySent from my Galaxy
-------- Original message --------From: Hexi ***@***.***> Date: 6/5/21  12:16 PM  (GMT-06:00) To: UMAprotocol/UMIPs ***@***.***> Cc: Hahz48 ***@***.***>, Author ***@***.***> Subject: Re: [UMAprotocol/UMIPs] Create BASK Basket DAOs KPI options (#313) 
@therealhexi commented on this pull request.



In BASK Basket DAOs KPI options:
> @@ -0,0 +1 @@
+

Is this PR submitted by mistake?

smb2796:
@Hahz48 your PR is empty, and the link is to JJC's discord profile. Can you update so that people can review?

I'm guessing this will need to be carried over to next week because of this, but we should fix asap! Let me know if you have any issues

Hahz48:
This was my attempt to get from discourse to github, which did not go so well. I also believe you guys are not going to do this one? I will fix the collateral one, with the suggestions you made.Sent from my Galaxy
-------- Original message --------From: Sean Brown ***@***.***> Date: 6/7/21  8:36 AM  (GMT-06:00) To: UMAprotocol/UMIPs ***@***.***> Cc: Hahz48 ***@***.***>, Mention ***@***.***> Subject: Re: [UMAprotocol/UMIPs] Create BASK Basket DAOs KPI options (#313) 
@Hahz48 your PR is empty, and the link is to JJC's discord profile. Can you update so that people can review?
I'm guessing this will need to be carried over to next week because of this, but we should fix asap! Let me know if you have any issues

BASK Basket DAOs KPI options
mhairimca: No, but I think there's been some difficulty in submitting it to github.

The original is on Discourse here 
https://discourse.umaproject.org/t/bdiflip-1221-as-a-supported-price-identifier/1121/14

Reinis-FRP: Unfortunately I could not figure out how to add comments to Discourse, so will drop them here:

- “Final fee of 6 BASK needs to be added to the store contract” – this belongs to collateral UMIP (PR #314)
- Rationale should also discuss the reasoning behind design decisions.
- `BASKPRICE-1221` in Implementation section not relevant as this is not used anywhere
- CoinGecko Live API is only available for DPI – please provide endpoint for BDI
- There is no historical data endpoint listed – how voters should be able to determine the price for requested timestamp?
- Is request timestamp already known for KPI options? Is it somehow related with 1221 in price identifier name (December 2021???)
- “rewards 1 BASK if the BDI market cap is greater than the DPI market cap” – not clear is it per 1 synthetic token minted? What is the planned amount of BASK collateral locked and amount of minted synths in total?
- Also not clear if voters should return just difference between two market caps in USD terms – will there be new financial contract library that will handle converting it to either 1 or 0?

smb2796:
Closing as this is stale, and I believe the intention is to use another price identifier being proposed.

PR #323, opened 2021-06-09
Add BINARY_QUERY price identifier

smb2796:


UMIPs/binary-query-pid.md
```diff
@@ -0,0 +1,74 @@
...
+| [Discourse Link](https://discourse.umaproject.org/t/add-binary-query-tbd-name-as-a-supported-price-identifier/1161)      |             |
+
+# Summary 
+
+The DVM should support price requests for the `BINARY_QUERY` price identifier. `BINARY_QUERY` is intended to be used with ancillary data to allow anyone to request an answer to a "yes or no" question from UMA governance. This UMIP does not attempt to put any other restrictions on the content of the query, and instead leaves construction of the query up to the requester within ancillary data.
+
+Settlement can happen in three ways:
+- Return the `p1` value from ancillary data if the answer is "NO".
+- Return the `p2` value from ancillary data if the answer is "YES".
+- Return the `p3` value from ancillary data if the answer cannot be determined.
```
mrice32, line 18: Will the voters scale these by 1e18 when returning the price?

UMIPs/binary-query-pid.md
```diff
@@ -0,0 +1,74 @@
...
+
+This construction sacrifices assurances of determinism in favor of greater price identifier flexibility. The places the burden of correct construction on price requesters, but, in return, allows for quicker and easier development without needing to pass through  UMA governance for each additional distinct query. This will allow for quite bespoke and speedy contract construction.
+
+# Implementation
+
+1. Voters should decode the ancillary data and attempt to interpret the UTF-8 question.
+2. If UMA voters believes that the answer to the question is no, they should vote return the p1 value (in the example given, they would return `0`).
+3. If UMA voters believe that the answer to the question is yes, they should return the p2 value (in the example given, they would return `1`).
+4. If a voter cannot make a determination about what the correct answer to the question is, or there is no question present, UMA voters should return 0.5.
+5. If there are no p1, p2, p3 values in the ancillary data voters should use 0, 1, 0.5 respectively.
```
mrice32, line 67: +1, love the defaults.

mrice32 (approved):
LGTM! One minor comment

UMIPs/binary-query-pid.md
```diff
@@ -0,0 +1,74 @@
...
+When this ancillary data dictionary is stored as bytes, the result would be: `0x7175657374696f6e3a446964207468652044616c6c6173204d6176657269636b732077696e20746865204e42412046696e616c7320696e20323032313f`
+
+# Rationale
+
+This construction sacrifices assurances of determinism in favor of greater price identifier flexibility. The places the burden of correct construction on price requesters, but, in return, allows for quicker and easier development without needing to pass through  UMA governance for each additional distinct query. This will allow for quite bespoke and speedy contract construction.
+
+# Implementation
+
+1. Voters should decode the ancillary data and attempt to interpret the UTF-8 question.
+2. If UMA voters believes that the answer to the question is no, they should vote return the p1 value (in the example given, they would return `0`).
```
mhairimca, line 64: Design issue rather than UMIP issue, perhaps, but could voters answer Yes, No and indeterminable on the Dapp and the values be filled in behind the scenes, or do they have to return numeric answers?

UMIPs/binary-query-pid.md
```diff
@@ -0,0 +1,74 @@
...
+
+When converted from bytes to UTF-8, the ancillary data should be a dictionary object containing q (question), p1, p2 and p3 keys and values:
+
+```
+q:Did the Dallas Mavericks win the NBA Finals in 2021?, p1:0, p2:1, p3:0.5
+```
+
+The q value should contain the question that UMA voters should answer. 
+
+When this ancillary data dictionary is stored as bytes, the result would be: `0x7175657374696f6e3a446964207468652044616c6c6173204d6176657269636b732077696e20746865204e42412046696e616c7320696e20323032313f`
```
Reinis-FRP, line 55: This decodes to "question:Did the Dallas Mavericks win the NBA Finals in 2021?"
Better update to `713a446964207468652044616c6c6173204d6176657269636b732077696e20746865204e42412046696e616c7320696e20323032313f2c2070313a302c2070323a312c2070333a302e35` in order to be consistent with the UTF-8 string above

smb2796: Woops. I updated the dictionary to take in p1, p2, p3 and just neglected to update this. Good catch, thank you.

UMIPs/binary-query-pid.md
smb2796: Conceivably we could filter for certain price identifiers and display those potential answers as "YES, NO, UNDETERMINABLE" in the voter dapp (similar to how we vote for UMIPs). But behind the scenes it would still have to return the p1, p2, p3 values. 

Personally I'd say this is more difficult than it is worth and potentially more confusing.

UMIPs/binary-query-pid.md
smb2796: Good question. I added this to the `Ancillary Data Specifications` section as an answer:

```
When returned, the `p1-p3` values should be scaled by 1e18. Typically voter dapps in the UMA ecosystem abstract this scaling away from voters, so, more often than not, voters should input the values as they appear in the ancillary data. Price requestors should be mindful of this, and not scale their ancillary data inputs.
```

smb2796:
Withdrawing from consideration this week and pushing to next. This due to a lack of urgency/no other umip votes, and would like to give the discussion on the community call about chained queries more thought.

mhairimca (approved):

CarelessDoge (approved):

PR #328, opened 2021-06-15
Add uDAO_KPI_UMA as a supported price identifier

britt-uma:
This is the draft for the next UMA KPI option based on the concept of integration. Known missing components at this time include: 
- link to future UMAverse dashboard
- alternative method to evaluate validity of integrations if the UMAverse dashboard is deemed too centralized (or if people just want to have a look for curiosity sake)

I also introduced the idea of DVM discretion (would like feedback), and set scaling in accordance with my best guess at expected integrations. I believe we will easily hit 12 and can reach 25 with a fair bit of confidence. Please feel free to challenge that number if you see fit.


Add uINT_KPI_UMA as a supported price identifier.md
```diff
@@ -0,0 +1,122 @@
...
+| **Created**             | *06/08/2021*                                             |
+| **Discourse Link**      | *Insert link to discourse topic  **after**  it has been moved into draft UMIPs_*            |
+
+# Summary 
+
+**The DVM should support requests for** UMA protocol integrations.
+
+Key Performance Indicator(KPI) options are synthetic tokens that redeem to protocol tokens with the redemption value determined by performance against that indicator. One example of a KPI metric is Integrations. Integrations serve to measure the activity of the protocol rather than assigning all of the value to the dollar amount generated by the protocol, as previously seen in UMA’s TVL KPI option. 
+
+This UMIP enables the DVM to support price requests based on the number of qualifying integrations that occur during a specified timeline.
```
therealhexi, line 18: Nit: `occurred`

Add uINT_KPI_UMA as a supported price identifier.md
```diff
@@ -0,0 +1,122 @@
...
+Using integrations as a key metric directly benefits the UMA protocol because it encourages the community to build relationships in the broader DeFi community with the intention of providing tools that may be useful to teams who are currently unaware of what UMA has to offer. This creates a win-win across the board because it allows us to showcase our products and our products can then help other protocols to complete their goals. 
+
+By establishing the terms of a qualifying integration, it allows for the community to reach out in a way that aligns with the current priorities of the UMA protocol- attracting attention to our strongest products and encouraging teams to test out a wide variety of use cases.
+
+
+**It is anticipated that this Price Identifier will be used to create KPI Options, however it is acknowledged that it may be used for a variety of purposes.**
+
+# Data Specifications
+
+Voters should assess the status of deployed contracts using (insert link to UMAverse dashboard?). For an integration to be displayed on the UMAverse dashboard, it must meet the following criteria:
```
therealhexi, line 36: link would be a must have before we mint this KPI option

Add uINT_KPI_UMA as a supported price identifier.md
```diff
@@ -0,0 +1,122 @@
...
+
+Any contract that falls into one of the categories listed below that has been funded with at least $10,000 of value. 
+* Call or Put options
+* KPI options
+* Range Bonds
+* Integration with Optimistic Oracle
+* Long/Short Pair (LSP) Contract
+
+Note: a contract that has been launched and funded by UMA will not count. However, a contract launched by UMA but funded by an external party does count.
```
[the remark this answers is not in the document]
therealhexi: +1, what is the definition of "funded" ?

Add uINT_KPI_UMA as a supported price identifier.md
```diff
@@ -0,0 +1,122 @@
...
+By establishing the terms of a qualifying integration, it allows for the community to reach out in a way that aligns with the current priorities of the UMA protocol- attracting attention to our strongest products and encouraging teams to test out a wide variety of use cases.
+
+
+**It is anticipated that this Price Identifier will be used to create KPI Options, however it is acknowledged that it may be used for a variety of purposes.**
+
+# Data Specifications
+
+Voters should assess the status of deployed contracts using (insert link to UMAverse dashboard?). For an integration to be displayed on the UMAverse dashboard, it must meet the following criteria:
+
+Any contract that falls into one of the categories listed below that has been funded with at least $10,000 of value.
```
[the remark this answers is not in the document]
therealhexi: Sharing my 2cents:

-> Do only Ethereum mainnet contracts count or any other chain where DVM message bridging is available?
All contracts should count including ones on other chain (e.g. Polygon) where there is a bridge.

-> What counts as "funding" - is it only collateral tokens locked as part of regular create and deposit functions (considering that its possible to send tokens to the contract by mistake)?
I would imagine we could use the TVL metrics of that particular token type to define funding.

The rest are great questions

Add uINT_KPI_UMA as a supported price identifier.md
```diff
@@ -0,0 +1,122 @@
...
+
+Step 5: return the appropriate value (in UMA) per uINT that corresponds to the total number of points tallied. If value returned is greater than 1.2000, return 1.2000.
+
+Example: if a voter establishes a tally of 12 points, they would say that 1uINT = (5 x 0.0167) + (7 x 0.0583) = 0.4917 UMA. 
+
+# Security Considerations
+
+How could metric manipulation occur?
+
+* By establishing a minimum value for qualifying integrations, it ensures that the cost of manipulation is higher than the possible reward for doing so. Additionally, the DVM voters have the ability to determine if it appears that someone is acting maliciously and can disqualify any integrations that majority believe should not be included.
```
therealhexi, line 104: How could we actually ensure the cost of manipulation is higher than the possible reward? What's the total amount of this KPI Option UMA plans to mint? What would the maximum payout be? And how much would it cost for a bad actor to even that maximum payout?

Add uINT_KPI_UMA as a supported price identifier.md
```diff
@@ -0,0 +1,122 @@
...
+
+How could this price ID be exploited?
+
+* I’m not sure it can be, since it is based on the definition of the metric. Thoughts?
```
therealhexi, line 108: This will be answered as the previous question gets answered.

Add uINT_KPI_UMA as a supported price identifier.md
```diff
@@ -0,0 +1,122 @@
...
+	
+Are there any current or future possible concerns with the way the price identifier is defined?
+
+* It could be determined that the scaling is not appropriate in the event of overwhelming failure of the option. This would happen if we achieved too few integrations or if we had an unexpected number of integrations from one particular team. The latter can be remedied by the voting power of the DVM, as discussed above. 
+* Additionally, there is the potential for ambiguities to present that have not been considered at the time of writing. In this case it will be up to the DVM to determine the resolution of those ambiguities. We anticipate hitting the ceiling, so it is possible that any ambiguities that arise will be non-significant. 
+* It is possible that over the course of the KPI option, UMA could determine there are other products they would like to prioritize that are not defined in this UMIP. In that case, another KPI option can be launched to include that product.
+
+Are there any concerns around if the price identifier implementation is deterministic?
+
+* No?
```
therealhexi, line 122: Should be no concern if we clearly define what "funded" mean, and which timestamp it applies.

therealhexi (commented):
Initial review

Add uINT_KPI_UMA as a supported price identifier.md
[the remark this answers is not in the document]
pemulis: `Should the contract be registered to the DVM?`

I think no, since LSPs won't be registered to the DVM, or any other contracts that only talk to the Optimistic Oracle.

`Is there any guidance on how to price locked collateral tokens (are stablecoins treated 1USD, should voters use established price identifiers if available, or leave this for human judgement)?`

I think human judgment is fine! It should be pretty clear if the threshold was hit or not.

Add uINT_KPI_UMA as a supported price identifier.md
```diff
@@ -0,0 +1,122 @@
...
+
+# Data Specifications
+
+Voters should assess the status of deployed contracts using (insert link to UMAverse dashboard?). For an integration to be displayed on the UMAverse dashboard, it must meet the following criteria:
+
+Any contract that falls into one of the categories listed below that has been funded with at least $10,000 of value. 
+* Call or Put options
+* KPI options
+* Range Bonds
+* Integration with Optimistic Oracle
```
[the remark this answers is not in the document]
pemulis: I think it might be good to leave this open-ended since there are many possible integrations with the OO. I think a contract that is designed to call on the OO for an important part of its functionality would fit the bill.

Add uINT_KPI_UMA as a supported price identifier.md
britt-uma: Noted. Thanks for the nit!

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,142 @@
...
+Using integrations as a key metric directly benefits the UMA protocol because it encourages the community to build relationships in the broader DeFi community with the intention of providing tools that may be useful to teams who are currently unaware of what UMA has to offer. This creates a win-win across the board because it allows us to showcase our products and our products can then help other protocols to complete their goals. 
+
+By establishing the terms of a qualifying integration, it allows for the community to reach out in a way that aligns with the current priorities of the UMA protocol- attracting attention to our strongest products and encouraging teams to test out a wide variety of use cases.
+
+
+**It is anticipated that this Price Identifier will be used to create KPI Options, however it is acknowledged that it may be used for a variety of purposes.**
+
+# Data Specifications
+
+The UMA team will maintain a "green light" list of qualified integrations. Each integration will be added one at a time, after it has been discussed in the #voting channel in the UMA discord. The following criteria will be considered during the evaluation of each integration:
```
claytonroche, line 36: I might say:

"Anyone can propose a product to be included in the integration list. The UMA team will take the lead with this by proposing and discussing in Discord."

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,142 @@
...
+
+This UMIP is meant to identify integrations with UMA that are good faith experiments and product launches, as a way to gauge DeFi community engagement with the protocol. This should not be considered an all-inclusive list, but here are some components to take into account when determining whether or not something is a good faith integration:
+
+**Planning:** Has there been contact with someone in the UMA community?
+What kind of planning stages occurred prior to launch? Was there a proposal from the super uman community? Was there an internal vote within a community to move forward with a launch? Were there brainstorming sessions to come up with possible use cases for various products? 
+
+**Intention:** WHAT is the goal? Has it been made clear that there is a desire to use UMA infrastructure to launch a product that will serve to improve a protocol or for the purpose of experimental hedging/speculation/betting?
+
+**Rationale:** WHY was it executed the way it was? (Note: struggling to find more words to say here, although I do believe this is the biggest factor in determining the good faith nature of a launch.)
```
claytonroche, line 62: I might want to say something like: "Was this product designed in such a way that it could or did work? Was it being given a chance at success?"

Add uINT_KPI_UMA as a supported price identifier v2.md
pemulis: Maybe: `The UMA team will maintain a "green light" list of qualified integrations, and anyone can propose a product in the integration list.` Followed by the rest of the paragraph?

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,142 @@
...
+
+Contracts included are not limited to Ethereum mainnet- contracts on other chains where DVM message bridging is available (e.g. Polygon) will be included in consideration.
+
+Note: Contracts deployed by UMA, for use by UMA do not count. 
+
+The applied timeframe for this UMIP will be between 1622505600 and 1633046399, meaning that all qualifying integrations launched during this timeframe should be counted. 
+
+**Revolution Clause** 
+
+The "green light" list should not be considered the *only* source of truth. The DVM has the ability and responsibility to challenge the validity of any integration that has been included or excluded from the list if the broader consensus of UMA tokenholders is that it is within the spirit of this UMIP to do so. Any and all questions surrounding the validity of an integration can be brought up in further detail in the voting channel on the UMA discord.
```
pemulis, line 53: `#voting channel`

pemulis: `Discord`

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,142 @@
...
+Key Performance Indicator(KPI) options are synthetic tokens that redeem to protocol tokens with the redemption value determined by performance against that indicator. One example of a KPI metric is Integrations. Integrations serve to measure the activity of the protocol rather than assigning all of the value to the dollar amount generated by the protocol, as previously seen in UMA’s TVL KPI option. 
+
+This UMIP enables the DVM to support price requests based on the number of qualifying integrations that occured during a specified timeline.
+
+A synthetic option is minted against a base collateral, in this case UMA, which expires at 23:59:59 (UTC) September 30th 2021. 
+
+
+# Motivation
+
+The primary motivation for the development of KPI options is to allow protocols to incentivize Defi users to assist them to reach the protocol's identified targets by leveraging their community resources by sharing their value with their community members.
```
pemulis, line 25: `DeFi`

Add uINT_KPI_UMA as a supported price identifier v2.md
pemulis: I like that phrasing.

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,142 @@
...
+* Someone creates and funds a call option for one month. When it expires, they then create another one for the next month. This would count as two integrations, provided that the intention and lifespan make sense for this product. 
+
+* An unidentified address funds a contract that was launched by UMA as a proof-of-concept for another team. The contract was funded in such a way to meet the minimium threshold, but there was no coordination or acceptance by the other team, and the product does not appear to be used by the intended project. This would not count as an integration, because even though the value threshold was met, there was no demonstration of planning, and intention is not clear.
+
+
+# Technical Specifications
+
+- **Price identifier name:**  *uINT_KPI_UMA*
+- **Base: Number of qualified integrations
+- **There is no quote currency in this option, as design feature. The collateral redemption is tied to the** *number of qualified integrations* **by design**
```
pemulis, line 79: `as a design feature`

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,142 @@
...
+* Choosing a starting timestamp of June 1 helps to reward the community for the integrations they have already started on and help keep momentum going while the current KPI expires. 
+
+* In order to reward the extra work that goes into larger integrations, any qualifying integration valued at or above $5 million will qualify for a bonus point that will increase the value of uINT by 0.0833 UMA. 
+
+
+# Implementation
+
+Step 1: Refer to "green light" list for finalized information on all qualified integrations. Assign 1 point for each integration.
+
+Step 2: Tally up the total number of points, and use the following criteria to determine the $UMA value of each uINT token (uINT/UMA). Note: any integrations with a (____) indicator qualify for a BONUS. See details below.
```
pemulis: Is `(____)` literally the indicator or is this a TBD?

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,142 @@
...
+
+Starting at 0, the value of each uINT increases according to the following scale:
+* 0.0167 for the first 10 points.
+* 0.0500 for the next 15 points.
+* A BONUS value of 0.0833 should be added for each integration on the "green light" list with a (____) indicator. Treat this as a separate component of your calculation. See example below. 
+
+Step 5: return the appropriate value (in UMA) per uINT that corresponds to the total number of points tallied. If value returned is greater than 1.2500, return 1.2500.
```
pemulis, line 111: Looked like a skip from Step 2 to Step 5.

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,142 @@
...
+
+Examples: 
+If a voter determines a tally of 12 points and no BONUS points, they would calculate that 1 uINT = (10 x 0.0167) + (2 x 0.0500) = 0.2667 UMA. 
+
+If a voter determines a tally of 22 points and 2 of them include BONUS points, they would calculate that 1 uINT = (10 x 0.0167) + (12 x 0.0500) + (2 x 0.0833) = 0.9333 UMA.
+
+If a voter determines a tally of 20 points and 7 of them include BONUS points, they would calculate that 1 uINT = (10 x 0.0167) + (10 x 0.0500) + (7 x 0.0833) = ??? (come back to this after discussion)
```
pemulis: Why is this `???`?

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,142 @@
...
+
+# Data Specifications
+
+The UMA team will maintain a "green light" list of qualified integrations. Each integration will be added one at a time, after it has been discussed in the #voting channel in the UMA discord. The following criteria will be considered during the evaluation of each integration:
+
+Any contract that falls into one of the categories listed below that has been funded with at least $10,000 of value. 
+* Call or Put options
+* KPI options
+* Range Bonds
+* Integration with Optimistic Oracle
```
pemulis, line 42: Would yield dollar EMPs count if they integrate with the Optimistic Oracle?

Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: Good catch, thanks!

Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: noted

Add uINT_KPI_UMA as a supported price identifier v2.md
smb2796: Good point. If not, it might be useful to define an additional list for contract types that do not count

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+
+
+# Implementation
+
+Step 1: Refer to "green light" list for finalized information on all qualified integrations. Assign 1 point for each integration.
+
+Step 2: Tally up the total number of points, and use the following criteria to determine the $UMA value of each uINT token (uINT/UMA). Note: any integrations with a 🐋 indicator qualify for a BONUS. See details below.
+
+Starting at 0, the value of each uINT increases according to the following scale:
+* 0.0167 for the first 10 points.
```
smb2796, line 107: These seem like weird splits to me. If I'm understanding this correctly all of these were maxed, the value would be 1.2502. Could we make these rounder numbers to make things simpler/cleaner?

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+* 0.0500 for the next 15 points.
+* A BONUS value of 0.0833 should be added for each integration on the "green light" list with a 🐋 indicator, with a maximum of 4 possible bonuses. Treat this as a separate component of your calculation. See example below. 
+
+Step 5: return the appropriate value (in UMA) per uINT that corresponds to the total number of points tallied. If value returned is greater than 1.2500, return 1.2500.
```
smb2796, line 111: Seems like if the values above are defined a little bit differently, this step would be unnecessary.

Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+
+Examples: 
+If a voter determines a tally of 12 points and no BONUS points, they would calculate that 1 uINT = (10 x 0.0167) + (2 x 0.0500) = 0.2667 UMA. 
+
+If a voter determines a tally of 22 points and 2 of them include BONUS points, they would calculate that 1 uINT = (10 x 0.0167) + (12 x 0.0500) + (2 x 0.0833) = 0.9333 UMA.
+
+If a voter determines a tally of 20 points and 7 of them would individually qualify for BONUS points, they would calculate that 1 uINT = (10 x 0.0167) + (10 x 0.0500) + (4 x 0.0833) = 1.000
```
smb2796, line 118: To confirm, the bonus points are double counted?

Add uINT_KPI_UMA as a supported price identifier.md
```diff
@@ -0,0 +1,122 @@
+## Headers
```
smb2796, line 1: @britt-uma looks like you are submitting two separate files. The one above looks like it is more updated. Can you remove one of these?

Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: The splits are weird, yes. The reason being is that they sort of rely on a lot of specifics that were asked for. Some of it comes from the number of tokens wanted to give out, the desired payout (in $USD), the assumption of the value of $UMA/USD, and how many integrations we wanted to aim for based on the starting point we feel best about. If it helps at all, here is a link to the google sheet we have been working on to set our scale up. [https://docs.google.com/spreadsheets/d/12i_KK95z9KPKM4qkQDenIBL1sUKc0CYV/edit#gid=2030671222](url)

Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: I agree, and I'm happy to reword the steps above if there is an easier way to add clarity. It was sort of residual from the last set of instructions, and I left it in there as a reminder that there is a ceiling.

Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: Yes. Counted once as a normal integration, then after all is said and done, bonus points get added where appropriate.

Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: This was one of the bold items in the template, so I didn't change it at all, but I thought it sounded weird too. Would you recommend leaving it or fixing it?

Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: This is something I think would be good to discuss on the community call. In short- I don't have a solid feeling for whether or not they should count. Feel free to give thoughts!

Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: fixed, thanks!

Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: Added this an a little more. Thanks!

Add uINT_KPI_UMA as a supported price identifier.md
britt-uma: Apologies, still learning git hub. I believe I was able to do that, but let me know if not!

Add uINT_KPI_UMA as a supported price identifier v2.md
claytonroche: Yep I like that John

Add uINT_KPI_UMA as a supported price identifier v2.md
claytonroche: My opinion: They should _not_ count. This opinion comes from our strategic goals of moving away from them.

How do we make them not count? That's a little tough. We could say that "Integrations must be of the type that are fully capitalized by design, such as possible with the LSP contract or certain parameterization of the EMP contract. However, other outside parties may use the OO, and their contracts can be designed to work in any fashion--and that counts as an integration.

A yield dollar or synthetic asset that allows for a partially capitalized position would not qualify. The rationale for this is to keep the product development aligned with UMA's strategic goals."

Add uINT_KPI_UMA as a supported price identifier v2.md
claytonroche: She wants to use 🐋

Britt's drafts/Add MPH, APW, SNOW, NDX as collateral currencies.md
```diff
@@ -0,0 +1,139 @@
+**UMIP #**  - tbd
```
smb2796, line 1: @britt-uma looks like this UMIP should be in a separate PR.

britt-uma: Trying to do that, not working out for me. Sorry!

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+* In order to reward the extra work that goes into larger integrations, any qualifying integration valued at or above $5 million in TVL for any amount of time will qualify for a bonus point that will increase the value of uINT by 0.0833 UMA. 
+
+
+# Implementation
+
+Step 1: Refer to "green light" list for finalized information on all qualified integrations. Assign 1 point for each integration.
+
+Step 2: Tally up the total number of points, and use the following criteria to determine the $UMA value of each uINT token (uINT/UMA). Note: any integrations with a 🐋 indicator qualify for a BONUS. See details below.
+
+Starting at 0, the value of each uINT increases according to the following scale:
```
smb2796, line 106: I wonder if it would be more appropriate to put some of these parameters (and the 10k integration threshold) into ancillary data. Would allow for greater flexibility, and I don't see any downsides except perhaps a little additional UMIP complexity.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+* 0.0167 for the first 10 points.
```
smb2796, line 107: I had left this comment before - but I think we should revisit these amounts. There's an inherit max value, when structured like this, that removes the necessity of a price cap. These numbers don't add up to the cap though (slightly above if I'm doing my math correctly) which seems unnecessarily complex.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+
+Due to the bespoke nature of this UMIP, it should not be considered a standard against which futuer UMIPs can me measured without considering the amount of risk in establishing a set of rules as qualitative as is the case here. When assessing the risk of using qualitative data to determine an outcome, several things have been considered, such as the intended "in house" use of this UMIP, the intended recipients of the uINT token, the intended outcome of the KPI option, and how to handle any ambiguities that could arise over the course of this option. 
+
+How could metric manipulation occur?
+
+* By establishing a discussion system for determining qualitative validity of an integration, it reduces the ability of any individual to game the system. The DVM voters have the ability to determine if it appears that someone is acting maliciously and can disqualify any integrations that majority believe should not be included.
+
+How could this price ID be exploited?
+
+* The definition of the metric is inherently qualitative, which makes it inprobable that it will be exploited.
```
smb2796, line 131: nit: but doesn't a qualitative UMIP make it more likely to be manipulated? Interested in your thoughts here

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+
+# Summary 
+
+**The DVM should support requests for** UMA protocol integrations.
+
+Key Performance Indicator(KPI) options are synthetic tokens that redeem to protocol tokens with the redemption value determined by performance against that indicator. One example of a KPI metric is Integrations. Integrations serve to measure the activity of the protocol rather than assigning all of the value to the dollar amount generated by the protocol, as previously seen in UMA’s TVL KPI option. 
+
+This UMIP enables the DVM to support price requests based on the number of qualifying integrations that occured during a specified timeline.
+
+A synthetic option is minted against a base collateral, in this case UMA, which expires at 23:59:59 (UTC) September 30th 2021.
```
smb2796, line 20: So this is when minting occurs? Not when the contract is launched?

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+
+# Data Specifications
+
+The UMA team will maintain a "green light" list of qualified integrations, which will be readily available upon request. Each integration will be added one at a time, after it has been discussed in the #voting channel in the UMA Discord. The following criteria will be considered during the evaluation of each integration:
+
+Any contract that falls into one of the categories listed below.
+* Call or Put options
+* KPI options
+* Range Bonds
+* Integration with Optimistic Oracle
```
smb2796, line 42: This gets a little complex for OO integrations where it might be difficult to determine when a contract was launched, minting occurred or how much value was in the contract. Also OO integrations probably wont fit as smoothly into the UMAverse whitelist. Perhaps we should base this off of price requests. Not sure, so we should discuss.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
chrismaree: does this need to be coupled to a particular settlement time?

britt-uma: This was copy/paste text from the previous KPI option- my understanding was that it just specified what the intended collateral was and when the synth would expire. If that's not clear or needs to be reworded (or even taken out) then I'm happy to do that

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+
+
+**It is anticipated that this Price Identifier will be used to create KPI Options, however it is acknowledged that it may be used for a variety of purposes.**
+
+# Data Specifications
+
+The UMA team will maintain a "green light" list of qualified integrations, which will be readily available upon request. Each integration will be added one at a time, after it has been discussed in the #voting channel in the UMA Discord. The following criteria will be considered during the evaluation of each integration:
+
+Any contract that falls into one of the categories listed below.
+* Call or Put options
```
chrismaree, line 39: is there any requirements for TVL locked in these options? what stops me from making 1000 call options to classify as an "integration"

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+By establishing the terms of a qualifying integration, it allows for the community to reach out in a way that aligns with the current priorities of the UMA protocol- attracting attention to our strongest products and encouraging teams to test out a wide variety of use cases.
+
+
+**It is anticipated that this Price Identifier will be used to create KPI Options, however it is acknowledged that it may be used for a variety of purposes.**
+
+# Data Specifications
+
+The UMA team will maintain a "green light" list of qualified integrations, which will be readily available upon request. Each integration will be added one at a time, after it has been discussed in the #voting channel in the UMA Discord. The following criteria will be considered during the evaluation of each integration:
+
+Any contract that falls into one of the categories listed below.
```
chrismaree, line 38: is there a reason that EMP/perp are explicitly ignored from this list? also how about linear long short pairs?

chrismaree: Call or put options, KPI options and range bonds are just a subset of the LSP contract. would it not be better to just say "use of the LSP contract"?

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
chrismaree: this is true with all LSP flavours as well where no registration needs to happen due to the OO being used for price settlement

britt-uma: I agree about the complexity of defining what the OO integration looks like. I have to confess that I don't know enough about how it works to contribute a meaningful opinion here. All I know is that we want to represent it somehow and we want to represent it specifically for the product types we care most about. Below are Clayton's two cents on the subject:

My opinion: They should not count. This opinion comes from our strategic goals of moving away from them.

How do we make them not count? That's a little tough. We could say that "Integrations must be of the type that are fully capitalized by design, such as possible with the LSP contract or certain parameterization of the EMP contract. However, other outside parties may use the OO, and their contracts can be designed to work in any fashion--and that counts as an integration.

A yield dollar or synthetic asset that allows for a partially capitalized position would not qualify. The rationale for this is to keep the product development aligned with UMA's strategic goals."

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+* KPI options
+* Range Bonds
+* Integration with Optimistic Oracle
+* Long/Short Pair (LSP) Contract
+
+Contracts included are not limited to Ethereum mainnet- contracts on other chains where DVM message bridging is available (e.g. Polygon) will be included in consideration.
+
+Note: Contracts deployed by UMA, for use by UMA do not count. 
+
+The applied timeframe for this UMIP will be between 1622505600 and 1633046399, meaning that all qualifying integrations launched during this timeframe should be counted.
```
chrismaree, line 49: can you please put human-readable conversions of this as well as the unix timestamps?

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
chrismaree: I see you answer this below :)

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,145 @@
...
+* A contract was launched by UMA as a proof-of-concept for another team. An unidentified address funds a contract, but there was no coordination or acceptance by the other team, and the product does not appear to be used by the intended project. This would not count as an integration, because even though the value threshold was met, there was no demonstration of planning, and intention is not clear.
+
+
+# Technical Specifications
+
+- **Price identifier name:**  *uINT_KPI_UMA*
+- **Base: Number of qualified integrations
+- **There is no quote currency in this option, as design feature. The collateral redemption is tied to the** *number of qualified integrations* **by design**
+ - **Intended Collateral Currency** *UMA*.
+- **Rounding:** *Round to 4 decimal places*
```
chrismaree, line 81: Surely this should be a whole number? i.e when can you have 0.5 intergrations

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: My thoughts: I suppose that depends on the quality of the group making the decisions. In this case, I think opening it up for greater discussion in the DVM community means that the results will be a strong reflection of the spirit of the umip. I think black and white parameters make something prone to manipulation because a bad egg can find a loophole. When it's up for interpretation, there's more opportunity to weed out things that don't belong.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: Let's chat about it on the call. Did you see the link I left before?

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: I think that could be a good discussion point if you have a proposal of how to do so.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: Yes, it was excluded because we're trying to promote/move towards products that are not at risk of liquidation. Aka, things that don't require us to trust deployers to run bots appropriately. 

Are linear LSP a different category than LSP? I would assume they are included

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: yes, good point. Will put in next edit. Thanks!

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: You are correct that you can't have a non-whole number integration. The rounding is for when calculating the value of uINT based on the number of integrations. If I have inappropriately placed that rounding info let me know and I can move it.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: There is no TVL requirement because each integration is being evaluated individually to determine the validity of it. In other words, you can try to do that but if you don't have a good reason as to WHY, then it will only count as one integration. This is part of the qualitative feature of this UMIP- it's all about intention.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
chrismaree: Added to this, if this financial product aims to use the LSP as the building block to launch the contract I'm not sure exactly how this would fit into that pattern. Ideally, you should be building an identifier that will work with one of our existing LSP financial product libraries (see [github](https://github.com/UMAprotocol/protocol/tree/master/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries)). One simplified design could work something like this:

- Use the linear library you define a lower bound and an upper bound
- set the lower bound to be 0 integrations and the upper bound to be 100 integrations
- any integration with more than 5m TVL equal 5 integrations to give the "bonus"
- at settlement, the OO simply returns this number of integrations and the financial product library translates this, in combination with the bounds, to the number between 0 and 1 used to settle the financial product library.

If built this way the identifier is answering a specific question: _how many integrations did UMA get over this time period_ rather than _how many uma should be paid out based on the integrations over this time period._

The key difference is that the OO and the DVM should be answering a _pricing_ question, not a payout question. the payout is left to the financial product in question.

Using ancillary data, the deployer can also define the period to _start_ looking at integrations for the given financial product. i.e consider integrations from 1 June to contract settlement.

britt-uma: I am open to more discussion on this, as I think it could potentially add simplicity. But there are a couple of notes from Clayton that may make that a challenge:

- We want to assign different values for the first x number of integrations, because we are back dating to June 1. These are considered slam dunk integrations, and the intention here is to recognize previous efforts. But we do want to put MORE weight on integrations that go from start to finish after the launch date.
- We want to cap the number of bonuses available without changing the expected number of integrations we hope to achieve. If a bonus counts as 5 integrations, that's setting the expectation that we will have those, rather than have them feel like a bonus. 

I do agree that the question should be about pricing rather than payout. The intended question to be answered by the DVM should be "What is the value of uINT/UMA based on the number of qualified integrations?" I only supplied the payout context to help articulate why the values are kinda wonky when calculating uINT. 

Let me know if that all makes sense and what your thoughts are.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,129 @@
...
+- **Base: Number of qualified integrations
+- **There is no quote currency in this option, as design feature. The collateral redemption is tied to the** *number of qualified integrations* **by design**
+ - **Intended Collateral Currency** *UMA*.
+- **Rounding:** *Round to nearest whole number* 
+ - If the value returned is greater than 15, round down to 15. (need to determine what the new goal is for number of integrations)
+ 
+
+# Rationale
+
+Specifics of this design were chosen to establish a a target number of meaningful integrations that we believe is within reach.  Adding a TVL minimum helps to align the desires of the UMA team and the UMA community to establish new and meaningful relationships with other communities with a real desire to use UMA products.
```
Reinis-FRP, line 86: "establish a a target" -> "establish a target"

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,129 @@
...
+
+* The decision to stick to the listed product types was made in order to bring focus to the products we want to push the most at this time.
+
+* Setting it up to allow for all qualified launches to be counted throughout the period rather than just at the time of expiry helps to ensure that we are not missing out on any integrations that might expire before the KPI option does. 
+
+* Choosing a starting timestamp of June 1 helps to reward the community for the integrations they have already started on and help keep momentum going while the current KPI expires. 
+
+# Implementation
+
+Step 1: Refer to "green light" list for finalized information on all qualified integrations. Assign 1 point for each integration. If an integration qualifies for a bonus, it will be indicated with a 🐋. These integrations are worth 3 points each (This number should be discussed prior to vote)
```
Reinis-FRP, line 100: Now it is not clear what are the parameters for an integration to qualify for 🐋? Is this left at discretion of UMA team by design? If so, it would be better to describe it more explicitly.

What is meant by "This number should be discussed prior to vote" - does it refer to internal UMA team discussion or a discussion by voters? Also, would the number of points given for each 🐋 integration be decided separately or the number of bonus points agreed would be applied to all 🐋 integrations equally?

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,129 @@
...
+
+By establishing the terms of a qualifying integration, it allows for the community to reach out in a way that aligns with the current priorities of the UMA protocol- attracting attention to our strongest products and encouraging teams to test out a wide variety of use cases.
+
+
+**It is anticipated that this Price Identifier will be used to create KPI Options, however it is acknowledged that it may be used for a variety of purposes.**
+
+# Data Specifications
+
+The UMA team will maintain a "green light" list of qualified integrations, which will be readily available upon request. Any one can propose adding an integration to the list. Each integration will be added one at a time, after it has been discussed in the #voting channel in the UMA Discord. The following criteria will be considered during the evaluation of each integration:
```
Reinis-FRP, line 36: Will the "green light" list include only the qualified contracts launched between 1622505600 and 1633046399 or the voters should filter out the launch time criteria themselves? If so, this should be explicitly stated in the Implementation section

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,129 @@
...
+* Any use of an LSP contract with a TVL of $100,000 or more (TBD if that's the right number)
+* Any externally launched contract with $100,000 TVL that integrates with Optimistic Oracle.
+
+Contracts included are not limited to Ethereum mainnet- contracts on other chains where DVM message bridging is available (e.g. Polygon) will be included in consideration.
+
+Note: LSP contracts deployed by UMA for use by UMA do not count, but if UMA deploys a contract for use by another team, it is elligible as long as it meets the criteria above. 
+
+The applied timeframe for this UMIP will be between 1622505600 (Jun 01 2021 00:00:00 UTC)  and 1633046399 (Sep 30 2021 23:59:59 UTC), meaning that all qualifying integrations launched during this timeframe should be counted. 
+
+## Ancillary data specifications
```
Reinis-FRP, line 47: As I understand this is only necessary if we want to generalize this UMIP and repeat UMA integrations KPI options in the future. In such case ancillary data could be used to set the start time bound and request timestamp to set end time mound for qualifying contract launch period. If so, then hardcoded timestamps should be removed from the rest of this UMIP and instructions added how the voters should determine the time range from ancillary data and request timestamp.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,139 @@
...
+**Additional Notes:** 
+
+LSP contracts deployed by UMA for use by UMA do not count, but if UMA deploys a contract for use by another team, it is eligible as long as it meets the criteria above.
+
+Contracts included are not limited to Ethereum mainnet- contracts on other chains where DVM message bridging is available (e.g. Polygon) will be included in consideration.
+
+The applied timeframe for this UMIP will be specified in ancillary data, and all qualifying integrations launched during this timeframe should be counted. 
+
+# Ancillary Data Specifications
+The following pieces of data will be included as ancillary data when the contract is deployed, and will provide additional details & instructions to help voters correctly identify the number of qualifying integrations.
```
smb2796, line 70: Few comments on this section:

1. Expiry timestamp shouldn't be passed in through ancillary data. The end bound on the time range considered should just be the price request timestamp (aka the expiry timestamp).
2. Could I suggest adding a starting value as well? Maybe call this "minValue" and have it be denominated in "integrations". Instructions would simply be: if the determined number of integrations is less than the minValue, then return minValue. Could also do this through a FPL, but I think it's cleaner/more flexible in ancillary data. 
3. Can you rename these to single word names? Following JS camelCase conventions, these could all just be: expiryTimestamp etc.
4. Can we provide an example of how it would be passed in with filled out key:value pairs? Should be formatted like [this](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-107.md#ancillary-data-specifications).
5. Would be good to give an expected format for the startTimestamp. startTimestamp would probably just be a unix timestamp, and baseIntegrations should not be scaled

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,139 @@
...
+* Choosing a starting timestamp prior to launch of this option is a design feature that helps to reward the community for the integrations they have already started on and help keep momentum going while the current KPI expires. Using ancillary data to provide this field allows the deployer of the contract to choose the starting timestamp as they see fit.
+
+* The rationale for choosing DAO integrations was to emphasize building DAO-to-DAO relationships. UMA’s product suite has increasingly gained focus and traction on DAO treasury management. However, it is not scalable for the development team to focus on promotion and integration efforts alone, and this KPI Option has been designed to help recruit the community to help.
+
+
+# Implementation
+
+Step 1: Refer to "green light" list for finalized information on all qualified integrations. Assign 1 point for each integration on the list. 
+
+Step 2: If an integration qualifies for a bonus, it will be indicated with a 🐋. These integrations are worth some multiplier of points, which will be specified in the ancillary data. Keep track of bonus points in addition to base points.
```
smb2796, line 106:
```suggestion
Step 2: If an integration qualifies for a bonus, it will be indicated with a 🐋. These integrations are worth some multiplier of points, which will be specified in the ancillary data as `bonusIntegrationsMultiplier`. Keep track of bonus points in addition to base points.  
```

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,139 @@
...
+* Setting it up to allow for all qualified launches to be counted throughout the period rather than just at the time of expiry helps to ensure that we are not missing out on any integrations that might expire before the KPI option does. 
+
+* Choosing a starting timestamp prior to launch of this option is a design feature that helps to reward the community for the integrations they have already started on and help keep momentum going while the current KPI expires. Using ancillary data to provide this field allows the deployer of the contract to choose the starting timestamp as they see fit.
+
+* The rationale for choosing DAO integrations was to emphasize building DAO-to-DAO relationships. UMA’s product suite has increasingly gained focus and traction on DAO treasury management. However, it is not scalable for the development team to focus on promotion and integration efforts alone, and this KPI Option has been designed to help recruit the community to help.
+
+
+# Implementation
+
+Step 1: Refer to "green light" list for finalized information on all qualified integrations. Assign 1 point for each integration on the list.
```
smb2796, line 104: Probably something about verifying that these integrations were within the time-range and meet the criteria within the UMIP.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: 1. Is this something indicated in UMIP or in contract or both?
2. Before doing that I'd like to solicit thoughts from Chris- he seemed to think there was a way around this that wouldn't require voters to account for the minimum value. But, yes let's talk more on this point.
3. yes
4. yes
5. Noted. Will specify

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: Will that be necessary if the green light list is evaluated on a continuous basis? Not opposed to that language, just asking.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
smb2796: > Before doing that I'd like to solicit thoughts from Chris- he seemed to think there was a way around this that wouldn't require voters to account for the minimum value. But, yes let's talk more on this point

1. The contract - not the UMIP
2. Yes, you could do this in a FPL, but you could also do it in ancillary data. Doesn't really matter - I just like the clearer voter visibility of doing it in ancillary data.

britt-uma: I will run that by Clayton to see if he has any input for us on that design feature

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
smb2796: Why would we add a startTimestamp parameter at all then? I think either works, but adding startTimestamp and actually using it would make this more reusable for rolling periods that are referring to different times. Otherwise everything would just refer to the same list

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
smb2796: Hm, maybe there is a misunderstanding here. Adding this gives us flexibility - we aren't defining any range now, so there aren't really any design considerations.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: Ah, I see what you mean. I sort of figured that the same list would be utilized for future reference if we like the way this option turns out, and that it would include all relevant details voters will need- including timestamp and other criteria used to validate its presence on the list. Perhaps I should state that above when I introduce the list?

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: I came to that realization as I was editing. I tried to make that clear in my rationale section- do you mind checking that out?

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Ancillary Data Specifications
+The following pieces of data will be included as ancillary data when the contract is deployed, and will provide additional details & instructions to help voters correctly identify the number of qualifying integrations.
+
+* `startTimestamp` (unix timestamp)
+* `maxBaseIntegrations` (non-negative interger)
+* `maxBonusIntegrations` (non-negative interger)
+* `bonusMinValue` (TVL represented in $USD)
+* `bonusIntegrationsMultiplier` (non-negative number rounded to 2 decimals)
+* `floorIntegrations` (non-negative interger)
```
smb2796, line 85:
```suggestion
* `floorIntegrations` (non-negative integer)
```

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,155 @@
...
+The "green light" list should not be considered the *only* source of truth, but rather a reflection of the discussions that take place within the community. The DVM has the ability and responsibility to challenge the validity of any integrations that have been included or excluded from the list if the broader consensus of UMA tokenholders is that it is within the spirit of this UMIP to do so. Any and all questions surrounding the validity of an integration can be brought up in further detail in the #voting channel on the UMA Discord.
+
+# Ancillary Data Specifications
+The following pieces of data will be included as ancillary data when the contract is deployed, and will provide additional details & instructions to help voters correctly identify the number of qualifying integrations.
+
+* `startTimestamp` (unix timestamp)
+* `maxBaseIntegrations` (non-negative interger)
+* `maxBonusIntegrations` (non-negative interger)
+* `bonusMinValue` (TVL represented in $USD)
+* `bonusIntegrationsMultiplier` (non-negative number rounded to 2 decimals)
```
smb2796, line 84: This shouldn't need any rounding right? Deployers should just specify the exact multiplier they want

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,155 @@
...
+
+The "green light" list should not be considered the *only* source of truth, but rather a reflection of the discussions that take place within the community. The DVM has the ability and responsibility to challenge the validity of any integrations that have been included or excluded from the list if the broader consensus of UMA tokenholders is that it is within the spirit of this UMIP to do so. Any and all questions surrounding the validity of an integration can be brought up in further detail in the #voting channel on the UMA Discord.
+
+# Ancillary Data Specifications
+The following pieces of data will be included as ancillary data when the contract is deployed, and will provide additional details & instructions to help voters correctly identify the number of qualifying integrations.
+
+* `startTimestamp` (unix timestamp)
+* `maxBaseIntegrations` (non-negative interger)
+* `maxBonusIntegrations` (non-negative interger)
+* `bonusMinValue` (TVL represented in $USD)
```
smb2796, line 83: Can you add instructions in implementation for the usage of `bonusMinValue`? Right now it isn't clear how it should be used

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
smb2796: It looks good! The only thing I would add to `Ancillary Data Specifications` would be default values to use if the specific ancillary data keys are not present, or ancillary data is not configured correctly.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: I guess I was just trying to keep someone from using something silly like 3.14159. But if we assume people won't do that, I can take that part out.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: Hadn't thought of that. Would you recommend defaulting to 0 so the contract is just null? This seems like perhaps a larger philosophical issue than a specific concern for this one UMIP. In general, how do we handle contracts that are not specified correctly?

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: Attempted to address this in implementation. Let me know if it's still unclear. A noteworthy component of this is that any implementation that did qualify for a bonus will have been discussed in Discord. If a voter shows up and hasn't participated in discussion, it may be hard for them to try and verify this on their own. Not impossible, but they would need to do some digging and converting to verify.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
britt-uma: Just pushed an update that attempted to recognize this- but I will need guidance on what to do if no maxBaseIntegrations are specified. I assume if this is the case, they will be baked into the contract. But that seems to me it would be an improper use of this UMIP.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,166 @@
...
+
+Ancillary data should be passed into the contract using the format in the example below.
+```
+startTimestamp:1622527200, maxBaseIntegrations:15, maxBonusIntegrations:3, bonusMinValue:$1,000,000, bonusIntegrationsMultiplier:3.00, floorIntegrations:3
+```
+
+If a contract deployer incorrectly formats or simply does not provide one of the parameters above, it will default to the following values.
+
+* `startTimestamp` (timestamp of contract deployment)
+* `maxBaseIntegrations` (???)
```
smb2796, line 95: I think just setting this to 0 works as well. This means that if nothing is specified in ancillary data, then the price returned should be 0.

Britt's drafts/Add uINT_KPI_UMA as a supported price identifier v2.md
```diff
@@ -0,0 +1,166 @@
...
+
+
+# Implementation
+
+Step 1: Refer to "green light" list for finalized information on all qualified integrations. Assign 1 point for each integration on the list, including any integrations that have qualified for a bonus. 
+
+Step 2: If an integration qualifies for a bonus, it will be indicated with a 🐋. These integrations are worth some multiplier of points *in addition* to the base points counted above. This multiplier will be specified in the ancillary data as `bonusIntegrationsMultiplier`. Keep track of bonus points and base points separately.  
+
+* Note for step 2: In order for an integration to earn a 🐋 marker, it must meet the minimum TVL threshold as defined in the ancilary data parameter `minBonusValue`. This value is denominated in $USD rather than the collateral token, so discussion of $USD value had to take place in the #voting channel in UMA Discord in order for an integration to earn this status. For additional details on bonuses, see the above section on bonus integrations.
```
smb2796, line 133: +1!

smb2796 (approved):
I think this looks good! I can just change the default value for baseIntegrations to 0 on my side, but other than that I think this is in good shape. I'm going to approve, merge in and move to last call.

PR #332, opened 2021-06-17
Update final fees - UPP

mhairimca:


therealhexi (approved):
This constant manual update is not going to be scalable. I suggest a proposal that runs through all the approved collateral at the end of each month, and adjust them accordingly. Once the UMIP is approved, we can just run a script that uses $400/TOKEN_USD to get the final fee.

bitznbrewz:
How often do the final fees need to be updated?  Does the store contract need to always maintain ~$400 worth of the collateral currencies in it?

mhairimca:
> This constant manual update is not going to be scalable. I suggest a proposal that runs through all the approved collateral at the end of each month, and adjust them accordingly. Once the UMIP is approved, we can just run a script that uses $400/TOKEN_USD to get the final fee.


Its not ideal, but it doesnt have to be exactly $400, only enough that the DVM is secure, but not so high that it is a barrier to valid disputes or liquidations.   We are going to adjust when they fall outwith the range of approx $300 - $700.  The reason that we are having to adjust so many all at once is because of the significant overall shift in the market in the last month that have thrown a whole load of store fees out of kilter.  

A variable amount of collateral for the store fee causes issues with the contract, which needs a set value of collateral.

final fees - UPP.md
```diff
@@ -0,0 +1,34 @@
...
+
+It is extremely expensive, in terms of gas, for the DVM voters to vote on a price request. The DVM also pays out a relatively large number of rewards on each vote. To ensure that price requests remain rare, they need to be more expensive to discourage any sort of purposeful triggering.
+
+While this does make the capital requirements for liquidators and disputers higher, the increased size of the final fee lockup is unlikely to deter honest participants.
+
+It is also necessary to ensure that the cost of dispute is not so high that is it off-putting to raise valid disputes, risking invalid liquidations being left unchallenged.
+
+## Specifics
+
+ - The PERL final fee will be increased to 4000 PERL.
```
smb2796, line 17: I think most of these values need to be updated due to recent price moves.

britt-uma:
Is it possible for us to have UMIPs refer to a final fee that is stored elsewhere, in a location that is continually updated? Perhaps as a future add on to UMAverse?

smb2796:
> Is it possible for us to have UMIPs refer to a final fee that is stored elsewhere, in a location that is continually updated? Perhaps as a future add on to UMAverse?

There needs to be a governance vote to adjust final fees, and once that vote has occurred the final fee is set to that static number of tokens until another proposal to adjust that comes along. Is this answering the question?

britt-uma:
> > Is it possible for us to have UMIPs refer to a final fee that is stored elsewhere, in a location that is continually updated? Perhaps as a future add on to UMAverse?
> 
> There needs to be a governance vote to adjust final fees, and once that vote has occurred the final fee is set to that static number of tokens until another proposal to adjust that comes along. Is this answering the question?

Not quite. I think what I'm getting at is- I know it currently IS this way, but MUST it be this way? Is this an area where we could consider making a change in the interest of less upkeep? Or would that inherently break a cog in the machine? Rather than having a governance vote to adjust final fees every month, could we hold a governance vote to change the way final fees are updated?

mrice32 (approved):

smb2796 (approved):
Looks good! Going to merge in and move to Last Call

PR #336, opened 2021-06-22
Add PolygonTVL and PolygonTVLinv as price identifiers

Timafon:
https://discourse.umaproject.org/t/add-polygontvl-and-polygontvlinv-as-a-supported-price-identifiers/1175

### DIFF
--- a/umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
+++ b/umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
@@ -105,6 +105,9 @@ PolygonTVLinv
    
 Round to nearest 3 decimal places (fourth decimal place digit >= 5 rounds up and < 5 rounds down)
 
+For receiving historical TVL data make a GET request: https://api.llama.fi/protocol/Polygon/:
+Returns historical data on the TVL of a protocol along with some basic data on it.
+
 
 # Security Considerations
 
--- a/umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
+++ b/umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
@@ -18,7 +18,7 @@ PolygonTVL = Polygon(Matic) TVL / 10^9
 PolygonTVLinv = 1 / Polygon(Matic) TVL * 10^9
 
 # Motivation
-TVL of Defi projects still pretty volatile so if you can predict where TVL will go you need some tools for investing based on TVL.
+TVL of Defi projects are still pretty volatile so if you can predict where TVL will go you need some tools for investing based on TVL.
 
 Polygon(Matic) TVL is needed to create a synthetic token that will reflect the Polygon(Matic) TVL. These tokens can be minted when investors think that Polygon TVL will decrease.
 
@@ -28,7 +28,7 @@ Inverted Polygon(Matic) TVL is needed to create a synthetic token that will refl
 
 To determine TVL, you should use the project Defi Llama (https://defillama.com/home) and calculate by formulas:
 
-PolygonTVL = Polygon(Matic) TVL / 10^9 
+PolygonTVL = Polygon(Matic) TVL / 10^9
 
 PolygonTVLinv = 1 / Polygon(Matic) TVL * 10^9.
 
--- a/umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
+++ b/umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
@@ -1,0 +1,108 @@
+## Headers
+
+| UMIP                |                                                               |
+| ------------------- | ------------------------------------------------------------- |
+| UMIP Title          | Add PolygonTVL and PolygonTVLinv as a supported price identifier |
+| Authors             | Artem Zvyagin (zvyaginartems@gmail.com), Kirill Mokhniuk (kirillmokhniuk@gmail.com), Vitaliy Gataulin (vitalygataulin@gmail.com)                                                       |
+| Status              | Draft                                                         |
+| Created             | June 07, 2021                                                 |
+| Discourse Link      | **Create a post in [UMA's Discourse](https://discourse.umaproject.org/c/umips/18) and link here**            |
+
+# Summary
+
+The DVM should support price requests for Polygon(Matic) total value locked (TVL).
+PolygonTVL and PolygonTVLinv reflects usual and inverted Polygon(Matic) TVL:
+
+PolygonTVL = Polygon(Matic) TVL / 10^9
+
+PolygonTVLinv = 1 / Polygon(Matic) TVL * 10^9
+
+# Motivation
+TVL of Defi projects still pretty volatile so if you can predict where TVL will go you need some tools for investing based on TVL.
+
+Polygon(Matic) TVL is needed to create a synthetic token that will reflect the Polygon(Matic) TVL. These tokens can be minted when investors think that Polygon TVL will decrease.
+
+Inverted Polygon(Matic) TVL is needed to create a synthetic token that will reflect the inverted Polygon(Matic) TVL. These tokens can be minted when investors think that Polygon TVL will increase.
+
+# Data Specifications
+
+To determine TVL, you should use the project Defi Llama (https://defillama.com/home) and calculate by formulas:
+
+PolygonTVL = Polygon(Matic) TVL / 10^9 
+
+PolygonTVLinv = 1 / Polygon(Matic) TVL * 10^9.
+
+**Defi Llama About:** (https://defillama.com/about)
+
+DefiLlama is committed to accurate data without ads or sponsored content, and transparency.
+We list DeFi projects from all chains.
+Thanks to CoinGecko, SNX Tools and PickleJar.info
+Based on Uniswap.info and its fork SushiSwap.vision.
+
+**How Defi Llama calculate TVL:** (https://docs.llama.fi/list-your-project/what-to-include-as-tvl)
+
+Generally, any asset that is held in one of the protocol's contracts can be considered as part of TVL, with two notable exceptions:
+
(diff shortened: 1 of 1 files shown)


umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
```diff
@@ -0,0 +1,108 @@
...
+- Quote Currency: None. This is a ratio
+- Rounding: Round to nearest 3 decimal places (fourth decimal place digit >= 5 rounds up and < 5 rounds down)
+
+# Rationale
+
+We had the option to use CoinGecko or DefiPulse as data providers. Unfortunately, CoinGecko API can't provide Polygon TVL data, so we can't use it currently.
+DefiPulse is not decentralized enough and it is not transparent how it calculates TVL.
+That why we decided to use DefiLlama as a source of TVL data.
+
+# Implementation
```
Reinis-FRP, line 88: Please add the rounding instructions to the Implementation section

umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
```diff
@@ -0,0 +1,108 @@
...
+PolygonTVLinv = 1 / Polygon(Matic) TVL * 10^9
+
+# Motivation
+TVL of Defi projects are still pretty volatile so if you can predict where TVL will go you need some tools for investing based on TVL.
+
+Polygon(Matic) TVL is needed to create a synthetic token that will reflect the Polygon(Matic) TVL. These tokens can be minted when investors think that Polygon TVL will decrease.
+
+Inverted Polygon(Matic) TVL is needed to create a synthetic token that will reflect the inverted Polygon(Matic) TVL. These tokens can be minted when investors think that Polygon TVL will increase.
+
+# Data Specifications
```
Reinis-FRP, line 27: The DefiLlama API endpoint provides only real time data. How the voters should get the historical TVL values for the specific price request timestamp?

therealhexi: +1 Historical data is a requirement

umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
```diff
@@ -0,0 +1,108 @@
...
+
+To determine TVL, you should use the project Defi Llama (https://defillama.com/home) and calculate by formulas:
```
therealhexi, line 29: Does it make sense to just use https://defillama.com/protocol/polygon instead of the defi llama home link?

umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
```diff
@@ -0,0 +1,108 @@
...
+PolygonTVLinv
+
+1. Make a GET request https://api.llama.fi/tvl/Polygon and take the answer (just a number, there is no nesting).
+2. Calculating by formula: PolygonTVLinv = 1 / Polygon(Matic) TVL * 10^9 (where Polygon(Matic) TVL is the response
+   from request above).
+
+
+# Security Considerations
+
+The only reason for concern is the shutdown of the servers hosting the site. All the source code of the project (backend, interface, adapter, sdk) can be found in the repository https://github.com/DefiLlama and deploy the servers. Also (unlikely) the TVL calculation may change critically (https://github.com/DefiLlama/DefiLlama-Adapters/blob/main/projects/polygon/index.js).
```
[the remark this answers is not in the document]
therealhexi: +1, what happens if the API response structure to change in the future? How would voters know the number is incorrect?

umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
smb2796: This seems like the main blocker to me. Does DefiLlama simply not have a historical API, or is it just not listed in the UMIP? I don't see anything in the docs, but @Timafon do you know?

mhairimca: Seems to be available see https://docs.llama.fi/api

"Get Protocol
Returns historical data on the TVL of a protocol along with some basic data on it. The fields `tokensInUsd` and `tokens` are only available for some protocols."

umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
```diff
@@ -0,0 +1,116 @@
...
+We had the option to use CoinGecko or DefiPulse as data providers. Unfortunately, CoinGecko API can't provide Polygon TVL data, so we can't use it currently.
+DefiPulse is not decentralized enough and it is not transparent how it calculates TVL.
+That why we decided to use DefiLlama as a source of TVL data.
+
+# Implementation
+
+-----------------------------------------
+PolygonTVL
+
+1. Make a GET request https://api.llama.fi/tvl/Polygon and take answer (just a number, there is no nesting).
```
smb2796, line 93: Hm this returns a large array of values. Can you query by timestamp?

Also it looks like this endpoint only returns values every 24 hours. If this is intended to be used with a liquidatable synthetic, this is a problem since this data is not updated frequently enough to be used for getting accurate prices. Do you know if there is more granular data available?

smb2796:
It's awesome that there is a historical API, but I think this needs to be pushed 1 week to figure out a couple of things.
1. We should find out what DeFiLlama's calculation methodology is, and then explain that in this UMIP.
2. The historical API listed needs some explanation around how to use it. Also there is an open question on if daily TVL results are sufficient or if there is more granular historical pricing data available.

umip-add_priceidentifiers_PolygonTVL_PolygonTVLinv.md
```diff
@@ -0,0 +1,116 @@
...
+Polygon(Matic) TVL is needed to create a synthetic token that will reflect the Polygon(Matic) TVL. These tokens can be minted and sold when investors think that Polygon TVL will decrease.
+
+Inverted Polygon(Matic) TVL is needed to create a synthetic token that will reflect the inverted Polygon(Matic) TVL. These tokens can be minted and sold when investors think that Polygon TVL will increase.
+
+# Data Specifications
+
+To determine TVL, you should use the project Defi Llama (https://defillama.com/protocol/polygon) and calculate it in USD by formulas:
+
+PolygonTVL = Polygon(Matic) TVL / 10^9
+
```
mhairimca, line 32: Why you are dividing by 10^9 should be explained in the rationale section

smb2796:
> It's awesome that there is a historical API, but I think this needs to be pushed 1 week to figure out a couple of things.
> 
> 1. We should find out what DeFiLlama's calculation methodology is, and then explain that in this UMIP.
> 2. The historical API listed needs some explanation around how to use it. Also there is an open question on if daily TVL results are sufficient or if there is more granular historical pricing data available.

It does seem like you had listed some methodology docs beforehand, but @Timafon any comments on the second point here? Other outcomes from yesterday's community call were:
1. Should discuss in security consideration the lack of data granularity and the fact that this is probably only useful with non-liquidatable contracts.
2. We typically put a clause around voters determining if the results differ from broad market consensus in our umips. This seems especially important here, given that the source you are recommending is an aggregator.
3. Really there should be a price feed built for defillama. If you intend to use this with the EMP/other liquidatable contracts, this should really be viewed as a blocker. If used with things like the LSP, this is still really important for OO price proposers/disputers, but less of a blocker.

TsarKir:
> > It's awesome that there is a historical API, but I think this needs to be pushed 1 week to figure out a couple of things.
> > 
> > 1. We should find out what DeFiLlama's calculation methodology is, and then explain that in this UMIP.
> > 2. The historical API listed needs some explanation around how to use it. Also there is an open question on if daily TVL results are sufficient or if there is more granular historical pricing data available.
> 
> It does seem like you had listed some methodology docs beforehand, but @Timafon any comments on the second point here? Other outcomes from yesterday's community call were:
> 
> 1. Should discuss in security consideration the lack of data granularity and the fact that this is probably only useful with non-liquidatable contracts.
> 2. We typically put a clause around voters determining if the results differ from broad market consensus in our umips. This seems especially important here, given that the source you are recommending is an aggregator.
> 3. Really there should be a price feed built for defillama. If you intend to use this with the EMP/other liquidatable contracts, this should really be viewed as a blocker. If used with things like the LSP, this is still really important for OO price proposers/disputers, but less of a blocker.

What do you mean by "price feed" in 3? @smb2796

smb2796:
> > > It's awesome that there is a historical API, but I think this needs to be pushed 1 week to figure out a couple of things.
> > > 
> > > 1. We should find out what DeFiLlama's calculation methodology is, and then explain that in this UMIP.
> > > 2. The historical API listed needs some explanation around how to use it. Also there is an open question on if daily TVL results are sufficient or if there is more granular historical pricing data available.
> > 
> > 
> > It does seem like you had listed some methodology docs beforehand, but @Timafon any comments on the second point here? Other outcomes from yesterday's community call were:
> > 
> > 1. Should discuss in security consideration the lack of data granularity and the fact that this is probably only useful with non-liquidatable contracts.
> > 2. We typically put a clause around voters determining if the results differ from broad market consensus in our umips. This seems especially important here, given that the source you are recommending is an aggregator.
> > 3. Really there should be a price feed built for defillama. If you intend to use this with the EMP/other liquidatable contracts, this should really be viewed as a blocker. If used with things like the LSP, this is still really important for OO price proposers/disputers, but less of a blocker.
> 
> What do you mean by "price feed" in 3? @smb2796

UMA price proposers/verifiers and liquidator/disputers typically need a programmatic way to calculate a price. This traditionally takes the form of a "price feed" that agrees with the typical UMA convention around calculating prices. You can see these price feeds [here](https://github.com/UMAprotocol/protocol/tree/master/packages/financial-templates-lib/src/price-feed).

smb2796:
Closing until there are updates on this! Feel free to open once ready to address these [comments](https://github.com/UMAprotocol/UMIPs/pull/336#issuecomment-875639608).

PR #339, opened 2021-06-22
Add mph and ndx price identifiers

bitznbrewz:
- Draft moved from previous pull request
- Merging should be blocked until Bancor
price feed is implemented


smb2796:
Closing until we have a bancor price feed that can support

PR #340, opened 2021-06-23
Add GYSR as a collateral type

mhairimca:


smb2796:
Closing since this was moved into #343

PR #342, opened 2021-06-23
Upp-Remove SGT

henrystats:


therealhexi (approved):

therealhexi:
What's a generally unacceptable range for collateral? Volume of < $10,000, or Fully Diluted Value of < $2mm? I saw GYSR are also around this category in the #343 do we want to add that at all?

akhanubis (commented):
Looks good, not sure if it's necessary to mention the SGT rug pull as the reason for the liquidity drop.

UPPs/upp-4.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Summary
+
+This UPP will remove SGT & DSD from the collateral currency whitelist introduced in [UMIP-8](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-8.md). 
+
+## Rationale
+
+## Delisting DSD
+
+DSD was approved as a collateral currency in [UMIP-36](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-36.md] and has since been far below the 1$ peg, and with a daily volume of less than $10k it isn’t suitable as a collateral currency.
```
smb2796, line 13: This motivation is good color, but I think the main motivation would need to be more along the lines of - It is not clear that these assets will maintain some sort of expected range of value.

PR #341, opened 2021-06-23
Add uTVL_KPI_SDT as price identifier for KPI Options

Hahz48:

PR #344, opened 2021-06-26
Add u tvl kpi sdt as supported price identifier

Tumbiola:
Made changes requested , need to find historical data

### DIFF
--- a/add uTVL_KPI_SDT as supported price identifier.md
+++ b/add uTVL_KPI_SDT as supported price identifier.md
@@ -1,0 +1,84 @@
+Headers
+| UMIP #
+
+| ------------------- | ------------------------------------------------------------- |
+| UMIP Title | Add uTVL_KPI_SDT as supported price identifier|
+| Authors | Greg Tumbiola - Tumbiola@gmail.com |
+| Status | Draft |
+| Created | June 12th, 2021 |
+| Discourse Link | Add uTVL_KPI_SDT as price identifier for KPI Options |
+
+Summary
+This UMIP enables the DVM to support price requests based on the TVL of Stake Dao.
+The metric being used for Stake DAO’s KPI Options is to reach the key target of 10^9 ($1 billion) in TVL.
+
+Motivation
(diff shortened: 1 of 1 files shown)


smb2796:
Seems like two prs were opened for the same UMIP. Most of the discussion on the uTVL_KPI_SDT price identifier happened over in #341, but is seems more productive to continue the conversation here since @Tumbiola you are listed as the author.

Outcomes from community call/feedback that need to be addressed and then this can be voted on! 
1. Historical TVL data like you already mentioned.
2. We should define the underlying methodology for how the SDT TVL API results are calculated behind the scenes. This will give UMA voters a way to fallback to a calculation method in case the API is not available or the results just need to be verified.

add uTVL_KPI_SDT as supported price identifier.md
```diff
@@ -0,0 +1,84 @@
...
+This price identifier offers a guarantee that these options will be of value, even if this key metric is poor. However, the amount of value that can be locked in the protocol is potentially limitless and consequently a ceiling price of 10 SDT is required to provide a cap limit.
+The methods used to calculate the dollar value of each of the collateral currencies have been chosen to adhere to previous design decisions in such calculations through UMIPs that have already been approved through our governance procedure.
+There is no need for price processing. This is a snapshot based on a particular time, however it may be useful for TVL Options holders to have oversight of the ongoing TVL and consequently query the value of their options on an ongoing basis.
+Implementation
+As there is no requirement for ongoing monitoring through liquidation or dispute bots, a price feed is not required. The only requirement is a query of the SDT TVL statistic at the timestamp 00.00(UTC) on the date of expiry according to the data and markets as defined above under - DATA SPECIFICATIONS.
```
Reinis-FRP, line 63: Are there any reasons why the price request timestamp cannot be used? Otherwise no need to hardcode it here

Reinis-FRP (commented):
In addition to @smb2796 comments please also change the formatting to Markdown

PR #349, opened 2021-07-08
feat: Register Updated Optimistic Oracle

nicholaspai:


UMIPs/umip-PLACEHOLDER.md
```diff
@@ -0,0 +1,53 @@
...
+- [Refactor ancillary data stamping logic to new AncillaryData library](https://github.com/UMAprotocol/protocol/commit/a3bf46270787cbaae4ed2218f064b1217c153a50#diff-ddccb329354a2589ec298040dd90f48a74ddaa5129516804145af7e6b8ca6327)
+- [Fix typo in comment for `stampAncillaryData` method](https://github.com/UMAprotocol/protocol/commit/d146d4feddcf99e157a909032a3071e95ca51eb9#diff-ddccb329354a2589ec298040dd90f48a74ddaa5129516804145af7e6b8ca6327)
+- [Response to audit of new `OptimisticOracle`](https://github.com/UMAprotocol/protocol/commit/dd211c4e3825fe007d1161025a34e9901b26031a#diff-ddccb329354a2589ec298040dd90f48a74ddaa5129516804145af7e6b8ca6327)
+
+The new `OptimisticOracle` uses a new `AncillaryData` library that manages converting different data types to `utf8`-decodable strings which can be found [here](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/common/implementation/AncillaryData.sol).
+
+Contract addresses can be found here:
+- [OptimisticOracle](https://etherscan.io/address/0x9f631aea633be790b45f2d8e91081efcded7d256#code)
```
mrice32, line 48: nit: this contract looks like it was deployed outside of our typical hardhat deployment system. Specifically, the optimizer is off, which means it will have higher gas costs. We should probably redeploy with these settings.

mrice32 (commented):
Looks great! One minor comment about the deployment.

chrismaree (approved):

UMIPs/umip-PLACEHOLDER.md
```diff
@@ -0,0 +1,53 @@
...
+The current supported optimistic oracle "stamps" data to the ancillary data informing voters that any price requests were sent by an optimistic oracle. This is useful information and gives voters more information, but the "stamping" does not conform with the above interface; it produces `bytes` data that cannot be directly decoded to `utf8` format. 
+
+This UMIP proposes to register an updated optimistic oracle that "stamps" ancillary data according to the above standard.
+
+## Technical Specification
+To accomplish this upgrade, the following actions need to be taken:
+- A new `OptimisticOracle` contract has been deployed. Address can be found in "Implementation" section.
+- A transaction will need to be proposed to add this new address to the `Finder` contract under the name `“OptimisticOracle”`. This is how other contracts will find the optimistic oracle and reference it.
+- The `OptimisticOracle` will need to be registered with the `Registry` so that it can make requests to the DVM.
+- (TBD) We need to un-register the current `OptimisticOracle`.
```
Reinis-FRP, line 26: Do we need to check any price request status for contracts using current OO or un-registration can take place at any time?

Reinis-FRP (commented):
Looks good, just one question on how un-registration of current OO will be handled

UMIPs/umip-PLACEHOLDER.md
nicholaspai: Redeployed and amended address

UMIPs/umip-PLACEHOLDER.md
nicholaspai: We will need to make sure that there are no pending requests for the current OO before we upgrade, otherwise those requests will be left "hanging" and someone will have to manually resubmit them to the new OptimisticOracle. As a sanity check, do you agree with this understanding @mrice32?

nicholaspai, line 26: @chrismaree @mrice32 do you think we need to unregister the old OO?

mrice32: @nicholaspai @Reinis-FRP great points!

In my opinion, it would be ideal to de-register, but doing so _at the same time_ as registering the new one brings along some risks.

- Even if there are no in-flight requests at the start of the vote, there could be by the end of the vote when the txn would be processed.
- Requests that get sent just before the swap would effectively be un-disputable, meaning there would be no way to stop a first proposer from corrupting the result. A second governance request could fix this, but that would take days and be too late.

I think this opens up a security issue. However, I think it's good practice to de-register active contracts that are no longer used to avoid security risks down the line.

Given all this, I think the best course of action is to officially deprecate the old OO with this request and de-register it in a separate UMIP after the swap. Any contracts correctly using the finder should not be affected. Only contracts that ignore the deprecation or incorrectly interface with the OO could be corrupted, which should be nobody.

nicholaspai: I support that plan

> think the best course of action is to officially deprecate the old OO with this request and de-register it in a separate UMIP after the swap. Any contracts correctly using the finder should not be affected.

Reinis-FRP (approved):

PR #352, opened 2021-07-13
Update DSD final fee

mhairimca:


Update DSD Final fee.md
```diff
@@ -0,0 +1,31 @@
+##  Summary
+
+  
+
+This UPP will increase the final fees for DSD to be around $400. This will increase the cost of initiating disputes for the DVM.
+
```
therealhexi, line 6: Any reason there are excessive new lines between each paragraph?

therealhexi (approved):
Looks good other than formatting

Update DSD Final fee.md
smb2796: I don't think there is. In interest of time, we can just fix once merged.

smb2796 (approved):

PR #359, opened 2021-07-14
Create General KPI options price identifier

ChandlerDeKock:


General KPI.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Motivation
+Price (or data) Identifiers for KPI options tend to be project-specific, time sensitive, and not reusable identifiers for other projects to make use of. Instead, this identifier sets a general standard for all metric-based feeds and removes the need to create individual proposals for KPI identifiers. 
+
+The user of the General KPI Price Identifier is intended to have their project report their specific metric themselves in the interest of incentivizing their users.
+
+## Data Specification
+Since the project running KPI options is expected to provide some visibility on the target metric, it is reasonable to assume that the user of the `General KPI Price Identifier` provides an endpoint for serving this data. The role of UMA DVM would be to verify and deliver human judgment failover considering the linked source code of the end server and general target objectives and calculation methodology.
+
+To accomplish this, the structure of the ancillary data will require these inputs:
```
britt-uma, line 24: Is there a character limit that users should be conscious of when inputting ancillary data?

General KPI.md
```diff
@@ -0,0 +1,54 @@
...
+
+- **Metric:**  short description reflecting the metric and units to be measured.
+- **Endpoint:** -link to data endpoint that should return JSON key-value pair response including the metric at expiration.
+- **Method:**  link to a descriptive source covering the objective and methodology for calculating a particular metric. This methodology should include how the calculation of the metric is measured. Links to repositories and explainer documentation are strongly recommended. Cleary articulate any post-processing required
```
britt-uma, line 28: Is there a way to add a requirement that this information must be stored on a platform that records version history? That would help to ensure that if the methodology is tampered with after deploying the contract, it can at least be detected by the DVM

General KPI.md
```diff
@@ -0,0 +1,54 @@
...
+- **Fallback:** (optional) - In the event of the end-point not working or reporting false outcomes, a fallback ensures that UMA token holders can arrive at the proper result. If the methodology source demonstrates a fallback alternative, this field can be optional. 
+
```
britt-uma, line 30: It would be good to outline default values for when fields are not specified and the resulting consequences (for example having the token expire worthless)

General KPI.md
```diff
@@ -0,0 +1,54 @@
...
+The data should also have the following accessibility requirements:
+
+- Be open and freely queriable by UMA token holders
+- Have the ability to reference the measure at any point in time historically, but specifically be able to report the contract expiry outcome.
+- Have clear documentation of the reported figures
+- Clear instructions on what processing is happening on the reported data. 
+
```
britt-uma, line 37: If possible, it would be good to include an example of and ancillary data section that meets the standard we expect users to meet.

General KPI.md
Reinis-FRP: 8192 bytes, Added to Ancillary Data Specifications

General KPI.md
Reinis-FRP: Added to the data accessibility requirements in Data Specifications

General KPI.md
Reinis-FRP: Added in the Implementation section as fallback value (defaults to 0)

General KPI.md
Reinis-FRP: Added Aragon v2 migration example to Ancillary Data Specifications

General KPI.md
```diff
@@ -0,0 +1,103 @@
...
+ 
+The user of this UMIP accepts responsibility to provide the information passed through via ancillary data is sufficient to resolve price requests. Suppose the user of the price identifier cannot give enough information for UMA token holders to determine the outcome. In that case, UMA token holders can expire the contract in a worthless state or a result that will deem the options to have a value of zero (or as close to zero as the terms allow). 
+
+This UMIP also allows the use of KPI measures that may, in some part, rely on information that a trusted party may report. The assumption is that the party reporting the numbers is either trusted or has a vested interest in ensuring the accuracy of the price, but this is not guaranteed. As such, the voter follows i) the methodology outlined in the ancillary data, ii) the fall back alternative outlined in the ancillary data, iii) the most commonly agreed metric to measure a KPI (If and only if this measure is trusted). If none of these outcomes are achieved or the data source is corrupted then the options can be expired to zero (or as close to worthless).
```
ChandlerDeKock, line 80: I am not sure all contracts can expire worthless if using an FPL that has a floor value. 
I see the potential of having the options needing to expire worthlessly but due to how the terms are structured, it may not be possible. 

This can be solved by i) accepting that a floor value is accepted or ii) being able to expire a non-zero floor value option at zero

General KPI.md
ChandlerDeKock: Kinda think we should have the UMA uTVL and uDAO options instead. Not a blocker or a need to change this but U do we having an example where one team has two different KPI's is useful.

Reinis-FRP: Added this, and this made me realize we need to also add arbitrary ancillary data fields if they are defined in the `Method` document.

General KPI.md
```diff
@@ -0,0 +1,129 @@
...
+
+As an example, possible ancillary data for UMA TVL KPI options (as specified in [UMIP-65](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-65.md) is listed below if it was using `General KPI Price Identifier` instead:
+
+```
+Metric:TVL in UMA financial contracts measured in billions of USD,
+Endpoint:api.umaproject.org/uma-tvl,
+Method:github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-65.md,
+Key:currentTvl,
+Interval:Updated every 10 minutes,
+Rounding:-7,
```
smb2796, line 77: why have rounding and scaling as specific ancillary data params? Couldn't this just be in the `method`?

General KPI.md
```diff
@@ -0,0 +1,129 @@
...
+- `Fallback` (optional): In the event of the end-point not working or reporting false outcomes, a fallback ensures that UMA token holders can arrive at the proper result. If the methodology source demonstrates a fallback alternative, this field can be optional.
+- `Key`: Which key value from the `Endpoint` response should be used by voters for further processing of the price request.
+- `Interval`: This describes how request timestamps for pricing queries should be rounded and what is the granularity of historical data update frequency.
+- `Aggregation` (optional): In case any time series data processing is required this describes processing method used (e.g. calculating TWAP, finding peak value, etc.) and also sets the start timestamp for such aggregation.
+- `Rounding`: This is integer number defining how many digits should be left to the right of decimal delimiter after rounding.
+- `Scaling` (optional): This is integer number defining power of 10 scaling to be applied after rounding.
```
smb2796, line 63: All prices are returned as 18 decimals. Am I misunderstanding this field?

General KPI.md
```diff
@@ -0,0 +1,129 @@
...
+
+```
+Metric:Number of qualifying UMA DAO integrations,
+Endpoint:api.umaproject.org/uma-dao-integrations,
+Method:github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-112.md,
+Key:currentIntegrations,
+Interval:Updated daily,
+Rounding:2,
+startTimestamp:1622527200,
+maxBaseIntegrations:15,
```
smb2796, line 90: Would all of these be present in ancillary data? I'd assume these would just be in the method doc, right?

General KPI.md
```diff
@@ -0,0 +1,129 @@
...
+
+Voters should ensure that their results do not differ from broad market consensus. In particular, voters should verify whether the `Endpoint` source code referenced in the linked `Method` document is consistent with the objective and methodology for calculating particular metric, as well as attempt to run the data service independently and check the accuracy of data provided by the `Endpoint` service. In the event of the `Endpoint` not working or reporting false outcomes voters should follow any fallback instructions found in the linked `Method` document or the optional `Fallback` parameter in the ancillary data.
+
+Voters should also carefully follow any versioning history to the linked `Method` document and the `Endpoint` service source code. While it is possible that methodology and code could be improved over time in order to clarify any original oversight or fix bugs, voters should ensure that any changes are consistent with the original KPI options program launch objectives.
```
smb2796, line 119: Wonder if we should have a documented recommendation or even requirement to how these should be hosted and referenced? Seems like maybe the API code could change, but the method should stay constant, no? Maybe it needs to be on IPFS or just a specific git commit hash needs to be referenced. Thoughts?

General KPI.md
```diff
@@ -0,0 +1,129 @@
...
+- Be open and freely queryable by UMA token holders
+- Have the ability to reference the measure at any point in time historically, but specifically be able to report the contract expiry outcome
+- Have clear documentation of the reported figures
+- Clear instructions on what processing is happening on the reported data
+- Methodology document and source code for the data endpoint service should be hosted on platform providing full versioning history with the ability to detect whether any modifications have been made
+
+# Technical Specifications
+
+-----------------------------------------
+- Price identifier name:  General KPI Price Identifier
```
chrismaree, line 46:
```suggestion
- Price identifier name: General KPI Price Identifier
```

chrismaree: in fact, this double space is done a few times in some of the lists that follow. please fix for consistency

General KPI.md
chrismaree: prehaps it is an additional scalling. i.e you take back the number from the API then scale it up by another 10x for voters. DVM returned 1e18 scaled but this could be pre-scalling? 

in any case some clarification would be great!

General KPI.md
```diff
@@ -0,0 +1,129 @@
...
+- `Aggregation` (optional): In case any time series data processing is required this describes processing method used (e.g. calculating TWAP, finding peak value, etc.) and also sets the start timestamp for such aggregation.
+- `Rounding`: This is integer number defining how many digits should be left to the right of decimal delimiter after rounding.
+- `Scaling` (optional): This is integer number defining power of 10 scaling to be applied after rounding.
+
+KPI options launcher can also add any arbitrary ancillary data fields as long as they are clearly documented in the linked `Method` document and voters are instructed on how to interpret them for price processing.
+
+When designing the ancillary data KPI options deployer should be aware that the total size of ancillary data cannot exceed 8192 bytes also accounting for any ancillary data stamping by Optimistic Oracle. This limit would be checked by the LSP creator contract upon the deployment.
+
+As an example, possible ancillary data for UMA TVL KPI options (as specified in [UMIP-65](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-65.md) is listed below if it was using `General KPI Price Identifier` instead:
```
chrismaree, line 69: it might be useful to also include an encoded version of this.

General KPI.md
```diff
@@ -0,0 +1,129 @@
...
+ 
+The user of this UMIP accepts responsibility to provide the information passed through via ancillary data is sufficient to resolve price requests. Suppose the user of the price identifier cannot give enough information for UMA token holders to determine the outcome. In that case, UMA token holders can expire the KPI option tokens in a worthless state. 
+
+This UMIP also allows the use of KPI measures that may, in some part, rely on information that a trusted party may report. The assumption is that the party reporting the numbers is either trusted or has a vested interest in ensuring the accuracy of the price, but this is not guaranteed. As such, the voter follows i) the methodology outlined in the ancillary data, ii) the fall back alternative outlined in the ancillary data, iii) the most commonly agreed metric to measure a KPI (If and only if this measure is trusted). If none of these outcomes are achieved or the data source is corrupted then the options can be expired to  worthless state.
```
chrismaree, line 105: just to clarify one thing: by "worthless state" do you mean that the DVM will return a price of 0? this is not necessarily "worthless" (i.e the short tokens in a linear contract would get all the collateral). It might be useful to:

a) clarify what you mean by worthless
b) potentially add an additional field to the ancillary data structure that defines the "zero" state. i.e if the computation fails due to any reason, fallback does not work and the commonly agreed method also does not work then return this number rather than just return "0".

General KPI.md
Reinis-FRP: It can be set anywhere, but putting it in ancillary data makes it easier for voters as this becomes more standardized procedure while the `Method` document focuses only on unique aspects on how to fetch raw metric data

General KPI.md
Reinis-FRP: Yes, this is not the same as voting dApp scaling. E.g. for uTVL it would have been transforming raw TVL in USD to billions of USD.
Added the clarification on this

General KPI.md
Reinis-FRP: `Method` doc would define them as additional required parameters and requestor would need to pass them to ancillary data as part of the request.

General KPI.md
Reinis-FRP: Should be OK as long as voters can track any modifications and decide whether they go against original objective.

General KPI.md
Reinis-FRP: Fixed

General KPI.md
Reinis-FRP: I have newlines between key-value pairs here just for readability, but actual encoded version would not have this.

Before committing this, what is the optimal approach to escape special characters? For example ":" in full url should be escaped and also "," needs special treatment. Can I define `\` as escape character?

General KPI.md
Reinis-FRP: b) added `Unresolved` parameter to ancillary data

Reinis-FRP: a) Clarified implications for zero price response to long token holders in the Rationale section. Reduced ambiguity in other places just to focus on actual price request response (not considering payout)

General KPI.md
Reinis-FRP: Changed to `\\` for escape character

Reinis-FRP: Added encoded versions

smb2796 (approved):

PR #371, opened 2021-07-27
Create general TOKEN_PRICE identifier

Reinis-FRP:


nicholaspai (commented):
Really thoughtful write up, well done @Reinis-FRP ! I had an idea to make the ancillary data example cover another edge case. Overall I like how the UMIP details various fallback mechanisms.

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,138 @@
...
+
+# Summary
+
+The DVM should support price requests for TOKEN_PRICE identifier that returns a price of any fungible token expressed in units of another fungible token, cryptocurrency or fiat currency. Ancillary data will guide voters in identifying which token should be priced, what should be its quote currency, which markets should be queried and any other processing instructions for resolving a particular price request.
+
+This UMIP outlines the critical considerations for any user of this identifier. The user acknowledges that extra steps are required to use the identifier for their specific TOKEN_PRICE to resolve at expiry.
+
+# Motivation
+
+This price identifier will allow a deployer to use any UMA-approved collateral token to be locked in Long Short Pair (LSP) contracts for issuing range bonds, success tokens, call options or any other financial instrument parameterized in its specific financial product library without the need for a use-specific UMIP. This will considerably reduce the time and effort involved in creating a contract where no such identifier exists.
```
smb2796, line 19: Why would the motivation be specific to the long short pair? Works for EMPs or OO requests as well, right?

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,138 @@
...
+
+# Price Feed Implementation
+
+This price identifier can use any price feed from the [UMA protocol repository](https://github.com/UMAprotocol/protocol/tree/master/packages/financial-templates-lib/src/price-feed) as long as it allows pricing the collateral token quoted in the required other token, crypto currency or fiat currency. The actual price feed configuration should be passed as JSON formatted object in the `configuration` parameter from the ancillary data.
+
+If the required application needs a new price feed implementation, the user of this price identifier can develop its own price feed and open the Pull Request. In case of new price feed implementation the user should wait before it is reviewed and merged into the main (master) branch of the [UMA protocol repository](https://github.com/UMAprotocol/protocol) in order for other proposers/disputers and UMA voters being able to run the price feed script with the passed configuration from the ancillary data.
+
+# Technical Specifications
```
smb2796, line 39: Meta comment, but generally I feel like tech specs doesn't really serve any purpose in umips today. Maybe we should consider just combining this section back with data specs?

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,138 @@
...
+- Price identifier name: TOKEN_PRICE
+- Base Currency: Determined by the `base` parameter from the ancillary data
+- Quote Currency: Determined by the `quote` parameter from the ancillary data
+- Rounding: Determined by the `rounding` parameter from the ancillary data
+
+## Ancillary Data Specifications
+
+When converting ancillary data to UTF8 string it must contain price request parameters expressed as a list of key-value pairs delimited by `,` (commas) and each key-value pair further delimited by `:` (colons). If a value should contain `,` or `:` characters, such value should be enclosed in double quotes. The below listed key parameters will be used to instruct voters how to resolve a given price request for this identifier and request timestamp:
+- `base`: Collateral token symbol that should be priced in the submitted price request.
+- `baseAddress`: `base` token deployment address on Ethereum network (or any other network if `baseChain` parameter is provided). In case the token is deployed on several networks the user of this price identifier is free to select any network as long as it allows correctly identifying the required base token.
```
smb2796, line 51: Is base address necessary if the address is implied through the markets that are chosen?

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,138 @@
...
+- `unresolved` (optional): This is numeric value that voters should return for unresolvable price request (defaults to zero if omitted).
+- `configuration`: Price feed configuration formatted as a JSON object that can be used to construct a price feed by [CreatePriceFeed.js](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js). Please consult UMA documentation covering the selection and configuration of price feed for more details. Also note that default UMA price feed configuration is formatted as JavaScript objects, thus all key names should be enclosed in doublequotes, no comments are allowed, math expressions should be kept in one line and there should be no trailing commas after the last element in JSON. Even though technically proposer/dispute bots would be able to pick up any correctly formatted configuration for available price feed implementation, this price feed configuration should only be used for pricing the `base` token in terms of the `quote` currency.
```
smb2796, line 59: > Even though technically proposer/dispute bots would be able to pick up any correctly formatted configuration for available price feed implementation, this price feed configuration should only be used for pricing the `base` token in terms of the `quote` currency.

Why is this restriction in place? Wouldn't this be appropriate for any supported pf config?

UMIPs/General Token Price.md
Reinis-FRP: Sure, added

UMIPs/General Token Price.md
Reinis-FRP: This restriction is kept in order to being able to generally describe fallback mechanism for voters to resolve if the `configuration` is not working

UMIPs/General Token Price.md
smb2796: fair - guess I was thinking of situations where there was a dex address listed

UMIPs/General Token Price.md
Reinis-FRP: For this generic price identifier there is no unique information in this section, so deleted.
For other UMIPs indeed pair details can be merged to data specification and rounding should be set in Implementation

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,132 @@
...
+
+If the required application needs a new price feed implementation, the user of this price identifier can develop its own price feed and open the Pull Request. In case of new price feed implementation the user should wait before it is reviewed and merged into the main (master) branch of the [UMA protocol repository](https://github.com/UMAprotocol/protocol) in order for other proposers/disputers and UMA voters being able to run the price feed script with the passed configuration from the ancillary data.
+
+## Ancillary Data Specifications
+
+When converting ancillary data to UTF8 string it must contain price request parameters expressed as a list of key-value pairs delimited by `,` (commas) and each key-value pair further delimited by `:` (colons). If a value should contain `,` or `:` characters, such value should be enclosed in double quotes. The below listed key parameters will be used to instruct voters how to resolve a given price request for this identifier and request timestamp:
```
therealhexi, line 43: Give an example of `"` as double quotes enclosed

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,132 @@
...
+- `base`: Collateral token symbol that should be priced in the submitted price request.
+- `baseAddress`: `base` token deployment address on Ethereum network (or any other network if `baseChain` parameter is provided). In case the token is deployed on several networks the user of this price identifier is free to select any network as long as it allows correctly identifying the required base token.
+- `baseChain` (optional): ChainID number for the network where base token is deployed. If omitted, Ethereum mainnet would be assumed.
```
therealhexi, line 46: Any reason we use `Chain` over `Network` in the naming here?

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,132 @@
...
+- `quote`: Quote currency symbol in which the `base` token should be priced in
+- `quoteDetails`: Additional details to unambiguously identify the quote currency. For cryptocurrencies or fiat currencies this should be the full name of the currency. For fungible tokens it should be the token address on Ethereum network or any other network if accompanied by its ChainID in this parameter.
```
therealhexi, line 48: Would be simplify if we refactor the `quoteAddress` out to a separate optional field? `quoteDetails` can be used for only quote types such as fiat currencies, or those who don't have an address

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,132 @@
...
+- `rounding`: This is an integer number defining how many digits should remain to the right of the decimal delimiter after rounding the obtained price.
```
therealhexi, line 49: Articulate the rounding rule, >= .5 rounds up, < .5 rounds down

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,132 @@
...
+- `configuration`: Price feed configuration formatted as a JSON object that can be used to construct a price feed by [CreatePriceFeed.js](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js). Please consult UMA documentation covering the selection and configuration of price feed for more details. Also note that default UMA price feed configuration is formatted as JavaScript objects, thus all key names should be enclosed in doublequotes, no comments are allowed, math expressions should be kept in one line and there should be no trailing commas after the last element in JSON. Even though technically proposer/dispute bots would be able to pick up any correctly formatted configuration for available price feed implementation, this price feed configuration should only be used for pricing the `base` token in terms of the `quote` currency.
+
+When designing the ancillary data TOKEN_PRICE user should be aware that the total size of ancillary data cannot exceed 8192 bytes also accounting for any ancillary data stamping by Optimistic Oracle. This limit would be checked by the LSP creator contract upon the deployment.
```
therealhexi, line 55: Where does the technical limitation of 8192 bytes come from? 8000 characters are not that large

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,132 @@
...
+base:UMA,baseAddress:0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828,quote:USD,quoteDetails:United States Dollar,rounding:6,fallback:"https://www.coingecko.com/en/coins/uma",configuration:{
+    "type": "medianizer",
+    "minTimeBetweenUpdates": 60,
+    "twapLength": 3600,
+    "medianizedFeeds": [
+      { "type": "cryptowatch", "exchange": "coinbase-pro", "pair": "umausd" },
+      { "type": "cryptowatch", "exchange": "binance", "pair": "umausdt" },
+      { "type": "cryptowatch", "exchange": "okex", "pair": "umausdt" }
+    ]
+  }
```
therealhexi, line 69: Do we need to stringify this json object? I'm assuming the entire ancillary data is modeled as a string? Also, any reason we didn't use json but rather used csv format for ancillary data?

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+When this ancillary data dictionary is stored as bytes, the result would be:
```
therealhexi, line 72: Any chance you could link a UTF encoding tool here for reader to cross verify?

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,132 @@
...
+4. Voters should run the price feed with the selected key name as `--identifier` and requested timestamp as `--time` parameter from the root of UMA [protocol](https://github.com/UMAprotocol/protocol) repository. Following on the example above, the command would be:
+   ```
+   yarn truffle exec ./packages/core/scripts/local/getHistoricalPrice.js --network mainnet_mnemonic --identifier UMAUSD_CUSTOM --time 1626984000
+   ```
+5. Obtained price feed script value should be rounded leaving `rounding` number of digits after the decimal delimiter. In case the `rounding` parameter is not provided or it is not possible to interpret its value, voters should apply default rounding of 6 digits after the decimal delimiter.
+6. It might be possible that some data sources do not return price data for the requested timestamp due to lack of trading activity or any other technical reason causing the price feed script to fail. In case the price feed `configuration` contains alternative pricing routes for the same asset pair (e.g. when calculating median), voters should reconfigure it without the failing data source and attempt to calculate the requested price from the remaining data sources.
```
therealhexi, line 109: Suggest moving the `edge case scenarios` into a separate bullet list to reduce complexity

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,132 @@
...
+
+When designing the ancillary data TOKEN_PRICE user should be aware that the total size of ancillary data cannot exceed 8192 bytes also accounting for any ancillary data stamping by Optimistic Oracle. This limit would be checked by the LSP creator contract upon the deployment.
+
+As an example, possible ancillary data for UMA token priced in USD (as specified in UMAUSD in [UMIP-57](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-57.md)) if the user was using TOKEN_PRICE price identifier instead:
+
+```
+base:UMA,baseAddress:0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828,quote:USD,quoteDetails:United States Dollar,rounding:6,fallback:"https://www.coingecko.com/en/coins/uma",configuration:{
```
therealhexi, line 60: This UMIP moves the price config definition from UMIP process to the contract deployer. This speeds up the process, but removed the careful UMA Team scrutiny from the process. What type of testing can TOKEN_PRICE contract deployer do to ensure the price config ancillary data works? This may be out of the scope for this UMIP, but a very important one for anyone to use this contract. You can add a ##Testing section if needed

therealhexi (commented):
Very well articulated UMIP, a big step towards scalability

UMIPs/General Token Price.md
```diff
@@ -0,0 +1,132 @@
...
+| Status              | Draft                                                                                            |
+| Created             | July 23, 2021                                                                                    |
+| Discourse Link      | [UMA's Discourse](https://discourse.umaproject.org/t/create-general-token-price-identifier/1265) |
+
+# Summary
+
+The DVM should support price requests for TOKEN_PRICE identifier that returns a price of any fungible token expressed in units of another fungible token, cryptocurrency or fiat currency. Ancillary data will guide voters in identifying which token should be priced, what should be its quote currency, which markets should be queried and any other processing instructions for resolving a particular price request.
+
+This UMIP outlines the critical considerations for any user of this identifier. The user acknowledges that extra steps are required to use the identifier for their specific TOKEN_PRICE to resolve at expiry.
+
```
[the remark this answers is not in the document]
Reinis-FRP: Added EMP usage in Security considerations

UMIPs/General Token Price.md
Reinis-FRP: It should be included in example below (`fallback:"https://www.coingecko.com/en/coins/uma"`)

UMIPs/General Token Price.md
Reinis-FRP: ChainID should be unique as introduced in [EIP-155](https://eips.ethereum.org/EIPS/eip-155), while ETC and ETH share the same NetworkID

UMIPs/General Token Price.md
Reinis-FRP: I was considering this initially, but this also requires having separate `quoteChain` parameter. Since `base` can be only token it makes sense to have `baseAddress` and `baseChain`, but for `quote` that seemed to introduce unnecessary complexity and increased parameter count.

UMIPs/General Token Price.md
Reinis-FRP: Added to Implementation section

UMIPs/General Token Price.md
Reinis-FRP: This is related to Ethereum gas limit for hashing the price request. See more detailed explanation [here](https://github.com/UMAprotocol/protocol/blob/15a8f31e3d3ce0df9b68b03ae56f8df789ae481a/packages/core/contracts/oracle/implementation/Voting.sol#L137).

UMIPs/General Token Price.md
Reinis-FRP: See more discussion at [AncillaryData.sol](https://github.com/UMAprotocol/protocol/blob/15a8f31e3d3ce0df9b68b03ae56f8df789ae481a/packages/core/contracts/common/implementation/AncillaryData.sol#L5) comments.
The reason why the whole ancillary data is string, not JSON object is so that OO can easily stamp on requester address at the end of request

UMIPs/General Token Price.md
Reinis-FRP: This should be covered in the added EMP usage scenario

UMIPs/General Token Price.md
Reinis-FRP: done

smb2796 (approved):
Looks like all comments were addressed. Great work

PR #374, opened 2021-07-30
Split out Technical Specification section for price identifier template

Reinis-FRP:


smb2796 (approved):
Looks good!

PR #376, opened 2021-08-04
Updates umip-108 to reflect gas fee market changes made by EIP-1559

nateawelch:
This PR updates the gas price umip in order to account for the changes to the ethereum gas fee market made by EIP-1559. Specifically it changes a few references from using the gas price of a transaction to using the effective gas price, which accounts for both the old and new transaction format

### DIFF
--- a/UMIPs/umip-108.md
+++ b/UMIPs/umip-108.md
@@ -56,6 +56,103 @@ For a price request made before `1633046400`, the price will be resolved to a 2-
 
 Updated rounding: 6 decimals
 
+In order to account for the changes to the fee market included in EIP-1559, the following pseudo-algorithm is used to calculate the exact data point to report by a DVM reporter:
+
+```python
+def return_median(t, N)
+    assert N in [1, 4, 24, 168, 720]
+    minimum_ranges = {1: 200, 4: 800, 24: 4800, 168: 33600, 720: 134400} # a mapping between the durations 1HR, 4HR, 24HR (1D), 168HR (1W), 720HR(1M) and the corresponding mimimum number of blocks that must have been mined within the period.
+    
+    # `t` is number of seconds since the epoch (unix time)
+    end_block = block.at(timestamp=t) # rounded down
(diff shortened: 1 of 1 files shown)


smb2796:
Tested this out on my end, and it looks great. Good stuff.

One structural comment. There is a lot of content in this umip that might not relate to the previous calculation methodology., especially within the Rationale section. Generally I think this change could just use some context by not being an overwrite to the original ugas umip, but rather an amendment in a later ugas pid.  

Also I think it is advantageous to keep an untouched record of the previous uGAS methodology. Can I suggest a change to this pr to make an amendment to [UMIP-108](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-108.md) instead. Should just be a note about EIP-1559 necessitating this change, and listing the 2 new queries. Let me know what you think.

nateawelch:
> Tested this out on my end, and it looks great. Good stuff.
> 
> One structural comment. There is a lot of content in this umip that might not relate to the previous calculation methodology., especially within the Rationale section. Generally I think this change could just use some context by not being an overwrite to the original ugas umip, but rather an amendment in a later ugas pid.
> 
> Also I think it is advantageous to keep an untouched record of the previous uGAS methodology. Can I suggest a change to this pr to make an amendment to [UMIP-108](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-108.md) instead. Should just be a note about EIP-1559 necessitating this change, and listing the 2 new queries. Let me know what you think.

Appreciate the feedback! Agreed that leaving UMIP-16 as is would be beneficial, I've updated the PR with your suggestion and moved changes to UMIP-108

smb2796 (commented):
Looks good and thanks for making that update! 

There are just a couple parts where it refers to using `the price will be resolved with the median monthly gas price calculation defined for GASETH-1M-1M in UMIP-20`. Can you just update this language to clarify that the new calculation method in this UMIP is going to be used.

smb2796 (approved):

smb2796:
Thanks for updating! Merging

PR #385, opened 2021-08-14
Fix to straight quotation marks

Reinis-FRP:

In order not to over-complicate the ancillary data parsing module better to use just straight quotation marks for escaping values with delimiters (: or ,)


therealhexi (approved):
Good to have things standardized.

smb2796 (approved):

PR #397, opened 2021-08-31
Add BarnBridge SMART Alpha KPI options implementation

Reinis-FRP:
This is WIP, requires input from BarnBridge team on SMART pool contract ABI to fetch on-chain data

### DIFF
--- a/Implementations/smart-alpha.md
+++ b/Implementations/smart-alpha.md
@@ -34,12 +34,9 @@ TVLPriceFeed:<TVL_PRICE_FEED>
 2. Determine the total value locked in terms of the underlying asset in the `Pool` contract at the request timestamp.
 3. Using the quote currency and price feed from Step 1 calculate the value of underlying asset balance from Step 2 expressed in terms of the quote currency at the request timestamp.
 4. Based on the SMART Alpha pool contract deployment timestamp and its epoch length identify all possible epoch periods that could fall in between the start timestamp (passed within the `Aggregation` parameter from the ancillary data) and the request timestamp. Even if there have been no user deposits/withdrawals and the epoch has not been manually advanced, this calculation should assume maximum available epoch periods. 
-5. For each identified SMART Alpha pool epoch period from Step 4 determine its junior side dominance and calculate epoch target points. In case the epoch has not been advanced junior side dominance for such epoch period should be taken from the last advanced epoch period.
-   * if junior dominance is lower or equal to 20% assign 0.5 target point;
-   * if junior dominance is higher than 20% and lower or equal to 40% assign 1 target point;
-   * if junior dominance is higher than 40% and lower or equal to 60% assign 2 target points;
-   * if junior dominance is higher than 60% and lower or equal to 80% assign 1 target point;
-   * if junior dominance is higher than 80% assign 0.5 target point.
+5. For each identified SMART Alpha pool epoch period from Step 4 determine its junior side dominance and calculate epoch target points. In case the epoch has not been advanced junior side dominance for such epoch period should be taken from the last advanced epoch period. Actual target points should be calculated using a simple step-wise linear function that increases from 0 to 2 as junior dominance rises from 0% to 50%, then decreases back to 0 as junior dominance rises to 100%:
+   * if junior dominance is lower or equal to 50% assign target points calculated by multiplying junior dominance share with coefficient 4;
+   * if junior dominance is higher than 50% assign target points calculated by multiplying senior dominance share (this is 1 minus junior dominance share) with coefficient 4.
 6. Calculate the arithmetic mean target points from Step 5.
 7. Multiply TVL from Step 3 with average target points from Step 6.
 8. Round the adjusted TVL from Step 7 to 2 decimals before returning it as a resolved price request.
--- a/Implementations/smart-alpha.md
+++ b/Implementations/smart-alpha.md
@@ -4,7 +4,7 @@ BarnBridge SMART Alpha KPI Options Calculation
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


Implementations/smart-alpha.md
```diff
@@ -0,0 +1,43 @@
+## Title
+
+BarnBridge SMART Alpha KPI Options Calculation
+
+## Summary
+
+SMART Alpha allows users to calibrate their exposure to the performance of an underlying asset represented by junior- and senior-side proofs of liquidity.  When the price goes up, seniors give up some of their assets to the juniors, and vice versa. It is epoch-based, meaning that users deposit the underlying asset for a set period of time and cannot enter and exit until the subsequent epoch. The goal of the KPI options program is to a) bootstrap senior-side secondary liquidity, and b) compensate for junior-side risk.
```
maxfiege, line 7: The last sentence should be changed to: "The goal of the KPI options program is to bootstrap secondary liquidity for both senior and junior SMART Alpha positions."

Implementations/smart-alpha.md
```diff
@@ -0,0 +1,42 @@
...
+SMART Alpha allows users to calibrate their exposure to the performance of an underlying asset represented by junior- and senior-side proofs of liquidity.  When the price goes up, seniors give up some of their assets to the juniors, and vice versa. It is epoch-based, meaning that users deposit the underlying asset for a set period of time and cannot enter and exit until the subsequent epoch. The goal of the KPI options program is to bootstrap secondary liquidity for both senior and junior SMART Alpha positions.
+
+This document will detail the calculation method for senior/junior balance adjusted TVL in requested SMART Alpha pool.
+
+## Intended Ancillary Data
+
+```
+Metric:SMART Alpha pool TVL adjusted by average senior/junior balance,
+Endpoint:"<BARNBRIDGE_API_ENDPOINT>",
+Method:"https://github.com/UMAprotocol/UMIPs/blob/master/Implementations/smart-alpha.md",
```
therealhexi, line 16: I'm assuming this PR will get merged into this file?

Implementations/smart-alpha.md
```diff
@@ -0,0 +1,42 @@
...
+Rounding:2,
+Pool:<ADDRESS_OF_SMART_ALPHA_POOL>,
+TVLCurrency:<TVL_CURRENCY>,
+TVLPriceFeed:<TVL_PRICE_FEED>
+```
+***Note 1:** <BARNBRIDGE_API_ENDPOINT> might not yet be available at the launch of test pilot, hence, `Endpoint` and `Key` parameters would be skipped. Voters would need to rely only on independent calculation based on this implementation document.*
+
+***Note 2:** This is generic ancillary data template, thus <START_TIMESTAMP> and <ADDRESS_OF_SMART_ALPHA_POOL> should be specified upon deploying specific KPI options contract for an incentivized SMART Alpha pool.*
+
+***Note 3:** `TVLCurrency` and `TVLPriceFeed` are optional parameters that should provide the currency and price feed in case TVL should be measured in different currency than pool price feed quote currency.*
```
therealhexi, line 29: Nit: Is there any value add for articulating the default base and quote currency when such fields are missing? Or does it entirely depend on the Pool?

Implementations/smart-alpha.md
```diff
@@ -0,0 +1,42 @@
...
+3. Using the quote currency and price feed from Step 1 calculate the value of underlying asset balance from Step 2 expressed in terms of the quote currency at the request timestamp.
+4. Based on the SMART Alpha pool contract deployment timestamp and its epoch length identify all possible epoch periods that could fall in between the start timestamp (passed within the `Aggregation` parameter from the ancillary data) and the request timestamp. Even if there have been no user deposits/withdrawals and the epoch has not been manually advanced, this calculation should assume maximum available epoch periods. 
+5. For each identified SMART Alpha pool epoch period from Step 4 determine its junior side dominance and calculate epoch target points. In case the epoch has not been advanced junior side dominance for such epoch period should be taken from the last advanced epoch period. Actual target points should be calculated using a simple step-wise linear function that increases from 0 to 2 as junior dominance rises from 0% to 50%, then decreases back to 0 as junior dominance rises to 100%:
```
therealhexi, line 37: How does Epoch Target Points get calculated? Where is it defined?

therealhexi (commented):
I'm slightly concerned with the complexity of the implementation. Is there any value in adding an example?

Implementations/smart-alpha.md
Reinis-FRP: Yes

Implementations/smart-alpha.md
Reinis-FRP: Base is always the underlying pool token and normally quote should be inferred from the linked price feed (discussed in Implementation Step 1). Thus, `TVLCurrency` and `TVLPriceFeed` are optional and only should be provided when TVL metric should be measured in different currency than pool quote currency.

Implementations/smart-alpha.md
Reinis-FRP: Target points are calculated as a function of junior/senior balance ratio (see below)

Reinis-FRP:
> I'm slightly concerned with the complexity of the implementation. Is there any value in adding an example?

This is still WIP and bit abstract, but I hope to solve it after adding more specific methods on how tho fetch required data on-chain.

therealhexi (approved):
More specific methods are good to have but not required. Looks good to me.

Implementations/smart-alpha.md
Reinis-FRP: Had to change this to make `TVLPriceFeed` required (except when TVL is measured in the underlying itself) since the pool contracts do not expose the Chainlink aggregator contract that is required for applying correct decimal scaling.

maxfiege (approved):

smb2796 (approved):

PR #406, opened 2021-09-15
Add uPUNK-1221 Price Identifier

rossgalloway:
New Price identifier that is an update to UMIP 84 to make the UMIP compatible with an EMP contract that does not allow ancillary data. This UMIP should also include a new price feed that removes any 0 price Cryptopunk trades.


UMIPs/UMIP add PUNKETH-1221.md
```diff
@@ -0,0 +1,224 @@
...
+
+# Take median of values
+median([cryptopunk_blockprice[cryptopunk]["value"] for cryptopunk in cryptopunks])
+
+```
+***TO-DO: Update Implementation***
+
+An Implementation can be found here in the [uPUNK price feed](https://github.com/UMAprotocol/protocol/pull/2917)
+
+The TWAP price queried prior to expiry can be determined using the existing [UniswapPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js). The only required input would be determining which pool has the highest volume.
```
Reinis-FRP, line 128:
```suggestion
The TWAP price queried prior to expiry can be determined using the existing [UniswapPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.ts). The only required input would be determining which pool has the highest volume.
```
Price feed implementations have been migrated to TypeScript

therealhexi (commented):
Reminder on the TODOs

UMIPs/UMIP add PUNKETH-1221.md
```diff
@@ -0,0 +1,234 @@
...
+
+# Create cryptopunks array to store ids
+cryptopunks = []
+
+# Create mapping to store most recent price
+cryptopunk_blockprice = {}
+
+# Find the last PunkBought event for each CryptoPunk
+for event in events_corrected:
+    punk = event.cryptopunk
```
Reinis-FRP, line 112: `events_corrected` does not have an item with property `cryptopunk` - do you mean `event.punkIndex`?
Please also check any `event.cryptopunk` referenced below.

UMIPs/UMIP add PUNKETH-1221.md
```diff
@@ -0,0 +1,234 @@
...
+
+    # Add punk to mapping if not previously added AND if price was
+    # traded at is higher than 0
+    if punk not in cryptopunks and event.value > 0:
```
Reinis-FRP, line 116: `events_corrected` does not have an item with property `value` - do you mean `event.price`?

UMIPs/UMIP add PUNKETH-1221.md
```diff
@@ -0,0 +1,234 @@
...
+        push(event.cryptopunk, cryptopunks)
```
Reinis-FRP, line 117: What is the ordering of the `push` function arguments? Above when item was added to `events_corrected` the item added was the second argument.

UMIPs/UMIP add PUNKETH-1221.md
```diff
@@ -0,0 +1,234 @@
...
+        cryptopunk_blockprice[event.cryptopunk] = {"block": event.block, "value": event.value}
```
Reinis-FRP, line 118: `events_corrected` does not have an item with property `block` - do you mean `event.blockNumber`?
Please also check any `event.block` referenced below.

UMIPs/UMIP add PUNKETH-1221.md
```diff
@@ -0,0 +1,234 @@
...
+        more_recent = event.block > current_value["block"]
+        price_gt_0 = event.value > 0
+        if more_recent and price_gt_0:
+            cryptopunk_blockprice[event.cryptopunk] = {"block": event.block, "value": event.value}
+
+# Take median of values
+median([cryptopunk_blockprice[cryptopunk]["value"] for cryptopunk in cryptopunks])
+
+```
+***TO-DO: Update Implementation***
```
Reinis-FRP, line 137: Do you plan to update this?

UMIPs/UMIP add PUNKETH-1221.md
cc7768: Yep.

UMIPs/UMIP add PUNKETH-1221.md
cc7768: I updated this to have the collection first and then the new item being pushed.

UMIPs/UMIP add PUNKETH-1221.md
cc7768: Originally this was meant to point at the JS price feed. Should we have it point at the Python one instead?

Reinis-FRP: JS is better, but till its completed link the Python code instead.

UMIPs/UMIP add PUNKETH-1221.md
```diff
@@ -0,0 +1,233 @@
...
+    if punk not in cryptopunks and event.price > 0:
+        push(cryptopunks, event.punkIndex)
+        cryptopunk_blockprice[event.punkIndex] = {"block": event.blockNumber, "price": event.price}
+
+    else:
+        # Find current values of block/value associated with a particular
+        # cryptopunk -- Defaults to {block: 0, value: 0} if they don't exist
+        current_value = {
+            "block": cryptopunk_blockprice[punk]["block"]
+            "value": cryptopunk_blockprice[punk]["value"]
```
Reinis-FRP, line 125:
```suggestion
            "value": cryptopunk_blockprice[punk]["price"]
```

Reinis-FRP, line 125: value or price? not sure if should be replaced in both places, but cryptopunk_blockprice[punk] has `price` property

Reinis-FRP (approved):

PR #411, opened 2021-09-22
Sherlock exploit protection implementation

smb2796:


Implementations/sherlock.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Summary
+
+This query is intended to allow users of [Sherlock](https://sherlock.xyz/) to request the DVM for arbitration in the case of a disagreement about an smart contract exploit insurance claim.
+
+This implementation will be used with the `General_KPI` price identifier, and should return one of the values passed within ancillary data depending on the results of arbitration. Additionally, within ancillary data will be a link to this document (for voter guidance on resolution steps) as well as any additional supporting resources that the price requester wishes to provide.
+
+## Motivation
+
+Sherlock is a new type of exploit protection protocol launching this September. The idea is to have smart contract security experts assess/price protocol coverage, then get paid based on the performance of the protocols (no hacks = large payout, large hack = no payout). Protocols who want coverage will pay Sherlock monthly premiums, and in return, Sherlock will use it’s staking pool to repay hacks at protocols. Sherlock is doing a guarded launch with $30M in staking pool funds and writing coverage for up to $10M per protocol. More details can be found here.
```
abg4, line 13: The link for more details is missing.

Implementations/sherlock.md
```diff
@@ -0,0 +1,75 @@
...
+
+Other exploit protection protocols rely on their tokenholders to decide whether claims should be paid out or not. Sherlock wants to do better and would like to use UMA’s Data Verification Mechanism as the final escalation to decide on claims payouts. This would help coverage buyers feel more comfortable that they have access to an unbiased party’s decision on a claim.
+
+For context, Sherlock agrees on coverage terms with every protocol. An example can be seen [here](https://docs.google.com/document/d/19OEnspNhO-U3CosXt9NRSyfWlnJFq2Z1VnqGA8yJAkM/edit#heading=h.8uxda5sbuvuu). These terms will be broadly the same across all protocols but there may be extra language added for protocol-specific exploit risks.
```
abg4, line 17: The link to see an example requires a request for access. Is this intentional?

Implementations/sherlock.md
```diff
@@ -0,0 +1,75 @@
+## Title
+
+Sherlock Exploit Protection
+
+## Summary
+
+This query is intended to allow users of [Sherlock](https://sherlock.xyz/) to request the DVM for arbitration in the case of a disagreement about an smart contract exploit insurance claim.
+
+This implementation will be used with the `General_KPI` price identifier, and should return one of the values passed within ancillary data depending on the results of arbitration. Additionally, within ancillary data will be a link to this document (for voter guidance on resolution steps) as well as any additional supporting resources that the price requester wishes to provide.
```
Reinis-FRP, line 9: The usage of `General_KPI` might seem confusing as it has completely different use case to resolve some metric and also the ancillary data fields are not compliant with [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md).

Instead, `YES_OR_NO_QUERY` from [UMIP-107](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-107.md) seems more appropriate as one of its intended use cases is insurance.

Still, better option would be to approve either Sherlock specific price identifier or more general price identifier that allows handling any insurance claims. One reason is that `YES_OR_NO_QUERY` does not have required ancillary data fields (though one can point them from the question parameter), but more importantly the price identifier lacks the required disclaimers in the security section that protocols using this price identifier should not expect UMA voters to be experts in the field and they do not accept any liability from potential claims from either requesting protocol or Sherlock.

Implementations/sherlock.md
```diff
@@ -0,0 +1,75 @@
...
+2. Because the protocol making the claim will likely be the party constructing the ancillary data and submitting a price request, voters should heavily rely on checking external resources for additional context. The Sherlock Coverage Terms Agreement should always be referenced. Additionally the Protocol's original claim as well as the Sherlock Claims Committee's original claims decision will likely always be published and should be referenced. Sherlock's Coverage Agreement details the exact types of exploits that Sherlock covers, and these should be the only exploits that are considered in arbitration. More information is provided in the `Additional Resources` section below.
+3. Voters should return a value of `1` if they determine that:
+   1. There was an exploit for the `Protocol` specified.
+   2. The exploit type is covered by Sherlock's Coverage Agreement.
+   3. The exploit happened as part of the `StartBlock` specified in ancillary data.
+   4. The total USD value lost in the exploit matches the USD `Value` within ancillary data.
```
Reinis-FRP, line 61: How should the USD value lost be measured - are prices taken at the moment of the hack? What if the hack was continuous and the USD price of lost assets was fluctuating?

Maybe also allow for situation when value lost was higher than claimed, just to account for pricing measurement errors.

Reinis-FRP (commented):
Very interesting use case!

Since this is new application for UMA protocol, how do we estimate Total Value Secured in the context to ensuring CoC>PfC? Is it all pools staked at Sherlock as they are at risk being claimed by any project through UMA DVM?

Also what guarantees and interest alignment UMA protocol can offer to claiming protocols that voters indeed would invest time and resources to research on highly advanced technical matters and judge fairly?

Implementations/sherlock.md
smb2796: Yep, waiting on team to provide and then will add it in.

Implementations/sherlock.md
smb2796: Nope. Same as before, waiting on team to provide a publicly accessible version.

Implementations/sherlock.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Motivation
+
+Sherlock is a new type of exploit protection protocol launching this September. The idea is to have smart contract security experts assess/price protocol coverage, then get paid based on the performance of the protocols (no hacks = large payout, large hack = no payout). Protocols who want coverage will pay Sherlock monthly premiums, and in return, Sherlock will use it’s staking pool to repay hacks at protocols. Sherlock is doing a guarded launch with $30M in staking pool funds and writing coverage for up to $10M per protocol. More details can be found here. 
+
+Other exploit protection protocols rely on their tokenholders to decide whether claims should be paid out or not. Sherlock wants to do better and would like to use UMA’s Data Verification Mechanism as the final escalation to decide on claims payouts. This would help coverage buyers feel more comfortable that they have access to an unbiased party’s decision on a claim.
```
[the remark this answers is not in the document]
smb2796: Currently there is no enforcement other than a social contract btwn Sherlock and claiming protocols.

Implementations/sherlock.md
```diff
@@ -0,0 +1,75 @@
...
+
+Relying on UMA’s DVM for claims decisions will significantly increase the value of on-chain exploit protection mechanisms in the eyes of protocols and end users. One of the biggest hurdles to adoption of coverage is the biased nature of claims processes. Sherlock believes that the involvement of UMA’s DVM will make exploit protection protocols more trustworthy and further normalize adoption of coverage by protocols and end users alike. By adding this product, UMA can blaze a trail that contributes to the safety of DeFi. Sherlock would love to set a precedent with UMA for how exploit claims disputes should be resolved on-chain. 
+
+## Intended Ancillary Data
+
+```
+Metric:Sherlock exploit claim arbitration,
+Method:"https://github.com/UMAprotocol/UMIPs/blob/master/Implementations/sherlock.md",
+Protocol:<PROTOCOL_REQUESTING_CLAIM>,
+Blockchain:<BLOCKCHAIN_OF_POTENTIAL_EXPLOIT>,
```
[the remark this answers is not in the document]
smb2796: good call

Implementations/sherlock.md
```diff
@@ -0,0 +1,75 @@
...
+Value:1,000,000,
+StartBlock:13,207,345,
+CoverageAgreement:Link to protocol specific coverage agreement with Sherlock
+```
+
+## Rationale and usage
+
+The framing of the price identifier puts the responsibility on the protocol to specify the correct blocktime, blockchain, and payout amount in order to receive a payout. If any of the information is not accurate, a payout will not occur. This is mitigated by the idea that a protocol will get up to 3 chances to correctly input the data in order to successfully trigger a payout.
+
+Because the potential request space is broad, it is possible that voters will not be able to arrive at a definitive result. In this case, the default return value should be “no payout” (or 0.5). This places the onus on the protocol to provide clear evidence that an exploit occurred. This seems like a proper approach because a protocol can always resubmit once they’ve accumulated better evidence. However, if a claim is paid out in error, there is little recourse for Sherlock at that point. It seems reasonable that a protocol should have to prove that a payout occurred and that the default should be “no action,” meaning a payout is not enacted.
```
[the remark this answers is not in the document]
smb2796: This was a typo. Good catch

Implementations/sherlock.md
```diff
@@ -0,0 +1,75 @@
...
+Protocol:UMA Protocol,
+Blockchain:Ethereum,
+Value:1,000,000,
+StartBlock:13,207,345,
+CoverageAgreement:Link to protocol specific coverage agreement with Sherlock
+```
+
+## Rationale and usage
+
+The framing of the price identifier puts the responsibility on the protocol to specify the correct blocktime, blockchain, and payout amount in order to receive a payout. If any of the information is not accurate, a payout will not occur. This is mitigated by the idea that a protocol will get up to 3 chances to correctly input the data in order to successfully trigger a payout.
```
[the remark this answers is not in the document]
smb2796: No, protocols could still ask again after a 0 result. The assumption is that it would just be less likely, since Sherlock will have a bonding system on their side and a result of 0 should only be returned when there is clear evidence that the claim is not supported.

Implementations/sherlock.md
smb2796: Good call on building a cushion into the value. It's very hard to give instructions for how the USD value could be measured, as this could vary dramatically across different types of hacks. This is why it seems appropriate to leave it open ended and have that be part of arbitration, but open to any suggestions.

UMIPs/sherlock.md
```diff
@@ -0,0 +1,87 @@
...
+Resources:<OPTIONAL_RESOURCE_LINKS>,
+CoverageAgreement:<COVERAGE_AGREEMENT_LINK>
+```
+
+Example:
+```
+Metric:Sherlock exploit claim arbitration,
+Protocol:UMA Protocol,
+ChainId:1,
+Value:1,000,000,
```
Reinis-FRP, line 44:
```suggestion
Value:1000000,
```
commas are not supported in the ancillary data parser hence would better skip them for numeric values. In other cases where commas are required the value can be enclosed in doublequotes

UMIPs/sherlock.md
```diff
@@ -0,0 +1,87 @@
...
+StartBlock:13,207,345,
```
Reinis-FRP, line 45:
```suggestion
StartBlock:13207345,
```
Same comment as above

UMIPs/sherlock.md
```diff
@@ -0,0 +1,87 @@
...
+
+Relying on UMA’s DVM for claims decisions will significantly increase the value of on-chain exploit protection mechanisms in the eyes of protocols and end users. One of the biggest hurdles to adoption of coverage is the biased nature of claims processes. Sherlock believes that the involvement of UMA’s DVM will make exploit protection protocols more trustworthy and further normalize adoption of coverage by protocols and end users alike. By adding this product, UMA can blaze a trail that contributes to the safety of DeFi. Sherlock would love to set a precedent with UMA for how exploit claims disputes should be resolved on-chain. 
+
+## Ancillary Data Specifications
```
Reinis-FRP, line 27: Maybe add the link to [docs on formatting ancillary data](https://docs.umaproject.org/uma-tokenholders/guidance-on-adding-price-identifiers#ancillary-data-specification)?

Reinis-FRP (approved):

PR #413, opened 2021-09-23
Add carbon price identifiers from BITA

EndymionJkb:


UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,157 @@
...
+# Rationale
+
+These are proprietary indices, and a feed should also become available on the Confluence site. The founders have a great track record: one is co-founder of Etho Capital, which published the
+best-performing ESG index in the US (ticker ETHO on NASDAQ), now with a 5+ year track record.
+
+Nevertheless, a third party licensed entity (on a different continent), whose entire business is publishing indices, is an even more reliable and resilient data source.
+
+Eventually these indices may be listed on a major stock exchange, such as NASDAQ, at which point there will be many additional data feeds.
+
+# Implementation
```
Yuen5405, line 73: Since this is a daily price feed, what time is the price feed being updated?
Assuming the price feed updates at 00:00 UTC every day, say a price request happened at 23:58 UTC on the 24th, should the price request accept the price published at 00:00 24th or wait for two more minutes for the latest one?

UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,157 @@
...
+no way to gain exposure to these types of funds in crypto.
+
+The first price identifiers are: CGLCAR, CKLCAR, and CNLCAR, corresponding to the global and regional low carbon indices from Confluence Analytics.
+Confluence Analytics proposes to create Perpetual Synthetics that track these indices, and then include them in liquidity pools.
+
+# Data Specifications
+
+BITA is an EU-registered [index administrator](https://www.bitadata.com/assets/files/index-services/BITA_Index_Administration_-_Service_One_Pager.pdf).
+They calculate indices in both fiat and crypto, adjusting for corporate actions, stock splits, etc., and produce daily detailed files for all holdings,
+which are also accessible through an API.
```
therealhexi, line 27: Nit: no need for a new line?

UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,157 @@
...
+
+There is a rate-limited public API for current and historical index prices.
```
therealhexi, line 29: What's the rate limit?

UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,157 @@
...
+
+# Price Feed Implementation
+
+The price feed is a REST API hosted at api.bitadata.com. Users first "log in" (with public credentials) to obtain a token, then use standard token authentication for subsequent requests.
+The public credentials provide access to the three indices named above.
+
+# Rationale
+
+These are proprietary indices, and a feed should also become available on the Confluence site. The founders have a great track record: one is co-founder of Etho Capital, which published the
+best-performing ESG index in the US (ticker ETHO on NASDAQ), now with a 5+ year track record.
```
therealhexi, line 67: nit: again new line

UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,157 @@
...
+```
+{
+  "message": "Authenticated succesfully",
+  "token": "eyJ0eXAiOiQiLCJhbGciOiJIUzI1NiJ9.eyJzd..."
+}
+```
+
+This token grants access to the "indexes_eod" and "index_value" endpoints.
+
+For the current price of one or more indices:
```
therealhexi, line 96: Nit: Call out specifically to replace `{{token}}` below with the token value obtained from the response above.

UMIPs/umip_add_priceidentifier_BITA.md
therealhexi: +1

EndymionJkb: I believe it is daily market close (this is tracking a fiat index), not midnight. So prices aren't going to change until the next morning. I am asking them to verify this, though. (It's a third party index calculator company; I don't control it.)

UMIPs/umip_add_priceidentifier_BITA.md
EndymionJkb: 100 requests per minute per user - will update in the text.

UMIPs/umip_add_priceidentifier_BITA.md
EndymionJkb: Updated in the text

UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,155 @@
...
+| ------------------- | ------------------------------------------------------------- |
+| UMIP Title          | Add CGLCAR, CKLCAR, and CNLCAR as supported price identifiers |
+| Authors             | Jeffrey Bennett (endymionjkb@gmail.com)                       |
+| Status              | Draft                                                         |
+| Created             | 9/6/2021                                                      |
+| Discourse Link      | discourse.umaproject.org/t/add-a-basket-of-price-identifiers-calculated-by-bita-bitadata-com/1322
+
+# Summary 
+
+The DVM should support price requests from BITA (bitadata.com) that reflect the prices of a set of published fiat indices.
```
Reinis-FRP, line 13: The summary does not seem to be accurate - BITA is only the data source, but the summary should focus more on actual price identifiers.

UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Motivation
+
+ESG (Environmental, Social, Governance) funds are high performing and growing in popularity in the traditional finance world, but there is currently
+no way to gain exposure to these types of funds in crypto.
+
+The first price identifiers are: CGLCAR, CKLCAR, and CNLCAR, corresponding to the global and regional low carbon indices from Confluence Analytics.
+Confluence Analytics proposes to create Perpetual Synthetics that track these indices, and then include them in liquidity pools.
```
Reinis-FRP, line 21: How would perpetual synthetics contract be implemented? If its based on [UMA protocol repository](https://github.com/UMAprotocol/protocol/tree/master/packages/core/contracts/financial-templates/perpetual-multiparty) then it also requires funding rate price identifier.
Also, please note that UMA perpetual contract is not yet fully tested in production.

UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,155 @@
...
+          {"timestamp":"2021-08-02","value":"990.15"},
+          {"timestamp":"2021-08-03","value":"991.40"},
+          ...
+          {"timestamp":"2021-09-03","value":"1023.52"}
+        ]
+      }
+    }                                                                                             
+```
+
+You can find full API documentation [here](https://docs.bitadata.com/).
```
Reinis-FRP, line 147: Please also add rounding instructions.

UMIPs/umip_add_priceidentifier_BITA.md
EndymionJkb: They just answered. I will update.

UMIPs/umip_add_priceidentifier_BITA.md
EndymionJkb: Ok; updated to list the indices, and just says they're published by BITA.

The same API would also support further indices, if they were added to the set of offerings, which is why I had it that way in the beginning. But you're right, this makes more sense.

UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,155 @@
...
+They calculate indices in both fiat and crypto, adjusting for corporate actions, stock splits, etc., and produce daily detailed files for all holdings,
+which are also accessible through an API.
+
+There is a rate-limited public API for current and historical index prices. The limit is 100 requests per minute per user.
+
+-----------------------------------------
+- Price identifier name: CGLCAR (Confluence Global ESG Low Carbon Index)
+- Base Currency: - USD
+- Quote Currency: - USD
+- Markets & Pairs: USD/CGLCAR
```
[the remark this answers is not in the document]
EndymionJkb: I was just copying from the template. They're indices, but the prices are in USD. I've updated to just show the Base Currency and that's it.

UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,155 @@
...
+- Example data providers: BITA
+- Cost to use: Free, rate-limited
+- Real-time data update frequency: daily
+- Historical data update frequency: daily
+
+# Price Feed Implementation
+
+The price feed is a REST API hosted at api.bitadata.com. Users first "log in" (with public credentials) to obtain a token, then use standard token authentication for subsequent requests.
```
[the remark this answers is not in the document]
EndymionJkb: I was coming to that conclusion as well. Looks like I need to add a new price feed implementation that calls this API, and reference it here.

UMIPs/umip_add_priceidentifier_BITA.md
EndymionJkb: Yes, it is from the UMA protocol perpetual, and I'm aware of the funding rate, and not sure what needs to be done there. In early conversations I'd heard there might be one for general use at some point; I need to talk to the team to understand what to do for that.

Would that all be in the same UMIP?

UMIPs/umip_add_priceidentifier_BITA.md
EndymionJkb: I'm not sure what to do there. The API returns high precision, floating point values. Should I say that since they are currency values, they should be rounded to 2 decimals?

UMIPs/umip_add_priceidentifier_BITA.md
Reinis-FRP: Technically funding rate identifier can also be in the same UMIP.
Also note that you would need to run funding rate proposer bots in addition to liquidation/dispute bots and on mainnet this could be quite costly.

UMIPs/umip_add_priceidentifier_BITA.md
Reinis-FRP: Rounding is normally done in order to ease reaching consensus.
Since historic API returns 2 decimal precision I would suggest that this price identifier should be rounded to 2 decimals (just in case the raw precision changes in the future)

UMIPs/umip_add_priceidentifier_BITA.md
EndymionJkb: Ok; I need to understand all this better. The indices could of course perform differently, so there would need to be three funding rate identifiers (one for each price identifier)?

There would be one new price feed implementation that each price identifier would use. Do I need separate bots for each index (9 bots), or can one bot handle multiple identifiers with the same price feed (3 bots)?

UMIPs/umip_add_priceidentifier_BITA.md
EndymionJkb: Since these are US currency values, I will specify 2 digit precision then (and have the price feed implementation do the rounding). The historical API actually returns full precision floating point as well; I just put two decimals in the example. I will change that as well to show what it really returns.

UMIPs/umip_add_priceidentifier_BITA.md
smb2796: One bot can handle multiple identifiers. You can also potentially use the same funding rate identifier for all three, if you make use of ancillary data.

UMIPs/umip_add_priceidentifier_BITA.md
```diff
@@ -0,0 +1,157 @@
...
+
+The DVM should support price requests for three fiat carbon indices: CGLCAR, CKLCAR, and CNLCAR, issued by Confluence Analytics, and computed and published
+by BITA (bitadata.com).
+
+# Motivation
+
+ESG (Environmental, Social, Governance) funds are high performing and growing in popularity in the traditional finance world, but there is currently
+no way to gain exposure to these types of funds in crypto.
+
+The first price identifiers are: CGLCAR, CKLCAR, and CNLCAR, corresponding to the global and regional low carbon indices from Confluence Analytics.
```
smb2796, line 21: Do you have more information about CGLCAR, CKLCAR and CNLCAR? Are these BITA specific indices? I'm having trouble finding much information about them.

EndymionJkb: BITA is where they're computed/published; Confluence Analytics is the index provider - determining constituents and weights. I'll include a link: https://www.confluenceanalytics.com/indexes.

therealhexi:
Other than the funding rate question, and price feed implementation, it looks good to me

smb2796:
Closing due to inactivity

PR #418, opened 2021-09-29
Add PieDAO DOUGH KPI options implementation

Reinis-FRP:

### DIFF
--- a/Implementations/piedao-dough.md
+++ b/Implementations/piedao-dough.md
@@ -1,0 +1,65 @@
+## Title
+
+Staked DOUGH Calculation
+
+## Summary
+
+PieDAO aims at launching KPI options to promote the new staking program to be released as of beginning October 2021. The KPI options will track the amount of DOUGH v2 staked at a given timestamp through a query to the subgraph provided, and convert it to a corresponding share from maximum KPI options payout.
+
+## Intended Ancillary Data
+
+```
+Metric:Total DOUGH v2 staked,
+Endpoint:"https://api.thegraph.com/subgraphs/name/pie-dao/vedough",
+Method:"https://github.com/UMAprotocol/UMIPs/blob/master/Implementations/piedao-dough.md",
+Key:data.globalStats[0].totalDoughStaked,
+Interval:request the last updated event at or before evaluation timestamp,
+EvaluationTimestamp:1635721589,
+Rounding:1
+```
+
+## Implementation
+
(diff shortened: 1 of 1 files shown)


Implementations/piedao-dough.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Intended Ancillary Data
+
+```
+Metric:Total DOUGH v2 staked,
+Endpoint:"https://api.thegraph.com/subgraphs/name/pie-dao/vedough",
+Method:"https://github.com/UMAprotocol/UMIPs/blob/master/Implementations/piedao-dough.md",
+Key:data.globalStats[0].totalDoughStaked,
+Interval:request the last updated event at or before evaluation timestamp,
+EvaluationTimestamp:1635721589,
```
smb2796, line 17: Why does the evaluation timestamp need to be baked into ancillary data? Could this not just be the expiry timestamp of the contract?

Implementations/piedao-dough.md
```diff
@@ -0,0 +1,65 @@
...
+In case of technical issues with subgraph raw metric should be resolved by querying on-chain value of DOUGH v2 token held by the staking contract:
+1. Identify the latest block at or before the `EvaluationTimestamp`.
+2. Call the `balanceOfAt` method on DOUGH v2 token contract `0xad32A8e6220741182940c5aBF610bDE99E737b2D` by passing the address of the staking contract `0x6Bd0D8c8aD8D3F1f97810d5Cc57E9296db73DC45` as the first parameter and identified block number from Step 1 as the second parameter.
+3. Take a note on the raw staked DOUGH v2 from the returned on-chain request.
+
+## Post processing
+
+Depending on the value of queried raw staked DOUGH v2 (this is before scaling down token decimals) proposers/voters should transform it to discrete share from maximum KPI options payout according to the table below:
+
+| Raw DOUGH v2 staked range                               | Human readable token amounts          | Resolved price |
```
smb2796, line 52: +100 on the separation between raw values and human readable values. Really like the distinction here.

smb2796 (approved):

Implementations/piedao-dough.md
Reinis-FRP: The KPI options program plans to measure staked DOUGH at the timestamp 1m after launch that is before the expiration. Effectively KPI options also work as vested rewards here.

PR #425, opened 2021-10-13
Add PoolTogether KPI options implementation

abg4:


Implementations/pooltogether-tvl.md
```diff
@@ -0,0 +1,93 @@
...
+Rounding:0,
+Scaling:0
+```
+***Note:** `totalLiquidityUSD` should be referenced for the timestamp that falls earlier but nearest to the price request timestamp.*  
+
+## Implementation
+
+With DeFiLlama:
+1. Voters should query the endpoint listed in Intended Ancillary Data.
+2. Voters should find the timestamp key in the `tvl` object that corresponds to the timestamp that is earlier but closest to the price request timestamp, and then use the corresponding `totalLiquidityUSD` value rounded to 0 decimal places.
```
Reinis-FRP, line 29:
```suggestion
2. Voters should find the timestamp key in the `tvl` object that corresponds to the latest daily timestamp that is at or before the price request timestamp, and then use the corresponding `totalLiquidityUSD` value.
```

1. The last returned timestamp by DeFiLlama usually is more granular than daily, and it might happen that different voters requesting the API at different times see different last timestamp. And if both of them are still earlier than request timestamps both voters would see different `totalLiquidityUSD` values. Hence, only the daily UTC midnight timestamps should be used.
2. If the request timestamp is midnight UTC then this should also be used (not day before).
3. If there are post-processing instructions for price resolvement then there is no need to round raw metric. This is relevant only if post-processing is done by FPL.

Implementations/pooltogether-tvl.md
```diff
@@ -0,0 +1,93 @@
...
+3. Multiply each balance returned from step 2 by the exchange rate of the respective Underlying Collateral Token listed above for the latest available block at or before the request timestamp. This returns the underlying collateral balance in USD.
+4. Each underlying collateral balance should be priced in USD for the same timestamp and the value should use the timestamp that falls earlier but nearest to the price request timestamp. DeFiLlama estimates this by using aggregated CoinGecko prices, but all voters should verify that results agree with broad market consensus.
+5. Calculate the sum of all underlying collateral balances in USD from Step 3.
+
+## Intended Application:
+
+The use case is to use KPI options to incentivize TVL. The initial payout function is:
+* If the returned value is equal to or greater than 500,000,000, voters should return 1.4.
+* If the returned value is less than 500,000,000, voters should return a value between 0.9 and 1.4 based on the ( TVL at expiry / 500,000,000 ) + 0.9.
```
Reinis-FRP, line 91: The same comment as in the DeFiLlama Implementation section above

Implementations/pooltogether-tvl.md
```diff
@@ -0,0 +1,93 @@
...
+
+It is intended to deploy the documented KPI option on Polygon using [LSP contract](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol) with `General_KPI` price identifier approved in [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md). The contracts would use [Linear LSP FPL](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol).
```
Reinis-FRP, line 93: Since this implementation is specific to the intended KPI options program better to also add `lowerBound` and `upperBound` for FPL settings.
And if it is known, then also planned `collateralPerPair`.
This would just make it easier for KPI option recipients to double-check their expected payouts.

Implementations/pooltogether-tvl.md
```diff
@@ -0,0 +1,93 @@
...
+  * https://thegraph.com/legacy-explorer/subgraph/pooltogether/pooltogether-v3_3_8
+  * https://thegraph.com/legacy-explorer/subgraph/pooltogether/pooltogether-v3_4_3
+* Celo Subgraph URL:
+  * https://thegraph.com/legacy-explorer/subgraph/pooltogether/celo-v3_4_5
+* BSC Subgraph URL:
+  * https://thegraph.com/legacy-explorer/subgraph/pooltogether/bsc-v3_4_3
+
+### General TVL Calculation:
+
+1. Construct subgraph query by making sure that the `block` parameter corresponds to the latest available block at or before the request timestamp, e.g.:
```
Reinis-FRP, line 49: The ancillary data specifies to use daily pricing information and also the results would differ from DeFiLlama which is updated daily.

Implementations/pooltogether-tvl.md
```diff
@@ -0,0 +1,93 @@
...
+    underlyingCollateralSymbol
+    underlyingCollateralToken
+    compoundPrizePool{
+      cToken
+    }
+  }
+}
+```
+2. Take a note of the prize pool contract addresses which are represented by `id`  and the `underlyingCollateralToken`. For networks that have multiple subgraph versions, remove duplicate query responses.
+3. Call the `accountedBalance` method on each prize pool contract from step 2 for the latest available block at or before the request timestamp. This will return the balance of controlled tokens (including timelocked) for each prize pool.
```
Reinis-FRP, line 65: Same comment on daily interval.

Implementations/pooltogether-tvl.md
```diff
@@ -0,0 +1,93 @@
...
+4. Scale down the balances returned from Step 3 with the decimals of the respective underlying collateral token (call the `decimals()` method on the underlying collateral token contracts from Step 2).
+5. Multiply each balance returned from step 4 by the exchange rate of its respective `underlyingCollateralToken` for the latest available block at or before the request timestamp. This returns the underlying collateral balance in USD.
+6. Each underlying collateral balance should be priced in USD for the same timestamp and the value should use the timestamp that falls earlier but nearest to the price request timestamp. DeFiLlama estimates this by using aggregated CoinGecko prices, but all voters should verify that results agree with broad market consensus.
```
Reinis-FRP, line 68: Same comment on daily interval.

Implementations/pooltogether-tvl.md
```diff
@@ -0,0 +1,93 @@
...
+
+PoolTogether TVL KPI options track value locked in the following Polygon prize pool contracts:
+* YieldSourcePrizePool (USDT Pool): [0x887E17D791Dcb44BfdDa3023D26F7a04Ca9C7EF4](https://polygonscan.com/address/0x887E17D791Dcb44BfdDa3023D26F7a04Ca9C7EF4)
+  * Underlying Collateral Token: [0xc2132D05D31c914a87C6611C10748AEb04B58e8F](https://polygonscan.com/address/0xc2132d05d31c914a87c6611c10748aeb04b58e8f)
+* YieldSourcePrizePool (USDC Pool): [0xEE06AbE9e2Af61cabcb13170e01266Af2DEFa946](https://polygonscan.com/address/0xee06abe9e2af61cabcb13170e01266af2defa946)
+  * Underlying Collateral Token: [0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174](https://polygonscan.com/address/0x2791bca1f2de4661ed88a30c99a7a9449aa84174)
+
+1. Call the `accountedBalance` method on each prize pool contract above for the latest available block at or before the request timestamp. This will return the balance of controlled tokens (including timelocked) for each prize pool.
```
Reinis-FRP, line 81: Same comment on daily interval.

Implementations/pooltogether-tvl.md
```diff
@@ -0,0 +1,93 @@
...
+2. Scale down the balances from Step 1 with the decimals of the respective underlying collateral token (call the `decimals()` method on the token contracts from the Underlying Collateral Token address above).
+3. Multiply each balance returned from step 2 by the exchange rate of the respective Underlying Collateral Token listed above for the latest available block at or before the request timestamp. This returns the underlying collateral balance in USD.
+4. Each underlying collateral balance should be priced in USD for the same timestamp and the value should use the timestamp that falls earlier but nearest to the price request timestamp. DeFiLlama estimates this by using aggregated CoinGecko prices, but all voters should verify that results agree with broad market consensus.
```
Reinis-FRP, line 84: Same comment on daily interval.

Implementations/pooltogether-tvl.md
```diff
@@ -0,0 +1,93 @@
...
+```
+***Note:** `totalLiquidityUSD` should be referenced for the timestamp that falls earlier but nearest to the price request timestamp.*  
+
+## Implementation
+
+With DeFiLlama:
+1. Voters should query the endpoint listed in Intended Ancillary Data.
+2. Voters should find the timestamp key in the `tvl` object that corresponds to the timestamp that is earlier but closest to the price request timestamp, and then use the corresponding `totalLiquidityUSD` value rounded to 0 decimal places.
+3. If the `totalLiquidityUSD` is equal to or greater than 500,000,000, voters should return 1.4.
+4. If the `totalLiquidityUSD` is less than 500,000,000, voters should return a value between 0.9 and 1.4 based on the calculation: ( `totalLiquidityUSD` / 500,000,000 ) + 0.9.
```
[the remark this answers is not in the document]
abg4: I think option b makes the most sense. Updating to (( `totalLiquidityUSD` / 500,000,000 ) / 2 ) + 0.9.

Implementations/pooltogether-tvl.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Intended Ancillary Data
+
+```
+Metric:TVL in PoolTogether financial contracts measured in USD,
+Endpoint:"https://api.llama.fi/protocol/pooltogether",
+Method:"https://github.com/UMAprotocol/UMIPs/blob/master/Implementations/pooltogether-tvl.md",
+Key:totalLiquidityUSD,
+Interval:Daily,
+Rounding:0,
```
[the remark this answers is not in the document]
abg4: Updated to 8. @smb2796 does this change make sense to you? And do you think this is a reasonable level of precision?

Implementations/pooltogether-tvl.md
abg4: Updated to be consistent with the change above.

Implementations/pooltogether-tvl.md
abg4: Updated to what we talked about using (daily midnight)

Implementations/pooltogether-tvl.md
abg4: Updated

Implementations/pooltogether-tvl.md
abg4: Updated

Implementations/pooltogether-tvl.md
abg4: Updated

Implementations/pooltogether-tvl.md
smb2796: So midnight utc aligns with the daily pricing interval of defillama right? It looks like it from api results.

smb2796: @Reinis-FRP @abg3090  with "Interval" do you think it is fitting to also specify the "time" that you want the daily queries to be at in the ancillary data? Maybe just specifying as "Interval:Daily 24:00 UTC"

Reinis-FRP: Agreed, just better to enclose the value in regular double quotes since 24:00 includes column so that it is consistent with ancillary data formatting specification.

Implementations/pooltogether-tvl.md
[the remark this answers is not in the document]
abg4: Updated to 6 based on our conversation.

Implementations/pooltogether-tvl.md
abg4: Added an example.

PR #437, opened 2021-11-08
Update IS_RELAY_VALID

Reinis-FRP:
- Updated LP fee rounding to floor
- Updated `BridgeAdmin` address
- Specified block number for calling `whitelistedTokens` and `depositContracts` on `BridgeAdmin`
- Added instructions for calculating utilization ratio from on-chain data
- Relay transactions mined before `quoteTimestamp` are invalid


UMIPs/umip-136.md
```diff
@@ -53,6 +53,10 @@ To get the true rate charged on these loans, we need to de-annualize it to get t
...
+
+Since the notion of `block.timestamp` might be different on other L2 chains compared to main ethereum L1 EVM, deposit contracts on L2 allow setting the `quoteTimestamp` within 10 minute range from current time. In order to deterministically calculate LP fees based on pool utilization ratio the relayer should wait till L1 block timestamp reaches the `quoteTimestamp`. Any relayed transaction that is mined with earlier block timestamp than `quoteTimestamp` field from `DepositData` struct in the `DepositRelayed` event should be invalid.
```
nicholaspai, line 58:
```suggestion
Since the notion of `block.timestamp` might be different on other L2 chains compared to main ethereum L1 EVM, deposit contracts on L2 allow setting the `quoteTimestamp` within 10 minute range from current time. In order to deterministically calculate LP fees based on pool utilization ratio the relayer should wait until L1 block timestamp reaches the `quoteTimestamp`. Any relayed transaction that is mined with earlier block timestamp than `quoteTimestamp` field from `DepositData` struct in the `DepositRelayed` event should be invalid.
```

nicholaspai (dismissed):
Looks great thanks @Reinis-FRP

nicholaspai (approved):
Looks great thanks Reinis

PR #439, opened 2021-11-10
Add QiDao MAI debt KPI options method doc

Reinis-FRP:


Implementations/qi-dao-mai-debt.md
```diff
@@ -0,0 +1,67 @@
...
+   curl -X POST \
+     -d '{"query": "{protocols(first: 1, block: {number: 21215934}) {totalBorrowed}}"}' \
+     'https://api.thegraph.com/subgraphs/name/0xlaozi/qi-dao-mai-finance-og'
+   ```
+4. Take a note on the raw borrowed MAI (`totalBorrowed` key value) from the returned subgraph response for each vault API endpoint and scale it down by 18 decimals.
+5. Add up all outstanding MAI debt amounts in all vaults across all chains from Step 4.
+
+## Post processing
+
+Since QiDao has opted to provide minimum guaranteed payout of 1 Qi per KPI options token and currently the audited financial product libraries do not allow setting minimum payout floor above zero voters should perform post-processing on the borrowed MAI and return a value between 1 and 2 to the submitted price request:
```
smb2796, line 57: This is fine for this implementation, but we really need to write an FPL to support these use cases as they are very common.

smb2796 (approved):
Looks really good once the subgraph endpoints are added. Great work!

PR #441, opened 2021-11-12
Implement 2Pi KPI options method

Reinis-FRP:


ChandlerDeKock (approved):
Look good

PR #447, opened 2021-11-30
Add XIO market cap KPI option method doc

abg4:

### DIFF
--- a/Implementations/xio-market-cap-rank.md
+++ b/Implementations/xio-market-cap-rank.md
@@ -25,17 +25,19 @@ XIO market cap is determined by comparing the market cap of XIO against other cr
 
 ```
 cmc_api_key = "YOUR_CMC_API_KEY"
+timestamp = "UNIX_timestamp" // uses UNIX format
 
 Query url:
-https://pro-api.coinmarketcap.com/v1/cryptocurrency/listings/historical?CMC_PRO_API_KEY={cmc_api_key}&date=1672603200&convert=XIO,USD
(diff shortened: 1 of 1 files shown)


Implementations/xio-market-cap-rank.md
```diff
@@ -0,0 +1,159 @@
...
+```
+/* Uses Axios for requests */
+const axios = require("axios");
+
+/* Set CMC and Cryptorank APIs */
+const cmc_api_key = "YOUR_CMC_API_KEY";
+const cryptorank_api_key = "YOUR_CRYPTORANK_API_KEY";
+
+/* Set CMC params */
+const cmc_date = "1637624697"; // uses UNIX format
```
smb2796, line 55: IMO this should be set as another variable that needs to be filled in, since really this could be used with any timestamp. Same with the cryptorank time below. Would just leave comments designating formatting for these two.

Implementations/xio-market-cap-rank.md
```diff
@@ -0,0 +1,159 @@
...
+6. The returned XIO Rank value represents the market cap rank of XIO.
+
+## Post processing
+
+The below parameters will be used to return a value between 0 and 100,000 to the submitted price request:
+- current_rank = the returned XIO market cap rank
+- highest_possible_rank = 100
+- lowest_possible_rank = 1,000
+
+1. If the returned current_rank value is below 100, the resolved price should be capped at 100,000.
```
smb2796, line 147:
```suggestion
1. If the returned current_rank value is below 100, the resolved price should be 100,000.
```

Implementations/xio-market-cap-rank.md
abg4: I agree, updated for timestamps to be set as variables.

PR #457, opened 2022-01-04
Use on-chain rate model parameters in IS_RELAY_VALID

Reinis-FRP:

### DIFF
--- a/UMIPs/umip-136.md
+++ b/UMIPs/umip-136.md
@@ -67,7 +67,7 @@ The call to `l1TokenRateModels` should return a stringified JSON object containi
 
 Rate model parameter values obtained above should be scaled down by 18 decimals.
 
-In case the call to `l1TokenRateModels` returns anything other that cannot be parsed to exact key-value pairs mentioned above (e.g. missing or extra parameter included) the relay should be considered invalid.
+In case the call to `l1TokenRateModels` returns anything that cannot be parsed to the exact key-value pairs mentioned above (e.g. missing or extra parameter included) the relay should be considered invalid.
 
 If the algorithm above doesn't produce a matching `realizedLPFeePct` (after floor rounding the result expressed as decimal to 18 decimals), the relay is invalid. If the token is not listed in the table above, the relay is invalid.
 
--- a/UMIPs/umip-136.md
+++ b/UMIPs/umip-136.md
@@ -59,14 +59,13 @@ Since the notion of `block.timestamp` might be different on other L2 chains comp
 
 Please see the example [implementation](https://github.com/UMAprotocol/protocol/blob/b588e83ca548a2a0d59b36f02ec9800afce28dec/packages/sdk/src/across/feeCalculator.ts#L78-L82)) for more details.
 
-The rates used the above computation are drawn from Aave (where applicable):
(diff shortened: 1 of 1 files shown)


UMIPs/umip-136.md
```diff
@@ -59,14 +59,13 @@ Since the notion of `block.timestamp` might be different on other L2 chains comp
...
-| ------ | --- | ---- | ---- | ---- |
-| ETH    | 65% | 0%   | 8%   | 100% |
-| USDC   | 80% | 0%   | 4%   | 60%  |
-| UMA    | 50% | 0%   | 5%   | 200% |
-| BADGER | 50% | 2.5% | 2.5% | 200% |
+The rate model parameters used for the above computation should be fetched on-chain from the `RateModelStore` contract deployed on mainnet at [0xd18fFeb5fdd1F2e122251eA7Bf357D8Af0B60B50](https://etherscan.io/address/0xd18fFeb5fdd1F2e122251eA7Bf357D8Af0B60B50) and governed by Across protocol [multi-signature account](https://etherscan.io/address/0xb524735356985d2f267fa010d681f061dff03715). Rate model parameters should be fetched by calling `l1TokenRateModels` method on the `RateModelStore` contract passing `l1Token` address as its argument at the block whose timestamp corresponds or is the latest available relative to the `quoteTimestamp` field.
+
+The call to `l1TokenRateModels` should return a stringified JSON object containing following key-value pairs:
```
nicholaspai, line 64: Do you think you need to include any language about what happens if `l1TokenRateModels` returns something that can't be parsed into a JSON or doesn't have exactly the UBar, R0, R1, and R2 params? (There are two cases in the latter case, the JSON has an extra param or the JSON has a missing param. I think both should be deemed erroneous)

nicholaspai (dismissed):
Thanks for doing this! I left a note wondering how we should handle incorrect output from the rate model contract

UMIPs/umip-136.md
```diff
@@ -59,14 +59,15 @@ Since the notion of `block.timestamp` might be different on other L2 chains comp
...
```
nicholaspai, line 64:
```suggestion
The call to `l1TokenRateModels` should return a stringified JSON object containing the following key-value pairs:
```

UMIPs/umip-136.md
```diff
@@ -59,14 +59,15 @@ Since the notion of `block.timestamp` might be different on other L2 chains comp
...
+* `UBar` corresponds to the <img src="https://render.githubusercontent.com/render/math?math=\bar{U}"> rate model parameter;
+* `R0`, `R1` and `R2` correspond to <img src="https://render.githubusercontent.com/render/math?math=R_0">, <img src="https://render.githubusercontent.com/render/math?math=R_1"> and <img src="https://render.githubusercontent.com/render/math?math=R_2"> parameters respectively.
+
+Rate model parameter values obtained above should be scaled down by 18 decimals.
+
+In case the call to `l1TokenRateModels` returns anything other that cannot be parsed to exact key-value pairs mentioned above (e.g. missing or extra parameter included) the relay should be considered invalid.
```
nicholaspai, line 70:
```suggestion
In case the call to `l1TokenRateModels` returns anything that cannot be parsed to the exact key-value pairs mentioned above (e.g. missing or extra parameter included) the relay should be considered invalid.
```

nicholaspai (dismissed):
Looks great, left some stylistic suggestions only!

mrice32 (approved):

PR #459, opened 2022-01-06
Add early expiration functionality to the YES_OR_NO_QUERY pid

smb2796:


UMIPs/umip-107.md
```diff
@@ -16,6 +16,7 @@ Price settlement can happen in three ways:
 - Return the `p1` value from ancillary data if the answer is "NO".
 - Return the `p2` value from ancillary data if the answer is "YES".
 - Return the `p3` value from ancillary data if the answer cannot be determined.
+- Return the `p4` value from ancillary data if the answer to the earlyExpirationCondition is no, and there is either an `earlyExpiration:1` key value pair present in ancillary data, or there is a specific last request timestamp listed in the ancillary data question and the active price request timestamp falls before that.
```
Reinis-FRP, line 19: The sentence above these bullets should now read "Price settlement can happen in four ways:"

UMIPs/umip-107.md
```diff
@@ -44,30 +45,40 @@ No price feed is needed (or possible) for this price identifier.
...
-q:Did the Dallas Mavericks win the NBA Finals in 2021?, p1:0, p2:1, p3:0.5
+q:Did the Dallas Mavericks beat the Miami Heat January 6th, 2022?, earlyExpirationCondition:Is the game over?, p1:0, p2:1, p3:0.5, earlyExpiration:1
 ```
 
 The q value should contain the question that UMA voters should answer with. 
 
 When this ancillary data dictionary is stored as bytes, the result would be: `0x713a446964207468652044616c6c6173204d6176657269636b732077696e20746865204e42412046696e616c7320696e20323032313f2c2070313a302c2070323a312c2070333a302e35`
 
-When returned, the `p1-p3` values should be scaled by 1e18. Typically voter dapps in the UMA ecosystem abstract this scaling away from voters, so, more often than not, voters should input the values as they appear in the ancillary data. Price requestors should be mindful of this, and not scale their ancillary data inputs.
+When returned, the `p1-p4` values should be scaled by 1e18. Typically voter dapps in the UMA ecosystem abstract this scaling away from voters, so, more often than not, voters should input the values as they appear in the ancillary data. Price requestors should be mindful of this, and not scale their ancillary data inputs.
```
pemulis: You wouldn't want to scale the `p4` value by `1e8`.

UMIPs/umip-107.md
```diff
@@ -44,30 +45,40 @@ No price feed is needed (or possible) for this price identifier.
...
+When converted from bytes to UTF-8, the ancillary data should be a dictionary object containing q (question), earlyExpirationCondition, p1, p2, p3 and p4 keys and values. p4 and earlyExpirationCondition are optional and only apply in certain situations.
 
 ```
-q:Did the Dallas Mavericks win the NBA Finals in 2021?, p1:0, p2:1, p3:0.5
+q:Did the Dallas Mavericks beat the Miami Heat January 6th, 2022?, earlyExpirationCondition:Is the game over?, p1:0, p2:1, p3:0.5, earlyExpiration:1
 ```
 
 The q value should contain the question that UMA voters should answer with. 
 
 When this ancillary data dictionary is stored as bytes, the result would be: `0x713a446964207468652044616c6c6173204d6176657269636b732077696e20746865204e42412046696e616c7320696e20323032313f2c2070313a302c2070323a312c2070333a302e35`
```
Reinis-FRP, line 56: better also to update hex string to be sync with question above

UMIPs/umip-107.md
```diff
@@ -44,30 +45,40 @@ No price feed is needed (or possible) for this price identifier.
...
 
-When returned, the `p1-p3` values should be scaled by 1e18. Typically voter dapps in the UMA ecosystem abstract this scaling away from voters, so, more often than not, voters should input the values as they appear in the ancillary data. Price requestors should be mindful of this, and not scale their ancillary data inputs.
+When returned, the `p1-p4` values should be scaled by 1e18. Typically voter dapps in the UMA ecosystem abstract this scaling away from voters, so, more often than not, voters should input the values as they appear in the ancillary data. Price requestors should be mindful of this, and not scale their ancillary data inputs.
+
+If there are no p1, p2, p3 or p4 values present, values should default to:
+- p1: 0
+- p2: 1
+- p3: 0.5
+- p4: -57896044618658097711785492504343953926634992332820282019728792003956564819968
```
Reinis-FRP, line 64: Is this magic number also to be scaled by 18 decimals?

UMIPs/umip-107.md
```diff
@@ -44,30 +45,40 @@ No price feed is needed (or possible) for this price identifier.
 
 ## Ancillary Data Specifications
 
-When converted from bytes to UTF-8, the ancillary data should be a dictionary object containing q (question), p1, p2 and p3 keys and values:
+When converted from bytes to UTF-8, the ancillary data should be a dictionary object containing q (question), earlyExpirationCondition, p1, p2, p3 and p4 keys and values. p4 and earlyExpirationCondition are optional and only apply in certain situations.
```
pemulis, line 48: I don't think we need `earlyExpirationCondition` but can explain that early expiration is valid if the question can be resolved. The specific conditions that must be met for the question to be resolved will vary but I don't think they need to be written out in each case.

In the example in the documentation, we could say that early expiration proposals can be accepted once the game is over but I don't think that needs to go in the ancillary data.

UMIPs/umip-107.md
```diff
@@ -44,30 +45,40 @@ No price feed is needed (or possible) for this price identifier.
...
 
 ```
-q:Did the Dallas Mavericks win the NBA Finals in 2021?, p1:0, p2:1, p3:0.5
+q:Did the Dallas Mavericks beat the Miami Heat January 6th, 2022?, earlyExpirationCondition:Is the game over?, p1:0, p2:1, p3:0.5, earlyExpiration:1
```
abg4, line 51: Would it make sense to include a p4 value for the example?

UMIPs/umip-107.md
```diff
@@ -44,30 +45,40 @@ No price feed is needed (or possible) for this price identifier.
...
+- p1: 0
+- p2: 1
+- p3: 0.5
+- p4: -57896044618658097711785492504343953926634992332820282019728792003956564819968
 
 # Rationale
 
 This construction sacrifices assurances of determinism in favor of greater price identifier flexibility. The places the burden of correct construction on price requesters, but, in return, allows for quicker and easier development without needing to pass through  UMA governance for each additional distinct query. This will allow for quite bespoke and speedy contract construction.
 
+p4 is intended to be used for situations where it is not a given that the price request (or contract settlement) should even occur yet. An example of this would be the UMA event based expiry LSP. A request to settle an event-based expiry LSP can be submitted at any time but may want to effectively be ignored. The requirements to return p4 should also be contained within the `q` value in the ancillary data.
```
pemulis, line 70: Proposed change:

`p4` is intended to be used for situations where it is not a given that the price request (or contract settlement) should even occur yet. An example of this would be the UMA event-based expiry LSP. A request to settle an event-based expiry LSP can be submitted at any time but if the question can not be resolved yet it should be ignored.

The default `p4` value is the minimum int256 value, and is used as a "magic number" to indicate that an event-based expiry proposal is invalid and the contract should continue as normal. For example, if the question is related to a basketball game on January 6th and a settlement request comes in on January 5th, the question can not be resolved yet, and voters should return the `p4` value with the magic number to reject the settlement request.

UMIPs/umip-107.md
```diff
@@ -44,30 +45,40 @@ No price feed is needed (or possible) for this price identifier.
...
+
 # Implementation
 
 1. Voters should decode the ancillary data and attempt to interpret the UTF-8 question.
-2. If UMA voters believes that the answer to the question is no, they should vote return the p1 value (in the example given, they would return `0`).
-3. If UMA voters believe that the answer to the question is yes, they should return the p2 value (in the example given, they would return `1`).
-4. If a voter cannot make a determination about what the correct answer to the question is, or there is no question present, UMA voters should return the p3 value (in the example given, they would return `0.5`).
-5. If there are no p1, p2, p3 values in the ancillary data voters should use 0, 1, 0.5 respectively.
-6. If there is no ancillary data or it is not interpretable to UTF-8, voters should return 0.5.
+2. Voters should first determine if this is an "early expiry" price request. This can either be designated in ancillary data by identifying that there is a key:value pair of `earlyExpiration:1` present, or by reading the question and determining that price request timestamp of the request is earlier than the final possible price request time noted in ancillary data.
```
pemulis, line 75: Proposed change (since all early expiry requests have this key-value pair appended):
2. Voters should first determine if this is an "early expiry" price request. This is the case when a key:value pair of `earlyExpiration:1` is present in ancillary data.

UMIPs/umip-107.md
```diff
@@ -44,30 +45,40 @@ No price feed is needed (or possible) for this price identifier.
...
+3. If this is an "early expiry" price request, voters should first answer the `earlyExpiryCondition` question. If the answer to this is no, voters should return the `p4` value. If the answer is yes, voters should continue the process to assess the `q` question.
```
pemulis, line 76: Proposed change:
3. If this is an "early expiry" price request, voters should first determine if the question in the ancillary data can be resolved definitively at this point in time. If not, voters should return the `p4` value. If yes, the voters should continue the process to assess the question.

UMIPs/umip-107.md
Reinis-FRP: in the LSP contract its checked against `type(int256).min`, so I assume the scaling should not be applied here

UMIPs/umip-107.md
pemulis, line 19: Proposed change (see more rationale below):
- Return the `p4` value from ancillary data if the answer to the question can not be definitively resolved yet and there is an `earlyExpiration:1` key value pair present in ancillary data.

smb2796: Thanks, updated

UMIPs/umip-107.md
smb2796: hm this is a good point and seems like a problem, since the canonical voter dapp would need to be updated to account for this

UMIPs/umip-107.md
smb2796: > early expiration is valid if the question can be resolved. The specific conditions that must be met for the question to be resolved will vary but I don't think they need to be written out in each case.

How would you determine the difference between p3 and p4 situations then?

UMIPs/umip-107.md
pemulis: @mrice32 see above about scaling values in the voter dapp for the early expiry "magic number"

UMIPs/umip-107.md
pemulis: If it's an early expiry attempt, return `p4`. If it's after the expiration timestamp, return `p3`.

pemulis: I'll add another comment that makes it more explicit that `p3` is never returned before the expiration timestamp and `p4` is never returned after the expiration timestamp.

UMIPs/umip-107.md
```diff
@@ -44,30 +45,42 @@ No price feed is needed (or possible) for this price identifier.
...
+
 # Rationale
 
 This construction sacrifices assurances of determinism in favor of greater price identifier flexibility. The places the burden of correct construction on price requesters, but, in return, allows for quicker and easier development without needing to pass through  UMA governance for each additional distinct query. This will allow for quite bespoke and speedy contract construction.
 
+p4 is intended to be used for situations where it is not a given that the price request (or contract settlement) should even occur yet. An example of this would be the UMA event-based expiry LSP. A request to settle an event-based expiry LSP can be submitted at any time but if the question can not be resolved yet it should be ignored.
+
+The default p4 value is the minimum int256 value, and is used as a "magic number" to indicate that an event-based expiry request is invalid and the contract should continue as normal. For example, if the question is related to a basketball game on January 6th and a settlement request comes in on January 5th, the question can not be resolved yet, and voters should return the p4 value with the magic number to reject the settlement request.
```
pemulis, line 72: Addendum:

Notice that a `p4` value would never be returned after the contract's expiration timestamp and a `p3` value would never be returned _before_ the contract's expiration timestamp. Consider an unresolvable question like, "Was the weather nice on January 6th, 2022?" If this question is asked on January 5th, 2022, you would return the `p4` value. If the question is asked on January 7th, 2022, you would return the `p3` value.

UMIPs/umip-107.md
smb2796: > since all early expiry requests have this key-value pair appended

This is not necessarily true. Requests that are not from LSPs do not have this appended necessarily

UMIPs/umip-107.md
smb2796: Or I guess return: -57896044618658097711785492504343953926634992332820282019728.792003956564819968, so 1e18 scaling works. seems funky though

UMIPs/umip-107.md
smb2796: So I want to leave this open to situations where someone could make YES_NO requests from contracts that are not the LSP and do not need to append ancillary data or distinguish between early/expiry requests

UMIPs/umip-107.md
smb2796: > I'll add another comment that makes it more explicit that p3 is never returned before the expiration timestamp and p4 is never returned after the expiration timestamp.

Price requests imo need to be able to be judged without reviewing the contract that is making the request. It is possible that these requests are made by contracts/eoa's that do not have expiration timestamps.

smb2796: Generally I think I agree though that we can remove `earlyExpirationCondition`! Will make those updates

UMIPs/umip-107.md
pemulis: Wack but could be a stopgap solution until the voter dapp is updated to recognize and handle the magic number.

UMIPs/umip-107.md
pemulis: > > I'll add another comment that makes it more explicit that p3 is never returned before the expiration timestamp and p4 is never returned after the expiration timestamp.
> 
> Price requests imo need to be able to be judged without reviewing the contract that is making the request. It is possible that these requests are made by contracts/eoa's that do not have expiration timestamps.

That should be fine! I think the early expiration tag is only stamped to the ancillary data when there _is_ an expiration timestamp, so anything without an expiration timestamp would be `p3` if unresolvable.

UMIPs/umip-107.md
pemulis: Ah, I see. Nevermind on this suggested change and a few others, then.

UMIPs/umip-107.md
smb2796: Same sort of comment as above. Not every requestor necessarily has an expiration timestamp. Can you generalize this somehow?

UMIPs/umip-107.md
pemulis: Nevermind on this suggestion since it is only correct for LSPs and contracts that append `earlyExpiration` in a similar way.

UMIPs/umip-107.md
smb2796: Used this and removed `earlyExpirationCondition`

UMIPs/umip-107.md
pemulis: Scratch that, based on feedback here. https://github.com/UMAprotocol/UMIPs/pull/459#discussion_r779741161

UMIPs/umip-107.md
pemulis: After reading your later replies, I may be wrong. Will review again when you finish the next round of updates.

UMIPs/umip-107.md
pemulis: Notice that a `p4` value would never be returned after the outcome, if resolvable, should be known, and a `p3` value would never be returned before that point. Consider an unresolvable question like, "Was the weather nice on January 6th, 2022?" If this question was asked on January 5th, 2022, you would return the `p4` value. If the question was asked on January 7th, 2022, you would return the `p3` value.

pemulis: Edited my comment above to (hopefully) make it flow better.

UMIPs/umip-107.md
```diff
@@ -12,10 +12,11 @@
 
 The DVM should support price requests for the `YES_OR_NO_QUERY` price identifier. `YES_OR_NO_QUERY` is intended to be used with ancillary data to allow anyone to request an answer to a "yes or no" question from UMA governance. This UMIP does not attempt to put any other restrictions on the content of the query, and instead leaves construction of the query up to the requester within ancillary data.
 
-Price settlement can happen in three ways:
+Price settlement can happen in four ways:
 - Return the `p1` value from ancillary data if the answer is "NO".
 - Return the `p2` value from ancillary data if the answer is "YES".
 - Return the `p3` value from ancillary data if the answer cannot be determined.
+- Return the `p4` value if the answer cannot be determined AND there is either an `earlyExpiration:1` key value pair present in ancillary data, or there is a specific last possible request timestamp listed in the ancillary data question and the active price request timestamp falls before that.
```
pemulis, line 19: `listed or implied`

Since people might specify a date rather than a timestamp in the ancillary data question and the oracle should still be able to resolve that by thinking a little bit.

UMIPs/umip-107.md
smb2796: Yeah I think this works well! Thank you

UMIPs/umip-107.md
```diff
@@ -44,30 +45,42 @@ No price feed is needed (or possible) for this price identifier.
...
-5. If there are no p1, p2, p3 values in the ancillary data voters should use 0, 1, 0.5 respectively.
-6. If there is no ancillary data or it is not interpretable to UTF-8, voters should return 0.5.
+2. Voters should first determine if this is an "early expiration" price request. This can either be designated in ancillary data by identifying that there is a key:value pair of `earlyExpiration:1` present, or by reading the question and determining that price request timestamp of the request is earlier than the final possible price request time noted in ancillary data.
+3. If this is an "early expiration" price request, voters should first determine if the question in the ancillary data can be resolved definitively at this point in time. If not, voters should return the p4 value. If yes, the voters should continue the process to assess the question.
+4. If UMA voters believes that the answer to the question is no, they should vote return the p1 value (in the example given, they would return `0`).
```
pemulis, line 79: `4. If UMA voters believe`

UMIPs/umip-107.md
smb2796: done, thank you

UMIPs/umip-107.md
pemulis: Reviewed with engineering and updated the documentation to reflect the stopgap solution. We can update the UMIP in the future after changing the code and design of the voter dapp to not scale to 18 decimals when the input is the magic number.

abg4 (approved):

PR #463, opened 2022-01-13
UMIP-create new GASETH price identifier for LSP contracts

rossgalloway:
This is an update and consolidation of the existing uGAS UMIPS to bring all the logic into one UMIP and update it to work with the LSP contracts.

- Compiles the rationale and motivation sections from previous UMIPs in to this one. No more jumping back to previous umips to understand what is going on.
- removed all the sushiswap/ self-referential price logic used for liquidation with the EMP contracts.
- Added an Ancillary data specification which lets a contract creator input a median price timeframe. To keep things simple, I kept the previous options (1HR, 4HR, 1D, 7D, 30D) but I did change the way they are described since a 1 month median is kind of useless since months are different lengths. Values in ancillary data should be rounded to the nearest approved timeframe (the options above) and if there is no data it defaults to 30 day.

Warrants review to make sure that some of these changes didnt break anything with the scripts or logic, which I have not tested.

Also, apologies for the shitshow of commits in the pull request. I still can't quite figure out how to use github cleanly.


UMIPs/GASETH-LSP.md
```diff
@@ -0,0 +1,198 @@
...
+
+# ANCILLARY DATA SPECIFICATIONS
+
+When converted from bytes to UTF-8, the ancillary data should be a number `T`.
+
+i.e. ``` T:720 ```
+
+`T` is the total number of hours over which the median gas price is calculated. (The above number shows a 30 day median.) In order to work with the previous calculations, this number should be rounded to the nearest previously approved median time period (in whole hours) as seen in the *Rationale* section. This number is to then be used to determine the minimum number of blocks used to calculate the median gas price over that time period as described in that section and the *implementation* section.
+
+When the above example ancillary data is stored as bytes, the result would be: `insert`
```
Reinis-FRP, line 60:
```suggestion
When the above example ancillary data is stored as bytes, the result would be: `0x543A373230`
```

UMIPs/GASETH-LSP.md
```diff
@@ -0,0 +1,198 @@
...
+
+    for gas_price, gas_used in sorted_TXes:
+        rolling_sum += gas_used
+        if rolling_sum > total_gas/2:
+            return gas_price
+```
+
+An alternative method, for a DVM reporter to calculate this exact data point, would be to run the below SQL query against a public dataset of Ethereum node data such as Google BigQuery's `bigquery-public-data.crypto_ethereum.transactions` dataset. 
+
+This query is parameterized with a UTC timestamp `t1`, a time range defined by the price identifier being used (i.e. 24HR) `N`, a lower bound of time `t2` defined by `t1 - N * 3600` and a minimum block amount `B` (explained in the *Rationale* section).
```
Reinis-FRP, line 134: Its confusing why time range here is denoted by `N` while in ancillary data section it is the value of `T`.

Also if minimum block amount `B` is the same as "Minimum number of mined blocks" from Rationale section then it would be consistent to name this variable `N` here.

UMIPs/GASETH-LSP.md
```diff
@@ -0,0 +1,198 @@
...
+**The rest of this section is replicated from the *Implementation section* in [UMIP-129](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-129.md). **
+
+Updated rounding: 6 decimals
+
+In order to account for the changes to the fee market included in EIP-1559, the following pseudo-algorithm is used to calculate the exact data point to report by a DVM reporter:
+
+```python
+def return_median(t, N)
+    assert N in [1, 4, 24, 168, 720]
+    minimum_ranges = {1: 200, 4: 800, 24: 4800, 168: 33600, 720: 134400} # a mapping between the durations 1HR, 4HR, 24HR (1D), 168HR (1W), 720HR(1M) and the corresponding mimimum number of blocks that must have been mined within the period.
```
Reinis-FRP, line 100: N for 720H in the Rationale section is 144000 (not 134400)

UMIPs/GASETH-LSP.md
```diff
@@ -0,0 +1,198 @@
...
+
+**2. Base Currency** - GAS
+
+**3. Quote currency** - ETH
+
+**4. Intended Collateral Currency** - WETH
+
+**5. Scaling Decimals** - 18 (1e18)
+
+**6. Result Processing** - Exact. Multiply by a million.
```
Reinis-FRP, line 48: Please also address multiplying by a million in the Implementation section.

Just to be clear - is it expected that voting dApp would also scale this value up by 18 decimals?

UMIPs/GASETH-LSP.md
```diff
@@ -0,0 +1,198 @@
...
+# IMPLEMENTATION
+DVM voters should use the timestamp from the contract that is being voted on. This timestamp can either be queried from the `expirationTimestamp` field in the LSP contract or from the `timestamp` field in the data of the `requestPrice` price function when `expire` is called. 
+
+**The rest of this section is replicated from the *Implementation section* in [UMIP-129](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-129.md). **
+
+Updated rounding: 6 decimals
+
+In order to account for the changes to the fee market included in EIP-1559, the following pseudo-algorithm is used to calculate the exact data point to report by a DVM reporter:
+
+```python
```
smb2796, line 97: Hmmm, I thought the plan was to simplify this calculation methodology by using something like a twap'd median of a few gas price providers. Did this turn out to not be feasible, or is this still the plan?

UMIPs/GASETH-LSP.md
```diff
@@ -0,0 +1,198 @@
...
+
+**6. Result Processing** - Exact. Multiply by a million.
+
+**7. Rounding** - Round to nearest 6 decimal places (seventh decimal place digit >= 5 rounds up and < 5 rounds down)
+
+# ANCILLARY DATA SPECIFICATIONS
+
+When converted from bytes to UTF-8, the ancillary data should be a number `T`.
+
+i.e. ``` T:720 ```
```
smb2796, line 56: Don't feel super strongly about this, but other UMIPs almost always specify time in seconds. Any particular reason you spec'd in hours?

rossgalloway: I can change to seconds. I have no preference. Hours seemed more intuitive for a casual reader, but seconds makes sense.

UMIPs/GASETH-LSP.md
```diff
@@ -0,0 +1,198 @@
...
+
+Updated table clarifying median price durations to use for this UMIP.
+
+ | Identifier | Number of Hours Contained (T) | Minimum number of mined blocks (N) |
+ |-------------|--------------|----------------------------------|
+ | GASETH-1HR | 1 | 200 |
+ | GASETH-4HR | 4 | 800 |
+ | GASETH-1D | 24 | 4800 |
+ | GASETH-7D | 168 | 33600 |
+ | GASETH-30D | 720 | 144000 |
```
rossgalloway, line 84: @smb2796 Here is the clarifying info I promised from our call about line 84.

This chart originated from [UMIP-16](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-16.md), where it gives a minimum mined block value for GASETH-1M[onth] of `134400`. If we look at the expected number of blocks per hour from the other timeframes, we see it is 200 blocks per hour. If we assume that this 1M value should stay consistent, then this month is really 28 days. See chart below:
image
This seems arbitrary and generally too short since there are typically about 30.4 days per month. I have therefore changed this number to reflect a 30 day period, which would call for 144000 blocks mined in this period if we keep a consistent 200 blocks/hour from the rest of the table.

Does this logic make sense? If so I will update line 100 of the script

UMIPs/GASETH-LSP.md
rossgalloway: Per our call, I am keeping this as hours to limit the changes needed to the script.

UMIPs/GASETH-LSP.md
rossgalloway: Per discussion with @smb2796 I will remove this scaling factor and leave this to the LSP contract implementation.

UMIPs/GASETH-LSP.md
rossgalloway: See 

> N for 720H in the Rationale section is 144000 (not 134400)

I will change this to 144000, but would like confirmation that my logic is correct as I have changed this number from the original UMIP. comment here: https://github.com/UMAprotocol/UMIPs/pull/463/files#r787112134

UMIPs/GASETH-LSP.md
```diff
@@ -0,0 +1,198 @@
...
+In order to account for the changes to the fee market included in EIP-1559, the following pseudo-algorithm is used to calculate the exact data point to report by a DVM reporter:
+
+```python
+def return_median(t, N)
+    assert N in [1, 4, 24, 168, 720]
+    minimum_ranges = {1: 200, 4: 800, 24: 4800, 168: 33600, 720: 134400} # a mapping between the durations 1HR, 4HR, 24HR (1D), 168HR (1W), 720HR(1M) and the corresponding mimimum number of blocks that must have been mined within the period.
+    
+    # `t` is number of seconds since the epoch (unix time)
+    end_block = block.at(timestamp=t) # rounded down
+    offset = N - 3600*N # there are 3600 seconds/hour
```
rossgalloway, line 104: Shouldn't this line read `offset = t - 3600*N` ?

UMIPs/GASETH-LSP.md
rossgalloway: I will change the ancillary data variable name to N to match with the existing scripts. I will also update the table in rationale to reflect this

UMIPs/GASETH-LSP.md
rossgalloway: fixed, but different bytes due to variable being `N` now and not `T`

UMIPs/GASETH-LSP.md
```diff
@@ -0,0 +1,194 @@
...
+
+```python
+def return_median(t, N)
+    assert N in [1, 4, 24, 168, 720]
+    minimum_ranges = {1: 200, 4: 800, 24: 4800, 168: 33600, 720: 144000} # a mapping between the durations 1HR, 4HR, 24HR (1D), 168HR (1W), 720HR(1M) and the corresponding mimimum number of blocks that must have been mined within the period.
+    
+    # `t` is number of seconds since the epoch (unix time)
+    end_block = block.at(timestamp=t) # rounded down
+    offset = t - 3600*N # there are 3600 seconds/hour
+    start_block = block.at(timestamp=(t-offset))
```
Reinis-FRP, line 101: `offset` is the evaluation period start timestamp (`t - 3600*N`), but why in `start_block` formula it searches for the block for timestamp (`t-offset`) as that would equate to 3600*N?

UMIPs/GASETH-LSP.md
smb2796: Yes this makes sense to me

UMIPs/GASETH-LSP.md
rossgalloway: @Reinis-FRP, @smb2796, I changed this in the most recent commit. I think that is the last open item, assuming you agree with the change.

smb2796: Yes, sorry I missed this before and completely agree with your edits! Offset should simply be hours (N) converted to seconds. I'm really not sure why we initially created the mapping as hours at all and not just seconds (as this seems more precise anyways)

smb2796 (approved):

UMIPs/GASETH-LSP.md
rossgalloway: These queries assume a 30 day (well, 28 but we are changing that) median price for the synth. How do you typically deal with updating these scripts to take input from ancillary (or other variable) data? New variable?

smb2796: I'm happy to do it if you want, will probably just need to be reminded!

PR #464, opened 2022-01-13
UMIP- create new PUNKETH price identifier for LSP contracts

rossgalloway:
This is an update and consolidation to the PUNKETH price identifier for use with LSP contracts. It is built around the PUNKETH price identifier (UMIP 84), but without the AMM based self referential price logic for liquiditations or requirement to specify am expiry within the price identifier used in UMIP-130. The price script is taken from UMIP-130

Like with the new uGAS price identifier, this warrants review to make sure that some of these changes didn't break anything with the scripts or logic, which I have not tested.

Also, apologies for the shitshow of commits in the pull request. I still can't quite figure out how to use github cleanly.


UMIPs/PUNKETH_LSP.md
```diff
@@ -0,0 +1,185 @@
...
+A Python implementation of this pseudo-code can be found [here](https://gist.github.com/cc7768/6fd4e97aa39a0a84ae2bb00085602c9e#file-upunk_index_price-py)
+
+
+# Technical Specifications
+
+- Price identifier name: `PUNKETH_LSP`
+- Base Currency: CryptoPunk NFTs
+- Quote Currency: ETH
+- Rounding: Round to 6 decimal places (seventh decimal place digit >= 5 rounds up and < 5 rounds down)
+- Estimated current value of post-expiry price identifier script: `enter after running`
```
Reinis-FRP, line 137: Please fill this value

UMIPs/PUNKETH_LSP.md
```diff
@@ -0,0 +1,185 @@
...
+
+* _CryptoPunks_: As mentioned earlier in this document, we chose to build an index using CryptoPunks because they were the original NFT. This originality has lead to them being highly valued and having consistent enough trade volume.
+* _30 day median_: The 30 day median allows for the index to reflect common trading prices across many CryptoPunks rather than to respond to particular transactions
+* _Unique CryptoPunks_: We only use the most recent trade price for each CryptoPunk. This is a security feature since if we used each transaction then a single person could trade one CryptoPunk amongst accounts they owned to manipulate the price.
+* _Median rather than the mean_: Calculating the mean incorporates the price of every single transaction which means that someone who owned a single CryptoPunk could have a small effect on the price. The median can still be manipulated but, given the uniqueness restriction above, it would require someone to own enough CryptoPunks to make up half of the monthly transactions.
+
+
+# Implementation
+
+### After Expiry
```
Reinis-FRP, line 152: There is no need for separate sub-section as LSP price requests come only at expiration

UMIPs/PUNKETH_LSP.md
```diff
@@ -0,0 +1,185 @@
...
+- Price identifier name: `PUNKETH_LSP`
+- Markets & Pairs: CryptoPunk Market contract `PunkBought` events. The CryptoPunk contract address is `0xb47e3cd837dDF8e4c57F05d70Ab865de6e193BBB` which you can see at https://etherscan.io/address/0xb47e3cd837ddf8e4c57f05d70ab865de6e193bbb
+- Example price providers: Infura and The Graph include information on CryptoPunk contract events
+- Cost to use: [Infura](https://infura.io/) supports up to 100,000 requests per day for free. Information also available on [The Graph](https://thegraph.com/)
+- Real-time price update frequency: Updated every block
+- Historical price update frequency: Updated every block
+
+_Note_: An important fact worth noting is that the original `CryptoPunkMarket` contract has a bug in it that results in CryptoPunk sales that are generated by "accepting a bid" to assign a value of 0 to the `value` field of the `PunkBought` event. We address this in our price feed but, if you choose to write your own implementation, it is important to be aware of this.
+
+# Price Feed Implementation
```
smb2796, line 48: Are there any desired ancillary data specifications? If so, you should create a separate section to explain these. Examples of this can be seen in previous umips, like [here](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md#ancillary-data-specifications). 

I would think that desired parameterization might include period of time (making hardcoded 30 day spec dynamic) and perhaps scaling factor. 

Maybe these are unnecessary, which is also fine. If you do want to include however, can you parameterize the pseudocode to use the ancillary data params?

UMIPs/PUNKETH_LSP.md
```diff
@@ -0,0 +1,185 @@
...
+
+# Compute the prices used for each event
+events_corrected = []
+for event in events:
+    # Get the event that generated the transaction
+    transaction = get_transaction(web3, event)
+
+    # Decode input data to find function that
+    # generated the event
+    fn = cryptopunk.decode_input(transaction.input)["function_name"]
```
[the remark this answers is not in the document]
rossgalloway: I have to defer to @cc7768 on this one as I did not write the code.

UMIPs/PUNKETH_LSP.md
rossgalloway: overlooked this, Will delete

UMIPs/PUNKETH_LSP.md
rossgalloway: Being able to set the median price time period would be good. I'll work on adding that in.

What would the scale factor do? Scale the output to work like leverage?

UMIPs/PUNKETH_LSP.md
```diff
@@ -0,0 +1,185 @@
...
+| Created             | Jan. 12, 2022                                                                  |
+| Discourse Link      | insert |
+
+
+# Summary
+
+This UMIP re-introduces the price identifier previously called `PUNKETH` and updates it for use in an Long/Short Pair token referred to as `uPUNKS`. The token is a synthetic index based on the recent trading prices of CryptoPunks. The previous price identifier ```PUNKETH-1221``` found in [UMIP 130](./umip-130.md) is not being re-used due to the fact that it was built for the EMP contract. This UMIP re-uses the logic of the previous ```PUNKETH``` price identifier from [UMIP 84](./umip-84.md) and uses the expiry timestamp per contract deployment to allow the creation of multiple overlapping uPUNK contracts. The ```PUNKETH_TWAP``` price identifier is ignored as the LSP is un-liquidatable and this logic is not necessary.
+
+The DVM should support requests for a price that resolves to the median most recent purchase price (in ETH) of each unique CryptoPunk traded in the last 30 days. The price resolution method to use will depend on the the timestamp the price request was made at.
```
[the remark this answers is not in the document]
smb2796: Yes it would!

UMIPs/PUNKETH_LSP.md
rossgalloway: as discussed in the meeting on 220118, I will not include the scaling factor and remove the scaling from the script. Any scaling will happen on the LSP side of things.

I will add the timeframe into ancillary data.

UMIPs/PUNKETH_LSP.md
```diff
@@ -0,0 +1,185 @@
+## Headers
+
+| UMIP-xxx            |                                                                                 |
+| ------------------- | ------------------------------------------------------------------------------- |
+| UMIP Title          | Create new PUNKETH_LSP price identifier                                  |
```
Reinis-FRP, line 5: just for consistency, have you considered replacing "_" to "-" as in GASETH-LSP?

UMIPs/PUNKETH_LSP.md
```diff
@@ -0,0 +1,196 @@
...
+# Price Feed Implementation
+
+The price feed methodology is the same as that of [UMIP-84](./umip-84.md) but with the median price timeframe passed in from ancillary data. The Pseudocode for calculating the price after expiry is shown below:
+
+```
+# Get the PunkBought Events from the cryptopunk contract
+# for the last T seconds (from ancillary data)
+events = getEvents(
+    w3, cryptopunk, "PunkBought",
+    block_TSecondsAgo, block_now
```
rossgalloway, line 69: I changed this line. `block_30daysAgo` -> `block_TSecondsAgo`

UMIPs/PUNKETH_LSP.md
```diff
@@ -0,0 +1,196 @@
...
+        price_gt_0 = event.price > 0
+        if more_recent and price_gt_0:
+            cryptopunk_blockprice[event.punkIndex] = {"block": event.blockNumber, "price": event.price}
+
+# Take median of (most recent) prices
+median([cryptopunk_blockprice[cryptopunk]["price"] for cryptopunk in cryptopunks])
+
+```
+
+A Python implementation of this pseudo-code can be found [here](https://gist.github.com/cc7768/6fd4e97aa39a0a84ae2bb00085602c9e#file-upunk_index_price-py)
```
rossgalloway, line 140: I have not updated the python script. This is outside my abilities.

UMIPs/PUNKETH_LSP.md
rossgalloway: What is the point of this line? Just as a gut check? This data is out of date the moment it gets added.

Reinis-FRP: Yes, this is mostly to check if value aligns with what is expected. Also better to provide timestamp for which the script was run so that others could verify it

UMIPs/PUNKETH_LSP.md
```diff
@@ -0,0 +1,196 @@
...
+- Rounding: Round to 6 decimal places (seventh decimal place digit >= 5 rounds up and < 5 rounds down)
+- Estimated current value of post-expiry price identifier script: `enter after running`
+
+
+# Rationale
+
+The `PUNKETH_LSP` price identifier had a few decisions that we believe were important to the design:
+
+* _CryptoPunks_: As mentioned earlier in this document, we chose to build an index using CryptoPunks because they were the original NFT. This originality has lead to them being highly valued and having consistent enough trade volume.
+* _30 day median_: The 30 day median allows for the index to reflect common trading prices across many CryptoPunks rather than to respond to particular transactions. This is the default median timeframe, and can be changed by inputting a different `T` variable into the contract ancillary data.
```
Reinis-FRP, line 157: Discussion on default 30d if T is missing should be covered in the Implementation section

rossgalloway: let me know if the updated language is acceptable

UMIPs/PUNKETH_LSP.md
smb2796: From @cc7768 - this is basically already done, since you can just overwrite the window value here: https://gist.github.com/cc7768/6fd4e97aa39a0a84ae2bb00085602c9e#file-upunk_index_price-py-L164

UMIPs/PUNKETH_LSP.md
smb2796: I think it looks good.

UMIPs/PUNKETH_LSP.md
rossgalloway: I changed it!

smb2796 (approved):

PR #470, opened 2022-01-25
Add Jarvis LP TVL KPI Option method doc

abg4:


Implementations/jarvis-lp.md
```diff
@@ -0,0 +1,53 @@
...
+This calculation is intended to track the TVL in USD of liquidity provider (LP) tokens staked in the [Jarvis WETH-JRT Sushiswap pool on Polygon](https://polygonscan.com/address/0xa769c6786c3717945438d4c4feb8494a1a6ca443).
+
+## Intended Ancillary Data
+
+```
+Metric:LP TVL provided to Jarvis WETH-JRT pool on Polygon,
+TVLCurrency:usd,
+Method:"https://github.com/UMAprotocol/UMIPs/blob/master/Implementations/jarvis-lp.md",
+jarvisLPContract:0xa769c6786C3717945438d4C4feb8494a1a6Ca443,
+Interval:daily,
```
Reinis-FRP, line 16: Does this imply price intervals should be rounded down to 00:00:00 UTC? Or can use the latest available block at or before the request timestamp?

Implementations/jarvis-lp.md
```diff
@@ -0,0 +1,53 @@
...
+2. Identify the LP contract reserve currencies by calling `token0()` and `token1()` methods on the LP contract from step 1.
+3. Get total LP reserves by calling `getReserves()` method on the LP contract at the latest available block at or before the evaluation timestamp. This should return `token0` and `token1` balances as index 0 `_reserve0` and index 1 `_reserve1` respectively.
+4. For both LP reserve tokens from Step 2, get the latest available pricing before the evaluation timestamp from CoinGecko:
+    * Based on CoinGecko [API documentation](https://www.coingecko.com/api/documentations/v3#/contract/get_coins__id__contract__contract_address__market_chart_range) construct price API request with following parameters:
+      * `id`: CoinGecko platform id - the chain of the LP contract from Step 1 in https://api.coingecko.com/api/v3/asset_platforms endpoint ("polygon-pos" is used for the Jarvis LP calculation);
+      * `contract_address`: reserve token address from Step 2;
+      * `vs_currency`: TVL measurement currency based on the passed `TVLCurrency` parameter in the ancillary data (e.g. "usd");
+      * `from`: start timestamp;
+      * `to`: request timestamp;
```
Reinis-FRP, line 35: Where does start timestamp come from as its not in ancillary data?
But if KPI options should just track spot TVL (no need for TWAP) then `from` can be calculated as request timestamp less fixed offset. In order to ensure consistent results independent on when results are verified its better to set offset to 30 days so that CoinGecko granularity is fixed at hourly intervals.

Implementations/jarvis-lp.md
```diff
@@ -0,0 +1,53 @@
...
+    * Example API request for JRT token pricing information: https://api.coingecko.com/api/v3/coins/polygon-pos/contract/0x596ebe76e2db4470966ea395b0d063ac6197a8c5/market_chart/range?vs_currency=usd&from=1642226400&to=1642917600
+    * Example API request for WETH token pricing information: https://api.coingecko.com/api/v3/coins/polygon-pos/contract/0x7ceb23fd6bc0add59e62ac25578270cff1b9f619/market_chart/range?vs_currency=usd&from=1642226400&to=1642917600
+    * Locate the `prices` key value from CoinGecko API response - it should contain a list of [ timestamp, price ] values. Choose the price at the latest timestamp before the evaluation timestamp (CoinGecko timestamps are in milliseconds);
+    * Voters should verify that obtained price results agree with broad market consensus.
+5. Scale down LP reserve balances from Step 3 with their respective decimals (call `decimals()` method on the token contracts from Step 2).
+6. Multiply each LP reserve token balance from Step 5 with its price from Step 4.
+7. Sum both LP reserve balance values from Step 6 to get the total value of the pool in USD at the latest available block at or before the evaluation timestamp.
```
Reinis-FRP, line 43: If `TVLCurrency` is passed in ancillary data then should replace USD here

Implementations/jarvis-lp.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Intended Application
+
+It is intended to deploy the documented KPI options separately on each supported chain using [LSP contract](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol) with `General_KPI` price identifier approved in [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md). The contracts would use [Linear LSP FPL](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol) with the `lowerBound` set to 0 and `upperBound` set to 2,300,000.
```
Reinis-FRP, line 47: If deployment on multiple chains is planned would this still track LP TVL on Polygon?

Implementations/jarvis-lp.md
abg4: Unless Sean disagrees, I think we can use the latest available block at or before the request timestamp. Let me know if you agree with the update I made to the interval.

Implementations/jarvis-lp.md
abg4: Good call, I updated `from` under the assumption we are not using a TWAP.

Implementations/jarvis-lp.md
abg4: Updated to use `TVLCurrency`.

Reinis-FRP (dismissed):

Implementations/jarvis-lp.md
smb2796: I think it is likely that we will need to use a TWAP and should design for that after confirming with Jarvis

Implementations/jarvis-lp.md
```diff
@@ -0,0 +1,53 @@
...
+    * Voters should verify that obtained price results agree with broad market consensus.
+5. Scale down LP reserve balances from Step 3 with their respective decimals (call `decimals()` method on the token contracts from Step 2).
+6. Multiply each LP reserve token balance from Step 5 with its price from Step 4.
+7. Sum both LP reserve balance values from Step 6 to get the total value of the pool in terms of `TVLCurrency` at the latest available block at or before the evaluation timestamp.
+
+## Intended Application
+
+It is intended to deploy the documented KPI options on Polygon using [LSP contract](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol) with `General_KPI` price identifier approved in [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md). The contracts would use [Linear LSP FPL](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol) with the `lowerBound` set to 0 and `upperBound` set to 2,300,000.
```
smb2796, line 47: If I remember correctly, there is a desire to have a step function and not a strictly linear payout here. As there is no current way to bake this into our FPLs, we should account for this in `Implementation` and `Intended Application`.

Summarizing previous conversation around that:
- At expiry, 1M JRT is given out per 1M of liquidity.
- For the KPI option, payout should not be per $, but per tranche of $1M.
- So if there is $1.2M of liquidity, the KPI will deliver 1M JRT.
- $1.9M would also be 1M JRT
- But if we have $2.01M would deliver 2M

smb2796 (changes requested):
Before merging, we should confirm and update accordingly:
- Use of TWAP
- Stepwise vs linear payout function

Implementations/jarvis-lp.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+Metric:LP TVL provided to Jarvis WETH-JRT pool on Polygon,
+TVLCurrency:usd,
+Method:"https://github.com/UMAprotocol/UMIPs/blob/master/Implementations/jarvis-lp.md",
+jarvisLPContract:0xa769c6786C3717945438d4C4feb8494a1a6Ca443,
+Interval:daily,
+Aggregation:Average end of day (midnight UTC) LP TVL since <START_TIMESTAMP>,
+Rounding:0,
+TVLCheckpoints:{"0":0,"1000000":1000000,"2000000":2000000,"3000000":3000000,"4000000":4000000}
```
smb2796, line 19: What are these tvl checkpoints used for? Since "payout value" increases proportionally to TVL threshold, do you think we could just instruct voters to round TVL value down to nearest million? Or do you want to create more flexibility by passing these in?

smb2796 (dismissed):

smb2796:
Personally think this looks really good! Just had one clarification question about the ancillary data

Implementations/jarvis-lp.md
abg4: Yea, good question. I think we are clear in the `Intended Application` section on how voters should round TVL down to the nearest million. I saw the `TVLCheckpoints` parameter in the YEL implementation doc so main reasons for including was consistency with other TVL implementations and being explicit on the thresholds. I can easily remove it if you don't think it adds value being included here?

Reinis-FRP: `TVLCheckpoints` was used for YEL because payouts per checkpoints were not linear.
If that is not required and checkpoints are in round millions then its simpler to use negative `Rounding` parameter (-6) as discussed in [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md). Just check if normal rounding (up from 5) works here. If you need floor rounding than this should be explicitly stated in the implementation.

abg4: Makes sense. I removed `TVLCheckpoints ` and updated the `Rounding` parameter to -6. We need to use floor rounding so I added a step to the implementation.

smb2796 (dismissed):

Reinis-FRP (approved):

PR #475, opened 2022-01-27
Approve Proposer contract to interact with DVM

mrice32:
Adds an UMIP to approve the Proposer contract to request and retrieve prices from the DVM.


Reinis-FRP:
This looks good and its nice that adding/removing contract creator permission for the Governor contract is explicitly included here.

Reinis-FRP (approved):

PR #481, opened 2022-02-13
Add UMA to polygon as collateral currency

inalittlewhile:


smb2796:
@inalittlewhile before this can be merged in, the PR needs to be fixed. You're currently editing multiple files bc it looks like you were reverting commits. Suggest closing this PR and starting clean with a new branch off of master.

UMIPs/Add UMA as Collateral Currency on Polygon.md
```diff
@@ -0,0 +1,38 @@
...
+## Motivation
+
+The addition of this collateral currency offers additional functionality to the UMA protocol and increases the range of contracts that can be built.
+
+## Technical Specification
+
+To accomplish this upgrade, the following changes need to be made:
+
+-   The UMA address https://polygonscan.com/token/0x3066818837c5e6ed6601bd5a91b0762877a6b731 needs to be added to the collateral currency whitelist introduced in UMIP-8.
+-   A final fee of 250 UMA needs to be added in the Store contract.
```
smb2796, line 22: The UMA final fee on mainnet is only 90. 250 is more in line, but you should propose updating the mainnet final fee as well in this umip

PR #484, opened 2022-02-15
Add uma as collateral currency to polygon

inalittlewhile:
The mistake on my part was committing changes to my master repo instead of committing them to new branches. I had to revert my master repo. So all new branches I open will show these reverts in a PR until they are pushed. This is my understanding. Is there another manner that would be better to proceed here? Thank you.


inalittlewhile:
What a mess. Got it and will do. Thanks Sean.

PR #487, opened 2022-02-17
docs: Add umip-151.md VolatilityDAOracle

rforte:
Co-authored-by: wadek2 <8403717+wadek2@users.noreply.github.com>


UMIPs/umip-151.md
```diff
@@ -0,0 +1,86 @@
...
+* `window` - This is the time window to which the methodology is applied. Window uses standardized time formats, where lower-denomination units are lowercase and higher-denominations are uppercase:
+    * `s` = seconds
+    * `m` = minutes
+    * `h` = hours
+    * `d` = Days
+    * `W` = Weeks
+    * `M` = Months
+    * `Y` = Year
+    * For example, 14d means 14 Day volatility.
+targetAsset - This is the asset to which volatility is applied. For example, ETH means Ethereum and BTC means Bitcoin.
```
therealhexi, line 66: Nit: indentation

Do we have any restrictions around the targetAsset? Could there be collission?

UMIPs/umip-151.md
```diff
@@ -0,0 +1,86 @@
...
+
+The following example demonstrates how this system works:
+1. Use `ancillaryData` and `timeStamp` from `proposePrice`:
+    * ancillaryData: `MFIV-14D-ETH`
+    * timeStamp: `1643288400`
+2. Use the `timestamp` and `ancillaryData` to look up the following:
+    * MFIV matches the methodology directory [here](https://github.com/Volatility-DAO/PIPS/tree/main/Approved/DAOracle). This is where you get the open-source code to validate the index.
```
therealhexi, line 73: Not seeing any source code as of now

therealhexi (commented):
Pretty thorough and well written UMIP

UMIPs/umip-151.md
```diff
@@ -0,0 +1,86 @@
+## Headers
+- UMIP 151
+- UMIP title: VolatilityDAOracle
+- Author: Volatility Group Inc.
+- Status: Draft
+- Created: February 16th 2022
+
+
+## Summary (2-5 sentences)
+The Volatility DAOracle is a collection of methodologies and implementations for indices and benchmarks. Each index can be verified by decentralized users through its data endpoint, open-source code, and methodology paper. This information can be looked up through the `timestamp` and `ancillaryData` of the `proposePrice` for each call to the Optimistic Oracle.
```
Reinis-FRP, line 10: In case of using Optimistic Oracle contract it would be better referring to the [`requestPrice` method](https://github.com/UMAprotocol/protocol/blob/0a4ce9af96481bec0c0eb2ab2ccd8da4dc87ccd5/packages/core/contracts/oracle/implementation/OptimisticOracle.sol#L155) here and throughout the rest of UMIP as instructions in this document also apply to potential proposers tasked with responding to price requests.

UMIPs/umip-151.md
```diff
@@ -0,0 +1,86 @@
...
+
+The index can be looked up and validated by participants of the DVM and/or disputers through the `timestamp` and `ancillaryData` passed into the Optimistic Oracle’s `proposePrice` function. The ancillaryData consists of the methodology, time window, and target asset concatenated into a single parameter: `id:{methodology}-{window}-{targetAsset}`. This parameter can be deconstructed and used to access data sources. If `ancillaryData` does not comply with these standards, DVM participants should return the minimum int256 value as a "magic number": `-57896044618658097711785492504343953926634992332820282019728792003956564819968` (i.e. `-2^(n-1) => -2^(256-1) => -2^255 => int256(uint256(1) << 255)`). We understand that this is a non-traditional format for `ancillaryData` but it is necessary to optimize storage within The DAOracle smart contracts.
```
Reinis-FRP, line 26:
```suggestion
The index can be looked up and validated by participants of the DVM and/or proposers/disputers through the `timestamp` and `ancillaryData` passed into the Optimistic Oracle’s `proposePrice` function. The ancillaryData consists of the methodology, time window, and target asset concatenated into a single parameter: `id:{methodology}-{window}-{targetAsset}`. This parameter can be deconstructed and used to access data sources. If `ancillaryData` does not comply with these standards, DVM participants should return the minimum int256 value as a "magic number": `-57896044618658097711785492504343953926634992332820282019728792003956564819968` (i.e. `-2^(n-1) => -2^(256-1) => -2^255 => int256(uint256(1) << 255)`). We understand that this is a non-traditional format for `ancillaryData` but it is necessary to optimize storage within The DAOracle smart contracts.
```

Reinis-FRP, line 26: If possible, it might be better to use different magic number since -2^255 is currently used in other price identifiers to respond to invalid early expiration requests.

UMIPs/umip-151.md
```diff
@@ -0,0 +1,86 @@
...
+
+1. Most indices will post to IPFS and their data snapshot can be queried through two different URLs:
+* One based on methodology, window, asset, and timestamp.
+* One based on the IPFS hash.
+2. In cases where the IPFS data is completely lost or broken, the PIP directory of the methodology will outline alternate ways of accessing the data in the README.md (e.g. it may list the exchanges or other third-party data provider from which the data originated and any special processing needed for aggregating the feed(s) for input to the open-source implementation).
+
+As part of the PIP process The Volatility DAO sets a parameter called `dataPeriod`. This represents the time period at which snapshots of input data & index values are saved to IPFS. Only the most recent snapshot can be relayed to The Volatility Oracle for audit and it can only be audited once, subsequent submissions of the same snapshot to The Volatility Oracle are denied. DVM participants should also deny repeat submissions of the same snapshot. `dataPeriod` can be set to any time resolution (e.g. 1 second or 1 hour) but should never be greater than the desired amount of time between audits.
```
Reinis-FRP, line 39: Please be more explicit on how "DVM participants should also deny repeat submissions of the same snapshot." - does it mean not responding to repeated price requests with the same identifier, timestamp, ancillaryData combination?

Note that Optimistic Oracle already enforces unique combinations of msg.sender, identifier, timestamp, ancillaryData by [requiring](https://github.com/UMAprotocol/protocol/blob/0a4ce9af96481bec0c0eb2ab2ccd8da4dc87ccd5/packages/core/contracts/oracle/implementation/OptimisticOracle.sol#L162) that such combination has never been requested.

UMIPs/umip-151.md
```diff
@@ -0,0 +1,86 @@
...
+    * For example, 14d means 14 Day volatility.
+targetAsset - This is the asset to which volatility is applied. For example, ETH means Ethereum and BTC means Bitcoin.
+
+The following example demonstrates how this system works:
+1. Use `ancillaryData` and `timeStamp` from `proposePrice`:
+    * ancillaryData: `MFIV-14D-ETH`
+    * timeStamp: `1643288400`
+2. Use the `timestamp` and `ancillaryData` to look up the following:
+    * MFIV matches the methodology directory [here](https://github.com/Volatility-DAO/PIPS/tree/main/Approved/DAOracle). This is where you get the open-source code to validate the index.
+    * `timestamp / window / targetAsset` are all used to query the IPFS data of the calculations. Instructions for doing this are required within the README of every PIP. You can see the MFIV README [here](https://github.com/Volatility-DAO/PIPS/tree/MFIV/Proposed/Adding_An_Index/Step_2/MFIV#readme).
```
Reinis-FRP, line 74: Please add instructions for rounding timestamp intervals

UMIPs/umip-151.md
wadek2: Not entirely sure how we should document this as we have combined some logic. We are using the requestPrice method but we combine logic. 

Specifically, the DAOracle only sends to UMA when there is a dispute. When a user calls the dispute function our smart contracts combine the logic of requestPrice and proposePrice while taking advantage of gas savings from not having to overwrite Request params that a normal requestPrice() => proposePrice() flow would entail. The Optimistic Oracle object is built with the following function:
```
  function requestAndProposePriceFor(
    bytes32 identifier,
    uint32 timestamp,
    bytes memory ancillaryData,
    IERC20 currency,
    uint256 reward,
    uint256 bond,
    uint256 customLiveness,
    address proposer,
    int256 proposedPrice
  ) external virtual returns (uint256 totalBond);
  ```
Then in a single transaction this object is disputed. Technically someone could request a price without the DAOracle and the params (indentifier, timestamp, & ancillaryData) would be the same.

So you want us to use requestPrice or should we specify the above function?

Comment (Reinis-FRP) [reply UMIPs/umip-151.md:10]: Regular [Optimistic Oracle contract](https://etherscan.io/address/0xc43767f4592df265b4a9f1a398b97ff24f38c6a6) does not have `requestAndProposePriceFor` method. Or are you referring to this method in your DAOracle contract? If so, do you expect it to resolve directly to DVM (in which case DAOracle would need to be registered to allow making DVM requests) or through regular Optimistic Oracle contract?

File: UMIPs/umip-151.md [hunk 7]
```diff
@@ -0,0 +1,111 @@
...
+
+## Technical Specification
+Each index has a data endpoint, open-source code, and methodology paper which are approved by the Volatility DAO. The Volatility DAO approves an index/benchmark through the Protocol Improvement Process which can be found on the [Volatility DAO Github](https://github.com/Volatility-DAO/PIPS). The approved PIP directory contains all necessary steps, software, and endpoints to validate an index.
+
+The index can be looked up and validated by participants of the DVM and/or disputers through the `timestamp` and `ancillaryData` passed into the Optimistic Oracle's `proposePrice` function. The ancillaryData consists of the methodology, time window, and target asset concatenated into a single parameter: `id:{methodology}-{window}-{targetAsset}`. This parameter can be deconstructed and used to access data sources. If `ancillaryData` does not comply with these standards, DVM participants should return the "magic number": `-3735928559` which equals . We understand that this is a non-traditional format for `ancillaryData` but it is necessary to optimize storage within The DAOracle smart contracts.
```
Reinis-FRP, line 26: > DVM participants should return the "magic number": `-3735928559` which equals .

This sentence is not finished

UMIPs/umip-151.md
wadek2: These are our own contracts. Specifically we do the following:

When there is a `Dispute` in our system prior to the expiration of the time window, a bond is posted on behalf of DAOracle stakers and an equal amount is pulled from the caller (i.e. the Disputer). This actually requests, proposes, and disputes with UMA's SkinnyOO which sends the bonds and the disputed proposal to UMA's DVM for settlement.

Reinis-FRP: dispute with UMA's SkinnyOO makes `requestPrice` call to DVM, so from UMA voters' perspective it would be better to reference this call.

UMIPs/umip-151.md
```diff
@@ -0,0 +1,112 @@
...
+    * `D` = Days
+    * `W` = Weeks
+    * `M` = Months
+    * `Y` = Year
+    * For example, 14d means 14 Day volatility.
+targetAsset - This is the asset to which volatility is applied. For example, ETH means Ethereum and BTC means Bitcoin.
+
+The following example demonstrates how this system works:
+1. Use `ancillaryData` and `timeStamp` from `proposePrice`:
+    * ancillaryData: `MFIV-14D-ETH`
```
Reinis-FRP, line 72: If following syntax definition above, shouldn't this be `id:MFIV-14D-ETH` instead?

UMIPs/umip-151.md
wadek2: Users will be able to look up the targetAsset within the methodology -> Implementation_Parameters.MD. Within these will be specific addresses for the potential of naming collisions. I will add instructions in the UMIP.

wadek2: Just added instructions for how we whitelist to remove potential for collission.

UMIPs/umip-151.md
wadek2: I changed it to `requestAndProposePriceFor` as this is the function we are calling in the SkinnyOO. I also changed references of OO to SkinnyOO.

UMIPs/umip-151.md
wadek2: I believe we addressed this. Please let us know if it needs more clarity.

wadek2: We just found an edge case where this rounding solution is problematic. We are going to address this by adjusting the format of the ancillaryData. This will take us a day or two as we need to make sure that we are not running into memory issues with Solidity.

wadek2: This is all set now.

PR #489, opened 2022-02-24
Add Boba WAGMIv1 KPI options implementation

Reinis-FRP:


smb2796 (approved):
From what I can see, the two major changes are the post-processing and making TVL be eth denominated. Both look good to me. Great work!

PR #492, opened 2022-03-08
Add ZODIAC as price identifier

pemulis:

This is the first draft of a UMIP to add a `ZODIAC` identifier that would allow the Optimistic Oracle to enforce a set of natural language off-chain rules governing a Gnosis Safe, allowing flexible and powerful "optimistic governance" for the first time.

### DIFF
--- a/Implementations/zodiac.md
+++ b/Implementations/zodiac.md
@@ -166,6 +166,66 @@ This methodology allows for a huge amount of flexibility for DAOs to manage thei
 ## Implementation
 The implementation of the Zodiac module is currently in a draft state [viewable here](https://github.com/UMAprotocol/protocol/pull/3843/files).
 
+TODO: Script that can decode transactions for voters.
+
+### Example Rules
+The `ZODIAC` identifier is designed to allow rules to be as flexible as possible while still being clear to UMA voters called in to resolve disputes. These examples are meant to inspire creativity in users creating their own rules and demonstrate the legalistic approach they should take.
+
+The actual rules for a given module will depend entirely on the intended use of the Gnosis Safe governed by that module and the community dynamics of the DAO in charge.
+
+As a reminder, these rules should be published publicly and referenced in a string stored in the module contract, as an IPFS hash, a website URI, or some other format.
+
+```
+EXAMPLE RULES
+
+ABC DAO was formed to bring our token holders to the moon, Mars, and beyond.
+
+To that end, we have established the following rules governing our treasury, contained in the Gnosis Safe at Ethereum address 0xbcd...234. This Gnosis Safe also has the ability to change certain parameters in our other smart contracts.
+
+1. Transactions proposed by 0xabc...123 are valid unless they violate other rules. This address is a multi-signature wallet controlled by the core development team.
+
+2. No more than 10% of the value contained in the Gnosis Safe at address 0xbcd...234, which is governed by the Zodiac module 0xcde...345, may be transferred out of the Gnosis Safe during a single 24-hour period.
+
+3. Proposals including transactions that call functions on the smart contract 0xdef...456 must be approved by a Snapshot vote backed by more than 50% of the total supply of $ABC ERC-20 tokens and must be proposed by 0xabc...123.
+
+4. Proposals that have been approved by a Snapshot vote backed by more than 50% of the total supply of $ABC ERC-20 tokens are valid.
+
+5. Transactions proposed by an address owned by Elon Musk are valid unless they violate other rules. Elon Musk can prove ownership of an address through a post on the @elonmusk Twitter account combined with a public press release.
+
+6. These rules may be changed at any time by a transaction approved by a Snapshot vote backed by more than 50% of the token supply of $ABC ERC-20 tokens, or by Elon Musk.
+
+7. Elon Musk may not propose transactions when it is the second Tuesday of the month, Pacific Time Zone.
+```
+
+### Successful Execution Flow
+1. DAO member Alice has a great idea for utilizing the DAO treasury and writes some Ethereum transactions that would execute her idea.
+2. Alice holds a Snapshot vote to approve their idea and associated transactions.
+3. The majority of $ABC tokens back Alice's proposal in the Snapshot vote.
+4. Alice submits the proposal to the Zodiac module governing the Gnosis Safe and notes in the ancillary data that the proposal was approved on Snapshot and includes a link to the Snapshot results. Alice includes a bond with her proposal.
+5. The proposal is not disputed within the challenge window and can be executed by any address.
+6. Bob, another member of the DAO, executes Alice's proposal since Alice is out watching a movie.
+7. Alice's transactions are executed and the treasury funds are spent according to her plan, which was approved by a Snapshot vote.
+
+### Disputed Execution Flow
+Starting after step 4 from the Successful Execution Flow:
+
+1. Bob, who is jealous of Alice's popularity in the DAO, disputes her proposal even though it has been approved by a Snapshot vote. Bob includes a bond with his dispute and rages against the proposal on Twitter.
+2. The dispute is escalated to UMA's data verification mechanism where UMA tokenholders analyze the proposal, the arguments from Alice and Bob, and the rules published by the DAO and referenced in the Zodiac module.
+3. After reading the rules, the vast majority of UMA token holders conclude that the proposal followed the rules and should be executed, and commit a value of `1` during the commit period.
+4. UMA token holders reveal their votes during the reveal period.
+5. After the reveal period, the settlement value is `1`. The Zodiac module receives the settlement value and allows any user to execute the proposal.
+6. Bob loses his bond and Alice gets her bond back and a portion of Bob's bond.
+7. Alice, vindicated, executes her proposal.
+
+### Blocked Execution Flow
+1. Elon Musk proposes to spend some DAO treasury funds, and includes a bond with his proposal, but it is the second Tuesday of the month, Pacific Time Zone.
+2. Alice disputes Elon's proposal and includes a bond with her dispute.
+3. The dispute is escalated to UMA's data verification mechanism where UMA tokenholders analyze the proposal and the rules published by the DAO and referenced in the Zodiac module.
+4. After reading the rules, the vast majority of UMA token holders conclude that the proposal did NOT follow the rules and should not be executed, and commit a value of `0` during the commit period.
+5. UMA token holders reveal their votes during the reveal period.
+6. After the reveal period, the settlement value is `0`. The Zodiac module receives the settlement value and deletes the invalid proposal.
+7. Elon loses his bond and Alice receives her bond back and a portion of Elon's bond.
+
 ## Security considerations
 
 ### Ambiguity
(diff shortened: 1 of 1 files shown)


Implementations/zodiac.md
```diff
@@ -0,0 +1,200 @@
+## Headers
+- UMIP <#> 
+- UMIP title: Add ZODIAC as a supported price identifier
+- Author: John Shutt (john@umaproject.org)
+- Status: DRAFT
+- Created: March 7, 2022
+- Discourse Link: TBD
+
+## Summary (2-5 sentences)
+The ZODIAC identifier is intended to be used with a [Zodiac module](https://gnosis.github.io/zodiac/docs/intro) that allows you to control a [Gnosis Safe](https://gnosis-safe.io/) according to a set of rules defined off-chain and enforced with UMA's [Optimistic Oracle](https://umaproject.org/optimistic-oracle.html). Any address can propose transactions that follow the rules and any address can dispute a proposal to UMA's Optimistic Oracle within a challenge window.
```
smb2796, line 10: > Any address can propose transactions that follow the rules and any address can dispute a proposal to UMA's Optimistic Oracle within a challenge window.

Does this not depend on the configuration of the zodiac module? Couldn't proposals be restricted to only an approved multisig for example, or is that also intended to be controlled at the OO level?

Implementations/zodiac.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Motivation
+The ZODIAC identifier, coupled with the Optimistic Oracle Zodiac module, will allow for a new era of flexible "optimistic governance," where management of DAO treasuries and other multi-signature wallets can be managed more effectively, without being limited to X-of-Y signing schemes or tokenholder votes, although the Zodiac module can enforce those things, too.
```
smb2796, line 13: Kind of answers my question above.

Implementations/zodiac.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+The module contract stores an array of `Proposal` objects and a string reference to an off-chain set of rules that have been publicly published, which may be an IPFS hash, a URI, or something else.
+
+```
+Proposal[] public proposals;
+string public rules;
+```
+
+When a user creates a proposal, they submit an array of transactions along with optional ancillary data that explains the intent and purpose of the transactions, which is useful for voters trying to understand the transactions and whether or not they follow the publisehd rules.
```
smb2796, line 46:
```suggestion
When a user creates a proposal, they submit an array of transactions along with optional ancillary data that explains the intent and purpose of the transactions, which is useful for voters trying to understand the transactions and whether or not they follow the published rules.
```

Implementations/zodiac.md
```diff
@@ -0,0 +1,200 @@
...
+1. The published rules for the Zodiac module.
+2. The rationale given by the proposer in ancillary data.
+3. The arguments presented by disputers.
+4. The voter's understanding of what the transactions *actually* do, regardless of the stated rationale.
+
+If a proposal follows the rules, an UMA voter should return a value of `1`.
+
+If a proposal doesn't follow the rules, an UMA voter should return a value of `0`.
+
+If a voter is unsure if a proposal follows the rules, an UMA voter should return a value of `0`.
```
smb2796, line 105: Would also add that if ancillary data/the ruleset is malformed, voters should return a value of 0. I understand that this probably covers it, but just to be extra explicit.

Implementations/zodiac.md
```diff
@@ -0,0 +1,200 @@
...
+
+It is the responsibility of the DAO users of the `ZODIAC` identifier to write clear and unambiguous rules. If their rules are unclear, they should expect proposals that get disputed to be rejected by UMA's oracle.
```
smb2796, line 107: What about situations where the proposal is for updating an unclear ruleset. If there are no rules initially for example, how could voters determine if a ruleset update is valid? Maybe an unnecessary edge case to address, but seems like this could potentially result in stuck governance where nothing could be passed.

Implementations/zodiac.md
```diff
@@ -0,0 +1,200 @@
...
+It's easiest to understand this by example.
+
+Let us consider a Magicland DAO with two rules in their published rules document:
+
+```
+1. The treasurer is allowed to move money from the treasury at their discretion for the good of the DAO.
+2. The rules can be changed by a Snapshot vote where more than 50% of the token supply votes to change the rules.
+```
+
+As it turns out, the Magicland treasurer is revealed to be a serial scam artist and proposes to move 10,000 ETH from the Magicland treasury to Tornado Cash through their own wallet address.
```
smb2796, line 126: Seems like this would be caught at the snapshot level no?

Implementations/zodiac.md
```diff
@@ -0,0 +1,200 @@
...
+    // Canonical value representing "True"; i.e. the transactions are valid.
+    int256(1e18)
+);
+```
+
+The requester in the `RequestPrice` event will be the Zodiac module itself, where voters can find the `rules` identified in the Zodiac module.
+
+[NOTE: Should the rules also be prepended to the ancillary data? Maybe. But consider the Important Special Case and the fact that we want voters to consider only the most up-to-date set of rules.]
+
+### Resolving Disputes
```
smb2796, line 90: Do you think there needs to be a set ancillaryData format for "Supporting Information"?

smb2796:
Very cool proposal. I think a few things would be really helpful:
- A full example of a standard governance ruleset
- An example of a governance vote being carried out from proposal to execution. 
- Link to a script that can perform tx decoding for voters

Implementations/zodiac.md
pemulis: You're right that a module contract using the identifier could add restrictions like that so I'll modify the language to not assume *any* address can propose transactions. Technical restrictions in the contract shouldn't be a concern for the OO but any restrictions outlined in the rules would be.

Implementations/zodiac.md
pemulis: Good idea; will add.

Implementations/zodiac.md
pemulis: I'll add something about this to Security Considerations. Realistically, I think social consensus would save the day and the DAO could communicate to UMA voters what rules they should reference even if they're not linked in the module due to mistakes on their end. We don't want that to happen, though.

Implementations/zodiac.md
pemulis: The rules say that the treasurer can move funds at their own discretion without a Snapshot vote, but their proposal can still be disputed with the OO.

Implementations/zodiac.md
pemulis: I think it can just be a brief description with length limits, with some prepended data. I'll add more detail.

Implementations/zodiac.md
```diff
@@ -0,0 +1,278 @@
...
+When a user creates a proposal, they submit an array of transactions along with optional ancillary data that explains the intent and purpose of the transactions, which is useful for voters trying to understand the transactions and whether or not they follow the published rules.
+
+```
+function proposeTransactions(Transaction[] memory transactions, bytes memory ancillaryData) public {
+    ...
+}
+```
+
+Each proposal has a unique ID, which is assigned in the transaction where a proposal is created. The ID is incremented by one with each new proposal.
+
```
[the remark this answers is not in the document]
pemulis: Not necessarily! A proposal is just a collection of transactions that is being proposed, like with the UMA Governor contract.

If you want to use this with Snapshot, you would specify in the `rules` that proposals that were approved through Snapshot are valid, so that disputers and UMA voters would know to check Snapshot to see if a proposal is valid.

But there are lots of other possibilities besides Snapshot, with some examples listed further down in the document, including rules that don't involve a voting system at all.

But to answer your other question, if there is any transaction within a proposal that is invalid, the whole proposal is invalid and should be disputed and rejected.

Implementations/zodiac.md
```diff
@@ -0,0 +1,278 @@
...
+`id:123,rules:QmTp2hEo8eXRp6wg7jXv1BLCMh5a4F3B7buAUZNZUu772j,explanation:These transactions were approved by the majority of tokenholders in a Snapshot vote on March 8, 2022.`
+
+### Resolving Disputes
+
+If a proposal is disputed, the disputer is encouraged to publish their reasoning in a public forum to assist UMA voters in their determination of whether or not the proposal was valid.
+
+Voters should consider four sources of public information in their ruling:
+
+1. The published rules for the Zodiac module.
```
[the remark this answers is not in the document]
pemulis: Human readable text version of the comprehensive rules governing the target (i.e. Gnosis Safe). You could think of this as the book of laws for a DAO.

It doesn't necessarily involve Snapshot, but if it did, you could ignore the Snapshot strategy in the published `rules` and focus on the output (was a proposal approved by Snapshot or not), treating the strategy as a black box. Or you could specify the strategy in both places if you want, or set some special rules around when and how to change the strategy, and so on.

But the big innovation here is that you aren't restricted by what is technically possible within Snapshot or a multi-sig wallet or anything and can just treat those things as one of many possible sources to consider within a set of natural language rules interpreted by proposers, disputers, and voters, and those natural language rules govern which proposals are valid.

Implementations/zodiac.md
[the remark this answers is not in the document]
mhairimcalpine: Ooh, my, that is clever!

I assumed the examples below were post snapshot vote, I didnt realise you could catch it without one!

Thanks for clarifying that it is per entire proposal rather than per transaction, or in lumps.

Implementations/zodiac.md
```diff
@@ -0,0 +1,278 @@
...
+4. Proposals that have been approved by a Snapshot vote backed by more than 50% of the total supply of $ABC ERC-20 tokens are valid.
+
+5. Transactions proposed by an address owned by Elon Musk are valid unless they violate other rules. Elon Musk can prove ownership of an address through a post on the @elonmusk Twitter account combined with a public press release.
+
+6. These rules may be changed at any time by a transaction approved by a Snapshot vote backed by more than 50% of the token supply of $ABC ERC-20 tokens, or by Elon Musk.
+
+7. Elon Musk may not propose transactions when it is the second Tuesday of the month, Pacific Time Zone.
+```
+
+### Successful Execution Flow
```
smb2796, line 200: Awesome stuff, thank you for the examples!

Implementations/zodiac.md
```diff
@@ -0,0 +1,278 @@
...
+
+The requester in the `RequestPrice` event will be the Zodiac module itself, where voters can find the `rules` identified in the Zodiac module.
+
+### Ancillary Data Format
+
+The `ancillaryData` submitted with proposals should consist of:
+
+- `id`, which is automatically prepended to the ancillary data and used to find the proposal details in the module
+- `rules`, which is a reference to the rules for the module, for example, an IPFS hash or a website URI
+- `explanation`, which is written in natural language by the proposer and explains the nature of the transactions and how they follow the module's rules
```
smb2796, line 92: `explanation` is optional correct? I'm also assuming that the `rules` content is automatically prepended as well, and only `explanation` can be edited by the proposer.

smb2796 (approved):

PR #494, opened 2022-03-10
Jarvis KPI Option Update

abg4:

Jarvis KPI Option update to provide a minimum KPI Option payout. The update is to change the payout to a minimum value of 1 JRT if liquidity is between 0 and $1M instead of 0 JRT.


Implementations/jarvis-lp.md
```diff
@@ -46,15 +46,14 @@ Rounding:-6
...
 9. Calculate the average from the daily values in Step 8 to obtain the LP TVL in terms of `TVLCurrency`. 
-10. Apply floor rounding on the output from step 9 using instructions from [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md) and `Rounding` parameter passed in ancillary data. As an example, if `Rounding` parameter was -6 and the output from step 9 was USD 800,000, the value should be rounded down to 0. Alternatively, if the output from step 9 was USD 1,200,000, the value should be rounded down to 1,000,000.
+10. Apply floor rounding on the output from step 9 using instructions from [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md) and `Rounding` parameter passed in ancillary data. As an example, if `Rounding` parameter was -6 and the output from step 9 was USD 3,800,000, the value should be rounded down to 3,000,000. Alternatively, if the output from step 9 was USD 2,200,000, the value should be rounded down to 2,000,000.
 
 ## Intended Application
```
smb2796, line 51: Can you change this to "Post-Processing"?

Implementations/jarvis-lp.md
```diff
@@ -46,15 +46,14 @@ Rounding:-6
...
 
-As an illustration, based on the intended ancillary data above, `upperBound` would be set to 4,000,000 implying the following payouts:
-* LP TVL below or equal to USD 1,000,000 would have returned 0 and long KPI option holders would receive 0 payout;
-* LP TVL above USD 1,000,000 and below or equal to USD 2,000,000 would have returned 1,000,000 and long KPI option holders would receive 1,000,000/4,000,000=25% of `collateralPerPair` for each token;
+With the intention of providing a minimum payout, if the LP TVL is below or equal to USD 2,000,000, voters should return 1,000,000. For values greater than USD 2,000,000, floor rounding should be used to calculate the returned value. As an illustration, based on the intended ancillary data above, `upperBound` would be set to 4,000,000 implying the following payouts:
```
smb2796, line 55:
```suggestion
With the intention of providing a minimum payout, if the LP TVL is below or equal to USD 2,000,000 after the necessary rounding is applied, voters should return 1,000,000. For values greater than USD 2,000,000, floor rounding should be used to calculate the returned value. As an illustration, based on the intended ancillary data above, `upperBound` would be set to 4,000,000 implying the following payouts:
```

Implementations/jarvis-lp.md
abg4: Updated

smb2796 (approved):

PR #506, opened 2022-03-30
Across v2 UMIP

mrice32:


UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Proposal Information
+
+From the selected event, one should be able to glean the following information:
+- `bundleEvaluationBlockNumbers`
+- `poolRebalanceRoot`
+- `relayerRefundRoot`
+- `slowRelayRoot`
+
+The `bundleEvaluationBlockNumbers` is an ordered array of block numbers for each destination chain. Which index
```
chrismaree, line 68: should this not just be ordered by chainId of the whitelisted chainIds at the given block number?(i.e mainnet is always first as chainId 1, then optimism as chainId 10 eat). This way we dont need to store a mapping between chainId and index within this array.

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+any `FilledRelay` events that have `isSlowRelay` set to true.
+
+For all FilledRelay events, you can find the SpokePool for the origin chain by querying
+[CrossChainContractsSet](https://github.com/across-protocol/contracts-v2/blob/aac42df9192145b5f4dc17162ef229c66f401ebe/contracts/HubPool.sol#L120)
+and finding all matching events where the l2ChainId matches the `originChainId` value in the FilledRelay event. The
+`spokePool` values in these events are all possible spoke pools that this deposit could have been from. Note: old
+SpokePools must be considered so upgrades don't block old deposits from being relayed.
+
+Note: in the sections below, if the relay is considered to be invalid at any point, that relay will not be considered
+when constructing the bundle.
```
chrismaree, line 104:
```suggestion
Note: in the sections below, if the relay is considered to be invalid at any point, that relay must not be considered
when constructing the bundle.
```

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+
+For each FilledRelay event found earlier, a
```
chrismaree, line 106: should we notate this like this?
```suggestion
For each `FilledRelay` event found earlier, a
```

same goes for references to `SpokePool` in other places and `l2ChainId` and others.

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+[RootBundleExecuted](https://github.com/across-protocol/contracts-v2/blob/3676377398714bd8f734485deead75afbcd8e8c4/contracts/HubPool.sol#L158-L167)
+should be queried to find a matching element for the `chainId` and `l1Token` (they would need to be at matching indices
+in the array). Once these values are found at matching indices, that index can be used to find the associated running
+balance value. That most recent running balance value should be added to the one computed above.
+
+For each tuple of `l1Token` and `chainId`, we should have computed a total running balance value. Determine if the absolute
+value of this running balance is > 1% of the total value of all LP tokens for this `l1Token` at the
+`ProposeRootBundle` event that came before the one being evaluated (process described above). Note: this UMIP should be
+amended in the future to change this 1% to a configured value. If this passes the threshold, then set net send amount
+to the running balance value and set the running balance value to 0.
```
chrismaree, line 188: would it make sence to define this magic number as a const at the top of the UMIP so it can be easily change? alternatively, you could reference this as a "setting" within the config contract.

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+
+Find all `FilledRelay` events in the block range that have `isSlowRelay` set to true. For each, use a similar method
+as above to map this event back to an `l1Token` at the `quoteTimestamp`. Use the `destinationChainId` as the `chainId`
+to determine which running balance that this event should apply to. If there is no running balance initialized for this
+`l1Token` and `chainId`, then do a historical lookup as above to determine the previous running balance value and add
+this value to it. If there is no historical running balance, initialize it to 0. Add `fillAmount` to this running
+balance value.
+
+Take the above running balances and net send amounts and group them by only `chainId` and sort by `chainId`. Within
+each group, sort by `l1Token`. If there are more than 25 `l1Tokens`, a particular chain's leaf will need to be broken
```
chrismaree, line 198: again, magic number should be put in config store, no?

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+
+Now that we have ordered leaves, we can assign each one a unique `leafId` starting from 0.
+
+With all of that information, each leaf should be possible to construct in the format given
+[here](https://github.com/across-protocol/contracts-v2/blob/3676377398714bd8f734485deead75afbcd8e8c4/contracts/HubPoolInterface.sol#L13-L42).
+
+Once the leaves are constructed, the merkle root can be constructed by hashing each leaf data structure using
+Solidity's standard process of `keccak256(abi.encode(poolRebalanceLeaf))`. Once the leaves are hashed, the tree should
+be constructed in the standard way such that it is verifyable using
+[OpenZeppelin's MerkleProof](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/MerkleProof.sol)
```
chrismaree, line 209: I replaced this with a permalink

```suggestion
[OpenZeppelin's MerkleProof](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/742e85be7c08dff21410ba4aa9c60f6a033befb8/contracts/utils/cryptography/MerkleProof.sol)
```

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+
+The `l2TokenAddress` is the corresponding L2 token address for the `l1Token` in the previous section. Note: see above
+sections for how to map L1 and L2 tokens via events on L1. This mapping should be done according to the highest
+`quoteTimestamp` of any relays in the group. If no relays are present, then as of the previous successful proposal.
+
+`refundAmounts` and `refundAddresses` are just computed by grouping the relays in this group by the `relayer` and
+summing the `amount - (amount * lpFeePct / 1e18)` for each relay. These should be sorted in descending order of
+`refundAmounts`. If two `refundAmounts` are equal, then they should be sorted by ascending `relayer` address.
+
+If there are more than 25 `refundAddresses` for a particular `l2TokenAddress` then these should be split up into 25
```
chrismaree, line 230: again magic number.

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+element leaves (sorted as described above) with only the first leaf for a particular `l2TokenAddress` able to contain a
+nonzero amountToReturn.
+
+Once these are computed for all relays, the leaves (or groups of leaves for > 25 elements) should be sorted by
+`chainId` as the primary index, then `l2TokenAddress` as the secondary index, and then the individual sorting of
+> 25 element groups as the tertiary sorting. Once these are sorted, each leaf can be given a `leafId` based on its
```
chrismaree, line 236: splitting this inequality makes markdown unhappy.
```suggestion
`chainId` as the primary index, then `l2TokenAddress` as the secondary index, and then the individual sorting of > 25
element groups as the tertiary sorting. Once these are sorted, each leaf can be given a `leafId` based on its
```

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+The ooRequester address is expected to be an instance of the
+[HubPool contract](https://github.com/across-protocol/contracts-v2/blob/aac42df9192145b5f4dc17162ef229c66f401ebe/contracts/HubPool.sol).
+
+If any of the expected details in the ooRequester are not available in the expected form because the HubPool does not
+match the expected interface, the identifier should return 0.
+
+To get the proposal data, the voter should find events that matches
+[this signature](https://github.com/across-protocol/contracts-v2/blob/aac42df9192145b5f4dc17162ef229c66f401ebe/contracts/HubPool.sol#L149-L157)
+on the ooRequester. The event that describes this proposal is the matching event with the highest block number whose
+timestamp is less than or equal to the timestamp of the price request. If there are two matching events that both
```
nicholaspai, line 54: How would it be possible that two events from the same `ooRequester` could be emitted on the same block? 

Would you have to `proposeRootBundle`, `disputeRootBundle`, and then `proposeRootBundle` atomically? In this case wouldn't you want the highest transaction index?

UMIPs/UMIP-across-v2.md
nicholaspai: How can you query whitelisted chain IDs at a given block number? Would you need to fetch the entire history of `SetPoolRebalanceRoute` events for this? This seems cumbersome

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+The `bundleEvaluationBlockNumbers` is an ordered array of block numbers for each destination chain. Which index
+corresponds to which chain is denoted by the config stored at TBD. From this, one should be able to determine a list
+of chainIds and ending block numbers.
+
+To determine the start block number for each chainId, search for the latest
+[RootBundleExecuted event](https://github.com/across-protocol/contracts-v2/blob/aac42df9192145b5f4dc17162ef229c66f401ebe/contracts/HubPool.sol#L158-L167)
+with a matching chainId while still being earlier than the timestamp of the request. Once that event is found, search
+for the
+[ProposeRootBundle](https://github.com/across-protocol/contracts-v2/blob/aac42df9192145b5f4dc17162ef229c66f401ebe/contracts/HubPool.sol#L149-L157)
+event that is as late as possible, but earlier than the RootBundleExecuted event. Once this proposal event is found, determine its
```
nicholaspai, line 77:
```suggestion
event that is as late as possible, but earlier than the RootBundleExecuted event we just identified. Once this proposal event is found, determine its
```

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Finding Slow Relays
+
+To determine all slow relays, follow the following process:
+
+1. For all valid `FilledRelay` events above, group them by `originChainId` and `depositId`.
+2. Remove all groups that contain a `FilledRelay` event where `totalFilledAmount` equals `amount`.
+3. Remove all groups that do not contain an event where `filledAmount` is nonzero and equal to `totalFilledAmount`.
+
+For all remaining groups, they should be stored in a list of slow relay groups.
```
nicholaspai, line 152: Should we specify exactly how to order the slow relay groups? As a strawman, the method I've been using in the off-chain client is:
- Assuming you've grouped all `FilledRelay` events (i.e. what you're left with after step 3 above) by `destinationChainId`.
- Order the dictionary of slow relays (`{ destinationChainId => FilledRelay[] }`) by `destinationChainId` with the highest chain ID's first.
- For each `destinationChainId` key, order the `FilledRelay` by the unfilled amount remaining, which is equal to `FilledRelay.amount - FilledRelay.totalFilledAmount`, from highest to lowest
- Flatten the dictionary of `FilledRelays`

mrice32: In constructing the process of constructing the tree below, I suggested the following ordering process:
```
Their primary sorting index should be `originChainId` and the secondary sorting index should be `depositId`.
```

Since the ordering doesn't matter for efficiency reasons, I figured we should just pick an ordering based on the values that are guaranteed to be unique.

Is there a strong reason for sorting by unfilled amount? Do you think that would have a benefit?

nicholaspai: Ah I see, nope I agree with what you've written then!

nicholaspai: I'll also change the dataworker implementation to reflect this ordering key

nicholaspai: Do you need to specify whether you're ordering ID's ascending or descending? I assume ascending but maybe worth clarifying

UMIPs/UMIP-across-v2.md
mrice32: Yep, exactly.

UMIPs/UMIP-across-v2.md
mrice32: Yeah, exactly, @nicholaspai. I left this TBD because I wasn't sure what the right mechanism was to get an updated list of all whitelisted chain ids was.

@chrismaree I agree that once the list is available, we could sort by the numerical value.

UMIPs/UMIP-across-v2.md
mrice32: Agreed, we should specify that.

mrice32: I addded a new note at the top to assume that all sorting is ascending unless specified.

UMIPs/UMIP-across-v2.md
mrice32: Done.

UMIPs/UMIP-across-v2.md
mrice32: Done.

UMIPs/UMIP-across-v2.md
mrice32: Done.

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,276 @@
...
+
+## Constructing the PoolRebalanceRoot
+
+To construct the `poolRebalanceRoot`, you need to form a list of rebalances.
+
+For all valid `FilledRelay` events above, group them by `repaymentChainId` and their associated `l1Token` found above.
```
nicholaspai, line 168: We should limit the number of `l1Tokens` to ~50 which should be plenty of capacity in practice and avoids out of gas execution errors on L1

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,276 @@
...
+sort should be on the transactionIndex, and the tertiary sort should be on the logIndex.
+
+Note 3: wherever unspecified, sorting should be ascending by default, not descending.
+
+## Constants
+
+The following constants are currently specified in this UMIP directly, but should be moved to an on-chain
+configuration in the future. Once that happens, this UMIP can be amended to specify where to pull them from.
+
+MAX_LEAF_ARRAY_SIZE = 25
```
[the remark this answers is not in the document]
nicholaspai: Maybe 200

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,276 @@
...
+
+Find all `FilledRelay` events in the block range that have `isSlowRelay` set to true. For each, use a similar method
+as above to map this event back to an `l1Token` at the `quoteTimestamp`. Use the `destinationChainId` as the `chainId`
+to determine which running balance that this event should apply to. If there is no running balance initialized for this
+`l1Token` and `chainId`, then do a historical lookup as above to determine the previous running balance value and add
+this value to it. If there is no historical running balance, initialize it to 0. Add `fillAmount` to this running
+balance value.
+
+Take the above running balances and net send amounts and group them by only `chainId` and sort by `chainId`. Within
+each group, sort by `l1Token`. If there are more than `MAX_LEAF_ARRAY_SIZE` `l1Tokens`, a particular chain's leaf will
```
nicholaspai, line 208: I don't think there needs to be a limit on # of slow relay leaves as the dataworker can always just execute them 1 at a time

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,276 @@
...
+summing the `amount - (amount * lpFeePct / 1e18)` for each relay. These should be sorted in descending order of
+`refundAmounts`. If two `refundAmounts` are equal, then they should be sorted by `relayer` address.
+
+If there are more than `MAX_LEAF_ARRAY_SIZE` `refundAddresses` for a particular `l2TokenAddress` then these should be
+split up into `MAX_LEAF_ARRAY_SIZE` element leaves (sorted as described above) with only the first leaf for a
+particular `l2TokenAddress` able to contain a nonzero amountToReturn.
+
+Once these are computed for all relays, the leaves (or groups of leaves for > 25 elements) should be sorted by
+`chainId` as the primary index, then `l2TokenAddress` as the secondary index, and then the individual sorting
+of > `MAX_LEAF_ARRAY_SIZE` element groups as the tertiary sorting. Once these are sorted, each leaf can be given a
```
nicholaspai, line 247: arguably, we don't need a max leaf count for relayer refund leaves either since the data worker can choose the amount to execute in a multicall, right?

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,276 @@
...
+`refundAmounts` and `refundAddresses` are just computed by grouping the relays in this group by the `relayer` and
+summing the `amount - (amount * lpFeePct / 1e18)` for each relay. These should be sorted in descending order of
+`refundAmounts`. If two `refundAmounts` are equal, then they should be sorted by `relayer` address.
+
+If there are more than `MAX_LEAF_ARRAY_SIZE` `refundAddresses` for a particular `l2TokenAddress` then these should be
+split up into `MAX_LEAF_ARRAY_SIZE` element leaves (sorted as described above) with only the first leaf for a
+particular `l2TokenAddress` able to contain a nonzero amountToReturn.
+
+Once these are computed for all relays, the leaves (or groups of leaves for > 25 elements) should be sorted by
+`chainId` as the primary index, then `l2TokenAddress` as the secondary index, and then the individual sorting
```
nicholaspai, line 246: This could more specifically be `repaymentChainId` right? I also could see it being one of the other chain ID's

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,276 @@
...
+
+The `amountToReturn` should be set to `max(-netSendAmount, 0)`.
+
+The `l2TokenAddress` is the corresponding L2 token address for the `l1Token` in the previous section. Note: see above
+sections for how to map L1 and L2 tokens via events on L1. This mapping should be done according to the highest
+`quoteTimestamp` of any relays in the group. If no relays are present, then as of the previous successful proposal.
+
+`refundAmounts` and `refundAddresses` are just computed by grouping the relays in this group by the `relayer` and
+summing the `amount - (amount * lpFeePct / 1e18)` for each relay. These should be sorted in descending order of
+`refundAmounts`. If two `refundAmounts` are equal, then they should be sorted by `relayer` address.
```
nicholaspai, line 239: We also need to group the refunds by `repaymentChainId` otherwise some refunds for another chain won't be executable

Reinis-FRP (commented):
Since this is quite involved for manual verification, this UMIP should also add link to price feed implementation once it is ready.

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,276 @@
...
+
+Three conditions must be met for the proposal to be deemed valid:
+1. The roots computed above match the ones in the proposal.
+2. The poolRebalanceLeafCount specified in the proposal event matches the number of pool rebalance leaves computed in
+   the PoolRebalanceRoot above.
+3. `bundleEvaluationBlockNumbers` must be reasonably recent (up-to-date within 30 minutes of the proposal transaction)
+   and include all `chainIds` that have a nonzero
+   [CrossChainContractsSet](https://github.com/across-protocol/contracts-v2/blob/3676377398714bd8f734485deead75afbcd8e8c4/contracts/HubPool.sol#L120)
+   at the time of proposal.
+4. No obvious griefing or manipulation of the system is being performed via this proposal.
```
Reinis-FRP, line 274: As this requires some human judgement, it would be helpful somewhere above in Summary or Motivation to include document link on overall architecture of Across v2 and how the system is supposed to be working.

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,276 @@
...
+Once the leaves are constructed, the merkle root can be constructed by hashing each leaf data structure using
+Solidity's standard process of `keccak256(abi.encode(poolRebalanceLeaf))`. Once the leaves are hashed, the tree should
+be constructed in the standard way such that it is verifyable using
+[OpenZeppelin's MerkleProof](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/742e85be7c08dff21410ba4aa9c60f6a033befb8/contracts/utils/cryptography/MerkleProof.sol)
+library. See examples [here](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/742e85be7c08dff21410ba4aa9c60f6a033befb8/test/utils/cryptography/MerkleProof.test.js)
+for how to construct these types of trees.
+
+## Constructing RelayerRefundRoot
+
+For each grouping in the section above that either a) has relayer refunds or b) has a negative net send amount, a
```
nicholaspai, line 226: Can you more specifically link to how this "grouping" is made? perhaps make its own section: "How to group FillRelay" events

UMIPs/UMIP-across-v2.md
nicholaspai, line 239: Can we just sort these by `refundAddress`? There won't be duplicate refundAddresses for an L2 token and it reduces dataworker complexity a bit

UMIPs/UMIP-across-v2.md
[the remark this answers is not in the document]
mrice32: Happy to distinguish. Where did the 100-200 number come from? I assumed both of these operations would cost up to 400k gas with a 10MM block gas limit. That might be a lot for a token transfer, so maybe we could bring that down to 200k worst case (aka 50), but I don't know if I would feel comfortable going lower.

The UMA token, for instance, has transfers that cost up to 128k gas alone. Not saying that's expected to be the norm, but I think for hard upper limits, we might want to be pretty conservative, so something like 200k per transfer with a 10MM gas limit.

Thoughts?

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,276 @@
...
+
+Note 3: wherever unspecified, sorting should be ascending by default, not descending.
+
+## Constants
+
+The following constants are currently specified in this UMIP directly, but should be moved to an on-chain
+configuration in the future. Once that happens, this UMIP can be amended to specify where to pull them from.
+
+MAX_LEAF_ARRAY_SIZE = 25
+TOKEN_TRANSFER_THRESHOLD = 1%
```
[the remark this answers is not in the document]
mrice32: Will do!

UMIPs/UMIP-across-v2.md
mrice32: @nicholaspai @chrismaree just for the sake of this UMIP, thoughts on hardcoding this list with an understanding that this UMIP will be amended once we add a config contract to store it?

UMIPs/UMIP-across-v2.md
mrice32: I think this is referring to the list of l1Tokens for a particular chain's rebalance leaf, which I think would need to be limited, right, so we don't run out of gas trying to go through all of them on mainnet, right?

UMIPs/UMIP-across-v2.md
mrice32: Yes. I can add more specificity here.

UMIPs/UMIP-across-v2.md
mrice32: I can add this.

UMIPs/UMIP-across-v2.md
mrice32: I see what you're saying. I thought they were already grouped that way, but will double check the language here.

UMIPs/UMIP-across-v2.md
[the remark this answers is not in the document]
nicholaspai: Agreed let's be super convenient, this is just based on very simple unit tests. Let's stick with 25 for now but we should distinguish between the two types of limits

UMIPs/UMIP-across-v2.md
nicholaspai: Sounds good, what's an example hard coding though?

UMIPs/UMIP-across-v2.md
```diff
@@ -42,13 +29,29 @@ sort should be on the transactionIndex, and the tertiary sort should be on the l
...
+```
+ooRequester:0x69CA24D3084a2eea77E061E2D7aF9b76D107b4f6
+```
+
 ## Constants
 
 The following constants are currently specified in this UMIP directly, but should be moved to an on-chain
 configuration in the future. Once that happens, this UMIP can be amended to specify where to pull them from.
 
-MAX_LEAF_ARRAY_SIZE = 25
```
Reinis-FRP: Rest of the text is still referencing `MAX_LEAF_ARRAY_SIZE`

mrice32: Fixed!

UMIPs/UMIP-across-v2.md
mrice32: Done.

UMIPs/UMIP-across-v2.md
mrice32: Done.

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,309 @@
...
+
+To determine the amount to modify the running balances:
+
+1. For each `FilledRelay` group above, initialize a running balance value at 0 and add the total relay repayments to
+   it. Each running balance value is defined by its `chainId` and `l1Token`.
+2. Find all `FundsDeposited` events on all `SpokePool`s from earlier within each `SpokePool`'s block range. Using the
+   `originChainId`, `originToken`, the `quoteTimestamp`, and the `SetPoolRebalanceRoute` event on the HubPool, use a
+   similar process as the 3 step one above to map this back to an `l1Token`. For that `l1Token` and `originChainId`,
+   initialize a running balance value if one doesn't exist already and subtract the `amount` from it.
+3. Find all `FilledRelay` events in the block range that have `isSlowRelay` set to true. For each, use a similar method
```
mrice32, line 202: Modified this to align with what @nicholaspai and I discussed yesterday.

This basically finds the intended slow relay amount and the amount of funds that were actually used and sends back any unused funds.

UMIPs/UMIP-across-v2.md
nicholaspai: Ah nevermind you're good

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,309 @@
...
+
+## Preliminary Information
+
+The ooRequester address is expected to be an instance of the
+[HubPool contract](https://github.com/across-protocol/contracts-v2/blob/aac42df9192145b5f4dc17162ef229c66f401ebe/contracts/HubPool.sol).
+
+If any of the expected details in the ooRequester are not available in the expected form because the HubPool does not
+match the expected interface, the identifier should return 0.
+
+To get the proposal data, the voter should find events that matches
```
nicholaspai, line 82:
```suggestion
To get the proposal data, the voter should find events that match
```

UMIPs/UMIP-across-v2.md
```diff
@@ -0,0 +1,309 @@
...
+
+## Constructing the PoolRebalanceRoot
+
+To construct the `poolRebalanceRoot`, you need to form a list of rebalances.
+
+For all valid `FilledRelay` events above, group them by `repaymentChainId` and their associated `l1Token` found above.
+
+For each group, sum the `fillAmount` values to get the total relay repayments for that group.
+
+Similarly, sum the `fillAmount * relayerFeePct / 1e18` to get the total LP fees for that group.
```
nicholaspai, line 192: Shouldn't this be `realizedLpFeePct` if we're computing LP fees?

mrice32: Yes, this is a typo!

mrice32: Done.

nicholaspai (approved):
LGTM, i'm sure we'll need to make edits later but this is a great start

nicholaspai (approved):
LGTM, great start that we'll make edits to later

PR #511, opened 2022-04-06
Update umip-140.md

binomfx:
Update UMIP-140: shares list changed

### DIFF
--- a/UMIPs/umip-140.md
+++ b/UMIPs/umip-140.md
@@ -9,7 +9,7 @@
 
 ## SUMMARY
 The DVM should support price requests for uSPAC10 price identifier<br>
-The purpose of these price identifier is to create synthetic token, price of which is linked to the value of index of **10** most active SPACs (Special Purpose Acquisition Companies) shares.<br> That synthetic token can be used for creating speculative strategies at IPO market.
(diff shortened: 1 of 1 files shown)


UMIPs/umip-140.md
```diff
@@ -441,7 +441,7 @@ axios.request(options).then(function (response) {
 ## SECURITY CONSIDERATIONS
 Security considerations are focused on the use of the token price for monitoring collateral ratios.
 
-The risk of manipulation of stock quotes included in the index is insignificant because a reliable source of quotes is used. In addition, users - wouters and disputers - have the opportunity to check the calculation of the index value based on independent sources of quotations themselves.
+The risk of manipulation of stock quotes included in the index is insignificant because a reliable source of quotes is used. In addition, users - vouters and disputers - have the opportunity to check the calculation of the index value based on independent sources of quotations themselves.
```
smb2796, line 444:
```suggestion
The risk of manipulation of stock quotes included in the index is insignificant because a reliable source of quotes is used. In addition, users - voters and disputers - have the opportunity to check the calculation of the index value based on independent sources of quotations themselves.
```

smb2796 (approved):

PR #515, opened 2022-05-04
UMIP-136: Update rounding guidelines

mrice32:


nicholaspai (approved):
This looks good! We should discuss ordering of updating bots and merging in this UMIP

Reinis-FRP (approved):
Sounds good!

mrice32:
Closing this as we should only apply it to the across v2 identifier.

PR #522, opened 2022-05-27
docs: update umip-152 based on audit feedback

pemulis:

This pull request updates UMIP-152 for the `ZODIAC` identifier to better reflect the revised contract code, proposal/dispute/execution/deletion flows, and some additional considerations around the use of transitional administrative multi-sigs that can delete proposals in the event of an emergency.

### DIFF
--- a/UMIPs/umip-152.md
+++ b/UMIPs/umip-152.md
@@ -125,7 +125,23 @@ The `ancillaryData` for the price request will be generated automatically and co
 
 `proposalHash: 0x...abcdef`
 
-### Resolving Disputes
+## Rationale
+It is impossible to capture every form of human organization in rigid programmatic structures. Without a flexible, natural language based system for outlining rules of operation, and a trustless and decentralized mechanism for enforcing those rules, DAOs can only exist in a stunted form, like an oak tree growing in a pot with no space for deep roots.
+
+This UMIP and the associated module code brings the flexibility of law to smart contracts, with UMA voters serving as judge and jury in case of disputes. Unlike the tradition of common law, however, UMA voters do not attempt to create precedent for ambiguous cases, resolve controversy, or fill in the gaps. If there is ambiguity about whether a particular proposal follows the rules, it is simply rejected.
+
+The onus is on the DAO to write clear rules and on the proposer to ensure their transactions follow the rules, with no room for ambiguity by a neutral third-party observer.
+
+This methodology allows for a huge amount of flexibility for DAOs to manage their shared resources without requiring UMA voters to make subjective judgements, which would be difficult for voters and could potentially create unpredictable and undesirable results for DAOs.
+
+## Implementation
+Voters should read the plain language rules referenced in the public `rules` value in the Optimistic Governor contract that requested a price and check the proposed transactions against those rules.
+
+This `rules` value will usually be an IPFS hash. If it is, voters can view the rules through a web browser by going to https://ipfs.io/ipfs/<RULES_HASH> or by using any other system to access the document using IPFS. If the `rules` value is a web URI, voters can simply enter the web URI into any web browser.
+
+The system is quite flexible, so voters may need to find other publicly verifiable data to determine if the proposed transactions follow the rules. For instance, they may need to check the result of a Snapshot vote to verify that a governance proposal was approved by the DAO token holders, or check the current price of a token to verify a proposal to make a trade according to a trading strategy outlined in the rules, or confirm public misbehavior by some address to verify a proposal to slash a bond posted by that address.
+
+If the proposer included an optional `explanation` in the ancillary data of the price request, that may be helpful for understanding the intent of the proposal, but voters should primarily consider the actual effects of the proposed transactions and verify that the transactions clearly and unambiguously follow the rules.
 
 If a proposal is disputed, the disputer is encouraged to publish their reasoning in a public forum to assist UMA voters in their determination of whether or not the proposal was valid.
 
@@ -148,17 +164,11 @@ It is the responsibility of the DAO users of the `ZODIAC` identifier to write cl
 
 It is important that the process for changing the rules be particularly clear and unambiguous, if a rule change process is allowed by the DAO. This allows the DAO to add, remove, or clarify rules as needed, analogous to a constitutional amendment.
 
-## Rationale
-It is impossible to capture every form of human organization in rigid programmatic structures. Without a flexible, natural language based system for outlining rules of operation, and a trustless and decentralized mechanism for enforcing those rules, DAOs can only exist in a stunted form, like an oak tree growing in a pot with no space for deep roots.
-
-This UMIP and the associated module code brings the flexibility of law to smart contracts, with UMA voters serving as judge and jury in case of disputes. Unlike the tradition of common law, however, UMA voters do not attempt to create precedent for ambiguous cases, resolve controversy, or fill in the gaps. If there is ambiguity about whether a particular proposal follows the rules, it is simply rejected.
-
-The onus is on the DAO to write clear rules and on the proposer to ensure their transactions follow the rules, with no room for ambiguity by a neutral third-party observer.
-
-This methodology allows for a huge amount of flexibility for DAOs to manage their shared resources without requiring UMA voters to make subjective judgements, which would be difficult for voters and could potentially create unpredictable and undesirable results for DAOs.
-
-## Implementation
 The implementation of the Optimistic Governor module can be [found here](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/zodiac/OptimisticGovernor.sol).
+
+The requester will be the Optimistic Governor contract itself, which means voters can use the requester address to find the contract on Etherscan and inspect the public `rules` value.
+
+Voters can use the [`DecodeTransactionData` script](https://github.com/UMAprotocol/protocol/blob/master/packages/scripts/src/DecodeTransactionData.js) to decode transaction data as needed.
 
 ### Example Rules
 The `ZODIAC` identifier is designed to allow rules to be as flexible as possible while still being clear to UMA voters called in to resolve disputes. These examples are meant to inspire creativity in users creating their own rules and demonstrate the legalistic approach they should take.
--- a/UMIPs/umip-152.md
+++ b/UMIPs/umip-152.md
@@ -168,8 +168,6 @@ The implementation of the Optimistic Governor module can be [found here](https:/
 
 The requester will be the Optimistic Governor contract itself, which means voters can use the requester address to find the contract on Etherscan and inspect the public `rules` value.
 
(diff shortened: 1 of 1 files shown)


UMIPs/umip-152.md
```diff
@@ -50,56 +72,67 @@ function proposeTransactions(Transaction[] memory transactions, bytes memory exp
...
 If a proposal is disputed, the disputer is encouraged to publish their reasoning in a public forum to assist UMA voters in their determination of whether or not the proposal was valid.
 
 Voters should consider four sources of public information in their ruling:
 
-1. The published rules for the Zodiac module.
-2. The rationale given by the proposer in ancillary data.
+1. The published rules for the Optimistic Governor module.
+2. The rationale given by the proposer in explanation (if any).
 3. The arguments presented by disputers.
 4. The voter's understanding of what the transactions *actually* do, regardless of the stated rationale.
```
mhairimcalpine, line 137: This last point is important, and I can see it getting overlooked, or people bamboozled in arguments of rights and wrongs of a proposal (say "donate $5 to saving tomatos") when the rules say only vegetable charities can be donated to, and not noticing that the actual tx says send the entire contents of the treasury to scammer.eth, particularly if the transaction is complex.

I'd want to think about some support for UMA tokenholders here, which brought this to the fore so that tokenholders could be certain about what proposals they were *actually* comparing to the rules.

UMIPs/umip-152.md
```diff
@@ -109,15 +142,15 @@ If a proposal doesn't follow the rules, an UMA voter should return a value of `0
...
 
 ### Important Special Case: Rule Changes
 
-It is possible for a bad proposal to be made that technically follows the published rules, either by a malicious actor or due to an oversight or mistake. It is also possible for the rules in a Zodiac module to be changed via the existing governance process.
+It is possible for a bad proposal to be made that technically follows the published rules, either by a malicious actor or due to an oversight or mistake. It is also possible for the rules in an Optimistic Governor module to be changed via the existing governance process or an emergency adminstrative action by the multi-sig (if any).
 
 As a failsafe mechanism, it should be allowed for a DAO to change their rules in response to malicious proposals, and in those cases the new rules should be considered the official rules by UMA voters, and disputed proposals should be rejected if they violate the old rules, even if they technically followed the rules at the time of the original proposal.
```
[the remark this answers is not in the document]
pemulis: It's going to be the responsibility of the DAO to maintain the rules in a secure way. Our recommendation in documentation will be to pin them on IPFS but it's flexible enough that they could point towards a URI.

UMIPs/umip-152.md
```diff
@@ -6,40 +6,62 @@
...
     Transaction[] transactions;
     uint256 requestTime;
-    bytes ancillaryData;
-    bool status;
 }
 ```
 
-The module contract stores an array of `Proposal` objects and a string reference to an off-chain set of rules that have been publicly published, which may be an IPFS hash, a URI, or something else.
+The module contract stores a mapping of proposal hashes to their proposal time, to verify proposals during execution and to prevent duplicate proposals. The contract also stores a string reference to an off-chain set of rules that have been publicly published, which may be an IPFS hash, a URI, or something else.
+
```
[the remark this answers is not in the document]
pemulis: It could actually reference a document on someone's Google Drive, theoretically, because it leaves that responsibility up to the DAO. The recommended method would be to pin the rules on IPFS, though. Linking to a particular commit on GitHub would be somewhere in between.

You could also store the entire rules on-chain in the `rules` string but it would get expensive really fast if the rules were long. That would be the most secure way if you have short and simple rules.

UMIPs/umip-152.md
```diff
@@ -6,40 +6,62 @@
...
+uint64 public liveness;
+// Extra bond in addition to the final fee for the collateral type.
+uint256 public bondAmount;
 string public rules;
+// This will usually be "ZODIAC" but a deployer may want to create a more specific identifier.
+bytes32 public identifier;
+OptimisticOracleInterface public optimisticOracle;
 ```
 
 When a user creates a proposal, they submit an array of transactions along with an optional explanation that explains the intent and purpose of the transactions, which is useful for voters trying to understand the transactions and whether or not they follow the published rules.
```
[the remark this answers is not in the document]
pemulis: That's  point. In the OG case, the explanation would be emitted in events and could be consumed elsewhere. For Polymarket, we could store the explanation in our own back-end and pull it into the interface from there.

The expected flow here is that people will enter the transactions and explanation through SafeSnap, at least for the initial use cases of the OG. If people were proposing transactions directly through an interface we control, maybe an expanded version of the current OO interface, it might make sense to allow explanations for other types of proposals even though those explanations won't interact with the chain.

UMIPs/umip-152.md
```diff
@@ -50,56 +72,67 @@ function proposeTransactions(Transaction[] memory transactions, bytes memory exp
...
-- `id`, which is automatically appended to the ancillary data and used to find the proposal details in the module
-- `module`, which is the address of the module contract
+The `ancillaryData` for the price request will be generated automatically and consist of the word `proposalHash` as a key, followed by a colon, followed by the proposal hash automatically generated from the transaction details.
 
-`explanation:These transactions were approved by the majority of tokenholders in a Snapshot vote on March 8, 2022.,id:123,module:0xabc...123`
+`proposalHash: 0x...abcdef`
 
 ### Resolving Disputes
 
 If a proposal is disputed, the disputer is encouraged to publish their reasoning in a public forum to assist UMA voters in their determination of whether or not the proposal was valid.
```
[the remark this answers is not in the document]
pemulis: I think we should figure out how to surface the best for/against opinions on disputed proposals without relying on the original proposer and disputer in every case. I don't think the explanation attached to a proposal will be a rigorous defense, for instance, it will just do something like point towards a Snapshot vote that passed. If an actual dispute arose, the "for" camp would need to go into more detail on other channels and it would be a similar problem to the debates around disputed prediction market proposals. For voter UX, we need to help them filter through the noise.

UMIPs/umip-152.md
pemulis: I think it's beyond the scope of the UMA voter system to check the transaction details every time but if the "against" camp can demonstrate that the transactions themselves are a trick, even though they were (for instance) approved by a Snapshot vote, it's something they can consider. Hopefully that will be a rare occurrence and it wouldn't be too much of a burden for our technical team members to take a look at the actual transactions to figure out what they think is the truth.

UMIPs/umip-152.md
```diff
@@ -50,56 +72,67 @@ function proposeTransactions(Transaction[] memory transactions, bytes memory exp
...
+
+Disputers and voters can find the `rules` reference in the Optimistic Governor contract. This contract address will be the `requester` in the `ProposePrice` event emitted by the Optimistic Oracle contract.
 
-The `ancillaryData` submitted with proposals should consist of:
+### Ancillary Data Format
 
-- `explanation`, which is written in natural language by the proposer and explains the nature of the transactions and how they follow the module's rules
-- `id`, which is automatically appended to the ancillary data and used to find the proposal details in the module
-- `module`, which is the address of the module contract
+The `ancillaryData` for the price request will be generated automatically and consist of the word `proposalHash` as a key, followed by a colon, followed by the proposal hash automatically generated from the transaction details.
```
Reinis-FRP, line 124: Instructions in this identifier should be more elaborate on how to link `proposalHash` from OO price request ancillary data to proposed transactions in OG since the `TransactionsProposed` event does not include `proposalHash` that could be easily filtered by verifiers.

smb2796:
@pemulis was thinking about your rule change [example](https://github.com/UMAprotocol/UMIPs/blob/4b45bf63f0d257be6c2862ebecf3cd75280b2ed7/UMIPs/umip-152.md#important-special-case-rule-changes).

Do you still see this as necessary to explicitly call out? With the change to drop disputes, it seems like much less of a security concern. The rule change example situation is pretty complex, and somewhat DVM breaking in a way (imagine if a rule change comes in halfway through a commit period).

UMIPs/umip-152.md
```diff
@@ -163,9 +196,7 @@ The onus is on the DAO to write clear rules and on the proposer to ensure their
 This methodology allows for a huge amount of flexibility for DAOs to manage their shared resources without requiring UMA voters to make subjective judgements, which would be difficult for voters and could potentially create unpredictable and undesirable results for DAOs.
 
 ## Implementation
-The implementation of the Zodiac module is currently in a draft state [viewable here](https://github.com/UMAprotocol/protocol/pull/3843/files).
```
smb2796, line 166: It'd be beneficial to summarize how voters should arrive at resolution values before getting into examples. This is generally expected in implementation section. Same as how you have it here: https://github.com/UMAprotocol/UMIPs/blob/4b45bf63f0d257be6c2862ebecf3cd75280b2ed7/UMIPs/umip-152.md#resolving-disputes

And should also instruct that voters should take to decode transaction data (can reference the script we have in protocol), instructions to check for other pending/disputed rule-change transactions (if you want to keep this), instruction on needing to reference linked ruleset. 

Good rule of thumb is that a new voter should be able to reference only the summary of the implementation section and have everything they need to independently determine their vote.

UMIPs/umip-152.md
pemulis: The `TransactionsProposed` event was updated to include the `proposalHash`, so this should be relatively easy now.

pemulis:
@smb2796 you made a good point about the rule change example being unnecessary to lay out now that we're dropping disputed proposals from the OG contract and the DVM is only handling the bonds. I'll remove those parts.

UMIPs/umip-152.md
```diff
@@ -109,63 +158,17 @@ If a proposal doesn't follow the rules, an UMA voter should return a value of `0
...
-
-The onus is on the DAO to write clear rules and on the proposer to ensure their transactions follow the rules, with no room for ambiguity by a neutral third-party observer.
-
-This methodology allows for a huge amount of flexibility for DAOs to manage their shared resources without requiring UMA voters to make subjective judgements, which would be difficult for voters and could potentially create unpredictable and undesirable results for DAOs.
-
-## Implementation
-The implementation of the Zodiac module is currently in a draft state [viewable here](https://github.com/UMAprotocol/protocol/pull/3843/files).
-
-TODO: Script that can decode transactions for voters.
+Voters can use the [`DecodeTransactionData` script](https://github.com/UMAprotocol/protocol/blob/master/packages/scripts/src/DecodeTransactionData.js) to decode transaction data as needed.
```
Reinis-FRP, line 171: This script works only if proposed transaction interacts with contracts from UMA protocol repository.

smb2796: ahhh good point

pemulis: The OG contract currently lives in the UMA protocol repository but will probably be split out into a new repo for Outcome. Removed the link to the transaction decode script for now.

Reinis-FRP (approved):

PR #524, opened 2022-06-02
suTVL implementation doc

abg4:

The current version denominates values in ETH by converting the collateral prices to ETH. Realizing it might be easier to change this to USD and do a separate step for getting the ETH price so that it will be easier to use the defillama adapter when this expires. Let me know if you agree.

### DIFF
--- a/Implementations/suTVL-KPI.md
+++ b/Implementations/suTVL-KPI.md
@@ -51,9 +51,9 @@ Note: [Chainlist.org](https://chainlist.org/) can be a good reference if mapping
 
 ### Other Outcome TVL & contracts
 
-At the time of writing, the LSP contract is currently the only active Outcome contract that has a TVL value. Outcome plans to deploy other contracts, such as the Optimistic Governor, that will be included in this value.
+At the time of writing, the LSP contract is currently the only active Outcome contract that has a TVL value. Outcome will also include optimistic distributor and optimistic governor contracts.
 
-As new contracts become active, please note that this implementation document is subject to change and will be updated to include steps for Voters and Proposers on how to obtain the TVL calculation.
+OD and OG contracts do not follow factory pattern, so supported Outcome contracts will be published on docs.
 
 **Additional information for UMA DVM participants:**
       
--- a/Implementations/suTVL-KPI.md
+++ b/Implementations/suTVL-KPI.md
@@ -1,0 +1,67 @@
+## Title:
+SuperUMAn DAO TVL KPI Option
+
+## Summary:
+The SuperUMAn DAO wishes to use KPI Options to incentivize the community to continue the pursuit of increased TVL in Outcome financial contracts. This implementation doc is intended to be used with the General_KPI price identifier. These objectives and corresponding payouts are detailed in the Implementation section.
+
+## Intended Ancillary Data:
+Metric:Outcome TVL denominated in the price of 10k ETH,
+Method:"https://github.com/UMAprotocol/UMIPs/blob/master/Implementations/suTVL-KPI.md", Rounding:6, Scaling:0
+
+## Implementation:
+
+### LSP Contracts
+
+1. Go to the [networks](https://github.com/UMAprotocol/protocol/tree/master/packages/core/networks) folder in the [UMA protocol monorepo](https://github.com/UMAprotocol/protocol). Each network file is represented by a chain Id json file. For each, check for `LongShortPairCreator` key-value pairs. An example would be:
+
+    {
+    "contractName": "LongShortPairCreator",
+    "address": "0x439a990f83250FE2E5E6b8059F540af1dA1Ba04D"
+    }
+
+Check each network folders commit history to ensure all LongShortPairCreator active contracts are being included.
+
+Note: [Chainlist.org](https://chainlist.org/) can be a good reference if mapping chain ID to the network name.
+
+2. For each `address` value from step 1, fetch all `CreatedLongShortPair` events emitted by each LongShortPairCreator address and take note of the `longShortPair` parameter for each event.
+
+3. For the list of contract addresses returned from step 2, call `expirationTimestamp` on each contract. Remove contracts that have an `expirationTimestamp` value less than the request timestamp so that you have a list of only active contracts.
+
+4. For each address from the list from step 3, call the `collateralToken` method to retrieve the collateral token address for each LSP contract.
+
+5. Call the `balanceOf()` method on the collateral contract from step 4 using the LSP contract address from step 2 as the argument at the latest available block at or before the request timestamp. This retrieves the value of the collateral token deposited into the LSP contract.
+
+6. Scale down the balances returned from Step 5 by calling the `decimals()` method on the collateral token contracts from Step 4.
+
+7. For each returned values from Step 6, retrieve the latest available pricing at or before each of the evaluation timestamps from CoinGecko:
+    * Based on CoinGecko [API documentation](https://www.coingecko.com/api/documentations/v3#/contract/get_coins__id__contract__contract_address__market_chart_range) construct price API request with the following parameters:
+      * `id`: CoinGecko platform id - should be set based on the chainId from the networks folder;
+      * `contract_address`: ERC-20 collateral token address from step 4;
+      * `vs_currency`: Should be set to "eth";
+      * `from`: should be set to a timestamp before the request timestamp that returns pricing intervals that are not less frequent than 1 hour to avoid inconsistent price intervals;
+      * `to`: end timestamp (identified by `expirationTimestamp`);
+    * Example API request for BOBA token pricing in USD: https://api.coingecko.com/api/v3/coins/ethereum/contract/0x42bbfa2e77757c645eeaad1655e0911a7553efbc/market_chart/range?vs_currency=eth&from=1646092800&to=1648771200
+    * Locate the `prices` key value from CoinGecko API response - it should contain a list of [ timestamp, price ] value pairs. Note that returned CoinGecko timestamps are in milliseconds;
+    * In case returned pricing interval is more frequent than 1 hour voters should extend the requested time range. This is necessary to have consistent pricing results independent of when the voters are calculating them.
+    * Voters should verify that obtained price results agree with broad market consensus.
+
+8. Multiply the values returned from step 6 and step 7. 
+
+9. Sum the total values from step 8 and divide the value by 10,000. Round the returned value to 6 decimal places.
+
+### Other Outcome TVL & contracts
+
+At the time of writing, the LSP contract is currently the only active Outcome contract that has a TVL value. Outcome plans to deploy other contracts, such as the Optimistic Governor, that will be included in this value.
+
+As new contracts become active, please note that this implementation document is subject to change and will be updated to include steps for Voters and Proposers on how to obtain the TVL calculation.
+
+**Additional information for UMA DVM participants:**
+      
+## Intended Application:
+It is intended to deploy the documented KPI Options on Polygon network using [LSP Contract](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol) with 'General_KPI' price identifier approved in [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md). This contract would use [Linear LSP FPL](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol) with the 'lowerBound' set to 0 and the 'upperBound' set to 1.
+
+'collateralPerPair' parameter for the LSP contract would be set to 1 so that the maximum payout per KPI option would reach 1 UMA if the max value denominated in 10k ETH is reached at the request timestamp.
+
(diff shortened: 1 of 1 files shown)


Implementations/suTVL-KPI.md
```diff
@@ -0,0 +1,67 @@
...
+
+3. For the list of contract addresses returned from step 2, call `expirationTimestamp` on each contract. Remove contracts that have an `expirationTimestamp` value less than the request timestamp so that you have a list of only active contracts.
+
+4. For each address from the list from step 3, call the `collateralToken` method to retrieve the collateral token address for each LSP contract.
+
+5. Call the `balanceOf()` method on the collateral contract from step 4 using the LSP contract address from step 2 as the argument at the latest available block at or before the request timestamp. This retrieves the value of the collateral token deposited into the LSP contract.
+
+6. Scale down the balances returned from Step 5 by calling the `decimals()` method on the collateral token contracts from Step 4.
+
+7. For each returned values from Step 6, retrieve the latest available pricing at or before each of the evaluation timestamps from CoinGecko:
```
smb2796, line 36:
```suggestion
7. For each returned values from Step 6, retrieve the latest available pricing at or closest and before each of the evaluation timestamps from CoinGecko:
```

Implementations/suTVL-KPI.md
```diff
@@ -0,0 +1,67 @@
...
+      * `id`: CoinGecko platform id - should be set based on the chainId from the networks folder;
+      * `contract_address`: ERC-20 collateral token address from step 4;
+      * `vs_currency`: Should be set to "eth";
+      * `from`: should be set to a timestamp before the request timestamp that returns pricing intervals that are not less frequent than 1 hour to avoid inconsistent price intervals;
+      * `to`: end timestamp (identified by `expirationTimestamp`);
+    * Example API request for BOBA token pricing in USD: https://api.coingecko.com/api/v3/coins/ethereum/contract/0x42bbfa2e77757c645eeaad1655e0911a7553efbc/market_chart/range?vs_currency=eth&from=1646092800&to=1648771200
+    * Locate the `prices` key value from CoinGecko API response - it should contain a list of [ timestamp, price ] value pairs. Note that returned CoinGecko timestamps are in milliseconds;
+    * In case returned pricing interval is more frequent than 1 hour voters should extend the requested time range. This is necessary to have consistent pricing results independent of when the voters are calculating them.
```
smb2796, line 45: Can you explain this some more? Not sure what it means

Implementations/suTVL-KPI.md
```diff
@@ -0,0 +1,67 @@
...
+    * Voters should verify that obtained price results agree with broad market consensus.
+
+8. Multiply the values returned from step 6 and step 7. 
+
+9. Sum the total values from step 8 and divide the value by 10,000. Round the returned value to 6 decimal places.
```
smb2796, line 50: Just because this is pretty involved and there might be some variance in data, it might be better to use less precision. 6 decimals worth of UMA doesn't seem super necessary, feel like it would be just as good with 2-3 and be less susceptible to inconsequential disagreements on proposed settlement value.

Implementations/suTVL-KPI.md
```diff
@@ -0,0 +1,67 @@
...
+
+Note: [Chainlist.org](https://chainlist.org/) can be a good reference if mapping chain ID to the network name.
+
+2. For each `address` value from step 1, fetch all `CreatedLongShortPair` events emitted by each LongShortPairCreator address and take note of the `longShortPair` parameter for each event.
+
+3. For the list of contract addresses returned from step 2, call `expirationTimestamp` on each contract. Remove contracts that have an `expirationTimestamp` value less than the request timestamp so that you have a list of only active contracts.
+
+4. For each address from the list from step 3, call the `collateralToken` method to retrieve the collateral token address for each LSP contract.
+
+5. Call the `balanceOf()` method on the collateral contract from step 4 using the LSP contract address from step 2 as the argument at the latest available block at or before the request timestamp. This retrieves the value of the collateral token deposited into the LSP contract.
```
smb2796, line 32: could be useful to include reference to this script since this could be pretty involved w/ bunch of contracts + networks: https://github.com/UMAprotocol/protocol/blob/master/packages/affiliates/liquidity-mining/FindBlockAtTimeStamp.js

Implementations/suTVL-KPI.md
```diff
@@ -0,0 +1,67 @@
...
+    * In case returned pricing interval is more frequent than 1 hour voters should extend the requested time range. This is necessary to have consistent pricing results independent of when the voters are calculating them.
+    * Voters should verify that obtained price results agree with broad market consensus.
+
+8. Multiply the values returned from step 6 and step 7. 
+
+9. Sum the total values from step 8 and divide the value by 10,000. Round the returned value to 6 decimal places.
+
+### Other Outcome TVL & contracts
+
+At the time of writing, the LSP contract is currently the only active Outcome contract that has a TVL value. Outcome plans to deploy other contracts, such as the Optimistic Governor, that will be included in this value.
```
smb2796, line 54: Maybe we should be more specific and just say:
- This will also include optimistic distributor and optimistic governor contracts
- OD and OG contracts do not follow factory pattern, so supported Outcome contracts will be published on docs

Implementations/suTVL-KPI.md
```diff
@@ -0,0 +1,67 @@
+## Title:
+SuperUMAn DAO TVL KPI Option
+
+## Summary:
+The SuperUMAn DAO wishes to use KPI Options to incentivize the community to continue the pursuit of increased TVL in Outcome financial contracts. This implementation doc is intended to be used with the General_KPI price identifier. These objectives and corresponding payouts are detailed in the Implementation section.
+
+## Intended Ancillary Data:
+Metric:Outcome TVL denominated in the price of 10k ETH,
```
mhairimcalpine, line 8: Metric: TVL in UMA LSP contracts denominated in the price of 10k ETH,

Implementations/suTVL-KPI.md
mhairimcalpine: @smb2796 final value should be in eth not UMA, 6 decimals seems appropriate

Implementations/suTVL-KPI.md
```diff
@@ -0,0 +1,67 @@
...
+
+At the time of writing, the LSP contract is currently the only active Outcome contract that has a TVL value. Outcome plans to deploy other contracts, such as the Optimistic Governor, that will be included in this value.
+
+As new contracts become active, please note that this implementation document is subject to change and will be updated to include steps for Voters and Proposers on how to obtain the TVL calculation.
+
+**Additional information for UMA DVM participants:**
+      
+## Intended Application:
+It is intended to deploy the documented KPI Options on Polygon network using [LSP Contract](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol) with 'General_KPI' price identifier approved in [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md). This contract would use [Linear LSP FPL](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol) with the 'lowerBound' set to 0 and the 'upperBound' set to 1.
+
```
mhairimcalpine, line 62: The KPI options would be on mainnet, not polygon.

Implementations/suTVL-KPI.md
abg4: The below is from the Coingecko docs so this was originally included to avoid proposers and voters using different intervals if there was a dispute. What do you think?

```
For historical data, granularity is automatically determined and cannot be adjusted with the following rules:

1 day from query time = 5 minute interval data
1 - 90 days from query time = hourly data
above 90 days from query time = daily data (00:00 UTC)
```

Implementations/suTVL-KPI.md
abg4: The sum of the values from step 8 is the Outcome TVL denominated in eth. Step 9 is calculating the value that will be proposed so the rounding will determine the payout in UMA. I updated to 3 but let me know if either of you disagree.

Implementations/suTVL-KPI.md
abg4: I agree. Updated to include the script.

Implementations/suTVL-KPI.md
abg4: I used Outcome TVL to keep it open for other contracts that do not use the LSP contract (OG/OD). If you think it would be better, I could update to: 

`Metric: TVL in UMA LSP, OG, and OD contracts denominated in the price of 10k ETH,`

Implementations/suTVL-KPI.md
abg4: Good catch, updated!

Implementations/suTVL-KPI.md
smb2796: eh whatever. I'm impartial, but technically `at or before` isn't telling you that it is the closest and before. Just needs to be before to meet criteria :)

Implementations/suTVL-KPI.md
smb2796: > final value should be in eth not UMA, 6 decimals seems appropriate

@mhairimcalpine It's not, since in context the returned value is denominated in UMA since it is turned into a percentage of the 1 UMA collateral. 3 seems good to me

Implementations/suTVL-KPI.md
smb2796: So you're saying that you should always use hourly intervals and use the price that falls closest/before the request timestamp?

Implementations/suTVL-KPI.md
```diff
@@ -0,0 +1,67 @@
...
+## Intended Ancillary Data:
+Metric:TVL in UMA LSP, OG, and OD contracts denominated in the price of 10k ETH,
+Method:"https://github.com/UMAprotocol/UMIPs/blob/master/Implementations/suTVL-KPI.md", Rounding:3, Scaling:0
+
+## Implementation:
+
+### LSP Contracts
+
+1. Go to the [networks](https://github.com/UMAprotocol/protocol/tree/master/packages/core/networks) folder in the [UMA protocol monorepo](https://github.com/UMAprotocol/protocol). Each network file is represented by a chain Id json file. For each, check for `LongShortPairCreator` key-value pairs. An example would be:
```
smb2796, line 15: Any worries about manipulation attempts? Should there be any time weighting on token balances?

Implementations/suTVL-KPI.md
Reinis-FRP: no, in case there is pricing for exact timestamp we want to use that, not the previous, hence the need for `at`

Implementations/suTVL-KPI.md
abg4: Yea, I think there should be a time weighting on token balances if @mhairimcalpine is ok with that? Would 5 days be good?

Implementations/suTVL-KPI.md
```diff
@@ -0,0 +1,67 @@
+## Title:
+SuperUMAn DAO TVL KPI Option
+
+## Summary:
+The KPI options minted through this implementation document are intended for distribution to the SuperUMAn DAO to incentive their members to continue the pursuit of increased TVL in Outcome financial contracts. This implementation doc is intended to be used with the General_KPI price identifier. These objectives and corresponding payouts are detailed in the Implementation section.
+
+## Intended Ancillary Data:
+Metric:TVL in UMA LSP, OG, and OD contracts denominated in the price of 10k ETH,
```
Reinis-FRP, line 8:
```suggestion
Metric:"TVL in UMA LSP, OG, and OD contracts denominated in the price of 10k ETH",
```

Implementations/suTVL-KPI.md
Reinis-FRP: yes, we want fixed granularity. otherwise proposer might have more granular price on expiration, but if its disputed this might not be available for UMA voters to verify.

Implementations/suTVL-KPI.md
```diff
@@ -0,0 +1,67 @@
...
+
+Check each network folders commit history to ensure all LongShortPairCreator active contracts are being included.
+
+Note: [Chainlist.org](https://chainlist.org/) can be a good reference if mapping chain ID to the network name.
+
+2. For each `address` value from step 1, fetch all `CreatedLongShortPair` events emitted by each LongShortPairCreator address and take note of the `longShortPair` parameter for each event.
+
+3. For the list of contract addresses returned from step 2, call `expirationTimestamp` on each contract. Remove contracts that have an `expirationTimestamp` value less than the request timestamp so that you have a list of only active contracts.
+
+4. For each address from the list from step 3, call the `collateralToken` method to retrieve the collateral token address for each LSP contract.
```
Reinis-FRP, line 30: Maybe its worth clarifying whether this should also include LSP collateral at this `suTVL` implementation.

smb2796: If we don't want to reflect this, I think TVL range should be adjusted instead of changing implementation to specifically exclude. I don't see it as big issue though

Implementations/suTVL-KPI.md
smb2796: I think just a couple hours would be fine. 5 days could actually have large effect on payout (always a last minute rush to get integrations across the line before expiry).

Reinis-FRP (approved):

PR #525, opened 2022-06-08
Add umip162 optional callbacks and event-based requests

md0x:
Signed-off-by: Pablo Maldonado <pablo@umaproject.org>

UMIP-162 optional callbacks and event-based requests

> This UMIP will result in the creation of a new optimistic oracle contract with enhanced control over pricing request callbacks, which will only be executed if the requester has declared its readiness to receive them. This modification also eliminates the use of try-catch structures, which were vulnerable to gas griefing. Additionally, the contract includes a new event-based price request, which is suitable for requests that depend on events without a fixed date, thereby evaluating pricing requests at the time of proposal. This will enable the implementation of a more robust and adaptable price requests.

### DIFF
--- a/UMIPs/umip-161.md
+++ b/UMIPs/umip-161.md
@@ -1,0 +1,36 @@
+## Headers
+| UMIP-161   |                                 |
+|------------|---------------------------------|
+| UMIP Title | Add OptimisticOracleV2          |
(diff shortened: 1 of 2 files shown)


therealhexi (approved):
Looks good

Reinis-FRP (approved):

PR #532, opened 2022-06-28
Add Thorswap Volume KPI Option

abg4:

### DIFF
--- a/Implementations/thorswap-volume.md
+++ b/Implementations/thorswap-volume.md
@@ -50,7 +50,7 @@ group by 1
 
 ## Intended Application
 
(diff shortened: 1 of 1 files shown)


Implementations/thorswap-volume.md
```diff
@@ -0,0 +1,58 @@
...
+
+select date_trunc('month', block_timestamp) as month,
+sum(swap_volume) as ts_swap_volume,
+sum(ts_swap_volume) over (order by month) as cumulative_ts_swap_volume
+from base 
+group by 1
+```
+
+1. Use the `volume per month` query above to retrieve the Thorswap monthly trade values. It should contain a list of { MONTH, TS_SWAP_VOLUME, CUMULATIVE_TS_SWAP_VOLUME } values.
+2. For the values returned from step 1, locate the `MONTH` key value that corresponds with the parameter specified in the ancillary data.  Choose the `TS_SWAP_VOLUME` value returned in the same array. Round the difference to 0 decimal places (e.g. 72166475.9878698 -> 72166475).
```
Reinis-FRP, line 49: Why the "difference" should be rounded? Difference between what?

Reinis-FRP (approved):

Implementations/thorswap-volume.md
abg4: Good catch, I updated the wording to not use 'difference'

PR #539, opened 2022-07-15
Implementation fie for Humble Tractors

neondaemon3:
Please review


abg4:
Using the SuperUMAn repo to host implementation (https://github.com/SuperUMAn-DAO/UMIPs/blob/master/Implementations/humble-tractors.md).

PR #542, opened 2022-07-21
Add Subgraph query implementation 

Reinis-FRP:


Implementations/subgraph-query.md
```diff
@@ -0,0 +1,62 @@
...
+- `TimestampKey:<TIMESTAMP_KEY>` (optional) where `<TIMESTAMP_KEY>` represents relative key path to the timestamp of the returned metric within the returned collection of entities. `<TIMESTAMP_KEY>` can consist of single key string or concatenated key strings joined by dots (`.`) when nested elements should be processed. `TimestampKey` is mandatory if `DailyAggregation` was provided and set to `true`.
+- `ChainId:<CHAIN_ID>` (optional) where `<CHAIN_ID>` represents numeric chain identifier when query timestamp needs to be converted to the block number for the relevant chain.
+- `Interval:"Daily 24:00 UTC"` is set in order to limit aggregation processing overhead for manual verifiers and preserve composability with other data sources. Note that subgraph does not return data on fixed intervals hence effective intervals might vary (see Implementation for further details).
```
smb2796, line 23: This can be altered by deployer or no?

Implementations/subgraph-query.md
```diff
@@ -0,0 +1,62 @@
...
+- `SubgraphId:<SUBGRAPH_ID>` (optional) where `SUBGRAPH_ID` is the identifier of deployed subgraph when querying decentralized network.
+- `Source:<SUBGRAPH_SOURCE>` where `<SUBGRAPH_SOURCE>` is URL link to the repository containing source code for the relevant subgraph.
+- `QueryString:"<QUERY_STRING>"` where `<QUERY_STRING>` represents [GraphQL API](https://thegraph.com/docs/en/querying/graphql-api/) query string. `<QUERY_STRING>` can also include macros as detailed in the Implementation section. This implementation does not support traversing of paginated query results. Hence, all the information for querying the tracked metric should be returned in a single query request.
+- `CollectionKey:<COLLECTION_KEY>` (optional) where `<COLLECTION_KEY>` represents key path to collection containing tracked metrics under returned `data` object. `<COLLECTION_KEY>` can consist of single key string or concatenated key strings joined by dots (`.`) when nested elements should be processed. `CollectionKey` parameter should be passed only when the tracked metric is returned within a collection of entities and it is mandatory when `TimestampKey` was provided.
```
smb2796, line 19: I'm struggling to understand the relationship between collection_key and metric_key without an example. Can you provide an example of a "collection of entities"?

Implementations/subgraph-query.md
smb2796: It looks like from implementation section that is should not be. If that is the case I think it should just be removed as a parameter since having it as one somewhat implies that it is alterable

Implementations/subgraph-query.md
```diff
@@ -0,0 +1,62 @@
...
+- `RawRounding:<RAW_ROUNDING_DECIMALS>` (optional) where `<RAW_ROUNDING_DECIMALS>` is similar to `Rounding` value in [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md), but applied on raw metric.
+- `Rounding:<ROUNDING_DECIMALS>` where `<ROUNDING_DECIMALS>` represents `Rounding` value as per [UMIP-117](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-117.md) applied on resolved value.
+
+## Implementation
+
+1. Determine daily query timestamp for metric data:
+    - If `RequestTimestamp` parameter is provided its value should be used as an override for the actual price request timestamp. Though this should always be ignored if the provided `RequestTimestamp` value exceeds actual price request timestamp.
+    - Daily query timestamp should be set to the closest 24:00 UTC timestamp at or before effective request timestamp expressed as numeric Unix timestamp.
+2. Process `QueryString` parameter to replace any supported macros with actual integer values:
+    - `<QUERY_DTS>` should be replaced with the value of daily query timestamp.
```
smb2796, line 39: Where is `QUERY_DTS` from?

Implementations/subgraph-query.md
Reinis-FRP: No, the `Interval` should be fixed to daily for all deployments in order to support interoperability when performing operations with other data sources

smb2796: Shouldn't we just remove as a parameter and just include as part of the implementation then?

smb2796:
Don't have much to add, I think this is super cool. Main piece of feedback would be that I just think it could benefit from an example ancillary data construction, since it is pretty complex

Implementations/subgraph-query.md
smb2796: bump

Implementations/subgraph-query.md
smb2796: bump

Implementations/subgraph-query.md
```diff
@@ -0,0 +1,134 @@
...
+    -  `<QUERY_DBN-[N]D>` where `[N]` is expected to be integer value of days should be replaced with the latest block number that is at or before the daily query timestamp that had been reduced by `[N] * 86400` seconds. As an illustration, `<QUERY_DBN-30D>` would be replaced by the corresponding block number for the timestamp that is 30 days before the daily query timestamp.
+    - For `<QUERY_DBN>` and `<QUERY_DBN-[N]D>` macros above identify the relevant blockchain by the value of passed `ChainId` parameter (defaults to `1` for mainnet Ethereum if not explicitly provided). As a helper one can use Snapshot's [block finder query](https://blockfinder.snapshot.org), but verifiers are free to use their own tools to identify the matching block number.
+    - `<PAGINATE>` indicates that the requested data contains multiple entries (`CollectionKey` parameter must have been provided) and the query should be paginated by the verifier querying the API multiple times.
+        - On the first run `<PAGINATE>` should be replaced with `first:1000`. If the returned data contains exactly 1000 entries the query should be repeated by following the steps below.
```
smb2796, line 47: Is 1000 the max number of entries that a subgraph can return? Should this just be a default and then skip can be what is a custom param?

Implementations/subgraph-query.md
```diff
@@ -0,0 +1,134 @@
...
+The voters should also be aware that subgraphs can be upgraded during the lifetime of requesting contract, hence they should use their human judgement to detect and mitigate any attempted malicious subgraph upgrades.
+
+Users of this subgraph implementation should be aware that the Graph protocol is planning to sunset the hosted service in Q1 2023. Hence, all contracts that are expected to expire after end of 2022 should use the decentralized subgraph network and provide the `SubgraphId` parameter instead of `Endpoint` for the hosted service.
+
+## Example use case
+
+As an illustration the below configuration of ancillary data would track 90 day TWAP of Idle Finance Senior wstETH tranche measured in ETH:
+
+```
+Metric:"TVL of IdleCDO AA Tranche - wstETH measured in ETH",
```
smb2796, line 78:
```suggestion
Metric:"TVL of Idle CDO AA Tranche - wstETH measured in ETH",
```

Implementations/subgraph-query.md
```diff
@@ -0,0 +1,134 @@
...
+Method:"https://github.com/UMAprotocol/UMIPs/blob/master/Implementations/subgraph-query.md",
+Endpoint:"https://api.thegraph.com/subgraphs/name/samster91/idle-tranches",
+QueryString:"{trancheInfos(<PAGINATE>,orderBy:timeStamp,orderDirection:desc,where:{Tranche:\"0x2688fc68c4eac90d9e5e1b94776cf14eade8d877\",timeStamp_lte:<QUERY_DTS>,timeStamp_gte:<QUERY_DTS-90D>}){timeStamp,contractValue}}",
```
smb2796, line 81: Should `orderBy:timeStamp,orderDirection:desc` always be part of query? I don't see anything in the implementation that defines this, and (if I didn't just miss it) what is the rationale for it not being present?

Implementations/subgraph-query.md
```diff
@@ -0,0 +1,134 @@
...
+- `TimestampKey:<TIMESTAMP_KEY>` (optional) where `<TIMESTAMP_KEY>` represents relative key path to the timestamp of the returned metric within the returned collection of entities. `<TIMESTAMP_KEY>` can consist of single key string or concatenated key strings joined by dots (`.`) when nested elements should be processed. `TimestampKey` is mandatory if `DailyAggregation` was provided and set to `true`.
+- `ChainId:<CHAIN_ID>` (optional) where `<CHAIN_ID>` represents numeric chain identifier when query timestamp needs to be converted to the block number for the relevant chain.
+- `Interval:"Daily 24:00 UTC"` is set in order to limit aggregation processing overhead for manual verifiers and preserve composability with other data sources. Note that subgraph does not return data on fixed intervals hence effective intervals might vary (see Implementation for further details).
+- `RequestTimestamp:<REQUEST_TIMESTAMP>` (optional) where  `<REQUEST_TIMESTAMP>` represents override value for price request timestamp.
```
abg4, line 24: Not necessarily recommending a change but what are your thoughts on changing this to something like `RequestTimestampOverride` for it to be more obvious that we are overriding the request timestamp?

Implementations/subgraph-query.md
```diff
@@ -0,0 +1,134 @@
...
+- `AggregationPeriod:<AGGREGATION_PERIOD>` (optional) where `<AGGREGATION_PERIOD>` represents time period in seconds for any time series data processing. This parameter is mandatory if `AggregationMethod` was provided while `TimestampKey` was not provided.
```
abg4, line 25: Is there a fallback if both `AggregationMethod` and `TimestampKey` are not provided?

Implementations/subgraph-query.md
Reinis-FRP: Agree, `Interval` is redundant on-chain as `Method` document provides all the information on time intervals. I also opened #548 to make `Interval` optional in `General_KPI` and addressed it in all standardized data source implementations.

Implementations/subgraph-query.md
Reinis-FRP: Example section added at the end of document

Implementations/subgraph-query.md
Reinis-FRP: Ancillary data has `<QUERY_DTS>` and `<QUERY_DBN>` as literal strings in `QueryString`. At the time of request verifiers should replace these with actual timestamp or block number before making the query. I added illustration on this in the example section.

Implementations/subgraph-query.md
Reinis-FRP: I took the name (without space) from deployed [tranche contract](https://etherscan.io/token/0x2688fc68c4eac90d9e5e1b94776cf14eade8d877).

Implementations/subgraph-query.md
Reinis-FRP: Ordering is not required when the query returns single entry or elements should be aggregated over the returned collection. Also `orderBy` value is specific to subgraph schema (e.g. not all subgraphs have `timeStamp`, it might be named differently or does not have timestamps at all).

Implementations/subgraph-query.md
Reinis-FRP: Missing `AggregationMethod` would imply skipping Step 6 altogether.

But on similar note if both `AggregationPeriod` and `TimestampKey` were missing with present `AggregationMethod` this implies misconfiguration so I added fallback to skip aggregation in Step 6.

Implementations/subgraph-query.md
Reinis-FRP: Good suggestion! Fixed here and also adding to other standardized implementations

smb2796 (approved):

PR #551, opened 2022-08-10
Create umip-uma covered call options.md

dannydoritoeth:
This UMIT proposes to use a small portion of the protocol treasury to sell weekly covered call options on the Optix protocol on Polygon(https://optixprotocol.com).


dannydoritoeth:
Requested to be closed.

PR #552, opened 2022-08-24
Add DXY Price Identifier

coinop-logan:


UMIPs/umip-add-dxy.md
```diff
@@ -0,0 +1,66 @@
...
+
+The DVM should support price requests for the [US Dollar Index](https://en.wikipedia.org/wiki/U.S._Dollar_Index) (DXY).
+
+The canonical identifier should be `DXY`.
+
+# Motivation
+
+Adding DXY would allow synthetic assets for the US Dollar Index to be created, which would in turn allow various derivatives to be built on top of this token.
+
+DXY is a well-defined and publicly accessible metric, and so will be trivial to resolve by the Oracle.
```
mhairimcalpine, line 21: Would be useful to provide more info on what DXY does/is calculated

UMIPs/umip-add-dxy.md
```diff
@@ -0,0 +1,66 @@
...
+DXY is a calculated index that is very difficult to manipulate in the short term without exhorbitant cost, so no TWAP is needed.
+
+We chose to use the daily open price, updated daily, because this is easier to fetch via various APIs for historical data.
+
+# Implementation
+
+Voters should first determine the day of the request's timestamp.
+
+Voters should then go to https://www.tradingview.com/chart/?symbol=TVC%3ADXY, and make sure the interval is set to "1 day" (see B on the below image). Voters should then mouse-over the day determined, and look at the Open price (C on the below image) and record this number. This is the value for the TVC market.
+
```
mhairimcalpine, line 55: What if this source is unavailable?

UMIPs/umip-add-dxy.md
```diff
@@ -0,0 +1,66 @@
...
+![tradingview ux](images/tradingview-ux.png)
+
+Voters should then add these values up and divide by 3 to get the mean value, then round to 3 decimal places to arrive at the final value.
+
+Voters should determine whether the returned price differs from broad market consensus. This is meant to provide flexibility in any unforeseen circumstances as voters are responsible for defining broad market consensus.
+
+# Security considerations
+
+The inclusion of this requested price identifier should not present a security concern for DVM. There is a clearly defined method of determining a canonical price, and the data needed is publicly available.
```
mhairimcalpine, line 66: Security risks include
 - if used to create synthetic assets, there is a risk that liquidations and disputes will not be done in a timely manner
  - the resolution sources are unavailable.

UMIPs/umip-add-dxy.md
```diff
@@ -0,0 +1,66 @@
...
+
+# Summary
+
+The DVM should support price requests for the [US Dollar Index](https://en.wikipedia.org/wiki/U.S._Dollar_Index) (DXY).
+
+The canonical identifier should be `DXY`.
+
+# Motivation
+
+Adding DXY would allow synthetic assets for the US Dollar Index to be created, which would in turn allow various derivatives to be built on top of this token.
```
Reinis-FRP, line 19: Which contracts are expected to mint synthetic assets and make price requests for DXY?
Note that EMP contracts are not supported (and also not suited for daily data granularity) while LSP contract payouts are capped - is this suitable for intended application or new types of contracts would be developed?

UMIPs/umip-add-dxy.md
```diff
@@ -0,0 +1,66 @@
...
+- Example data providers:
+    - https://www.tradingview.com/symbols/TVC-DXY/
+    - https://www.marketwatch.com/investing/index/dxy
+
+- Real-time data update frequency:
+    - Daily
+
+# Price feed implementation
+
+Implemented as a price feed "plugin" [here](https://github.com/Signo-App/uma-protocol/blob/new-price-feed/packages/financial-templates-lib/src/price-feed/MarketStackPriceFeed.ts). Uses the MarketStack platform.
```
Reinis-FRP, line 36: Please also provide example price feed configuration.

UMIPs/umip-add-dxy.md
```diff
@@ -0,0 +1,66 @@
...
+
+# Rationale
+
+DXY is a calculated index that is very difficult to manipulate in the short term without exhorbitant cost, so no TWAP is needed.
+
+We chose to use the daily open price, updated daily, because this is easier to fetch via various APIs for historical data.
+
+# Implementation
+
+Voters should first determine the day of the request's timestamp.
```
Reinis-FRP, line 52: Please confirm that daily bounds are in UTC and also example data providers report daily data in UTC.

UMIPs/umip-add-dxy.md
coinop-logan: @Reinis-FRP Why do you say that EMPs aren't suited for daily data granularity?

We're aware that EMPs are no longer supported; however we plan to fork and keep using them ourselves.

UMIPs/umip-add-dxy.md
coinop-logan: Added clarification for determining the day, and have confirmed that TradingView always reports daily open/close according to the UTC time zone.

UMIPs/umip-add-dxy.md
Reinis-FRP: One risk is intra-day volatility that liquidators cannot act upon. Even if position should be liquidated based on real-time price DVM would resolve to daily opening price. If the ratio of collateral value relative to outstanding debt falls below 1 the position would not be liquidated on the next daily open price as that would incur loss for the liquidator. This can be mitigated by setting high collateral requirement.

UMIPs/umip-add-dxy.md
```diff
@@ -0,0 +1,103 @@
...
+- Real-time data update frequency:
+    - Daily
+
+# Price feed implementation
+
+Implemented as a price feed "plugin" [here](https://github.com/Signo-App/uma-protocol/blob/new-price-feed/packages/financial-templates-lib/src/price-feed/MarketStackPriceFeed.ts). Uses the MarketStack platform.
+
+An example price feed config to pass into this script is:
+
+`{"apiKey":"your_api_key_for_marketstack","type": "marketstack","lookback":604800, "symbolString":"DXY.INDX"}`
```
Reinis-FRP, line 49: This needs to be updated in line with Implementation section where average of 3 `symbolString` values is calculated. You can use either [ExpressionPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.ts) or [MedianizerPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.ts) with `computeMean` set to `true`

coinop-logan: Ah, we didn't realize that the price feed would have to arrive at exactly the same price as the DVM would return, although in retrospect that makes sense. Can you confirm that this is a requirement? In other words, if the price feed returns a slightly different result than the DMV would, that's a problem?

coinop-logan: Our problem is that MarketStack only has one source of data. Would it be acceptable for the Implementation to use this single source of data, or is it required to average a few sources as most UMIPs do?

murraci: I think the solution here is to pick a DXY price feed, i.e, Marketwatch, and then specify that the synth tracks that particular data service provider rather than DXY generally. That should be fine right?

Reinis-FRP: yes, and you would still have access to fallback method in case Marketwatch is not available or is deemed malitious

murraci: OK cool that should be fine. Thanks.

UMIPs/umip-add-dxy.md
coinop-logan: I've added a note to the security considerations.

Reinis-FRP (approved):

PR #554, opened 2022-09-11
fix: add WETH on Optimism

pemulis:

This UMIP will add the WETH token on Optimism as an approved collateral currency.

https://discourse.umaproject.org/t/add-weth-as-collateral-currency-on-optimism/1803


Reinis-FRP (commented):
Maybe should use the opportunity and add WETH as collateral on all the other supported networks?

UMIPs/umip-weth.md
```diff
@@ -0,0 +1,38 @@
+**UMIP #**  - tbd
+
```
mhairimcalpine, line 2: Should this go in as a full UMIP or as a UPP to UMIP 10?
https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-10.md

smb2796: It's traditionally a umip if it is a new currency like this. UPPs really only for contract parameter or final fee changes (we really shouldn't have a "UPP" imo since the onchain process is exactly the same and it just introduces confusion!)

Reinis-FRP (approved):

smb2796 (approved):

PR #556, opened 2022-09-16
feat: add target balance configuration

mrice32:
This UMIP should be paired with https://github.com/across-protocol/relayer-v2/pull/301 to add a way to configure spoke target balances in Across v2.

### DIFF
--- a/UMIPs/umip-157.md
+++ b/UMIPs/umip-157.md
@@ -95,19 +95,25 @@ The following constants are currently specified in this UMIP directly, but shoul
 ## Token Constants
 The following constants are also stored in the `AcrossConfigStore` contract but are specific to an Ethereum token address. Therefore, they are fetched by querying the config store's `tokenConfig(address)` function.
 - "rateModel"
+  - This is a JSON struct of rate model parameters.
+  - These parameters should fully specify the rate model for this token as described in
+    [UMIP 136](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-136.md).
+  - Each field in this struct should be an integer represented as a string (to allow unlimited precision).
+  - The rateModel struct is only valid if it contains all of the following parameters: `UBar`, `R0`, `R1`, and `R2`.
 - "spokeTargetBalances"
   - This is contains a JSON mapping from chainId to JSON structs.
   - Each struct contains two sub-fields, "target" and "threshold".
-  - Each is a string containing an integer.
+  - Each is an integer represented as a string (to allow unlimited precision).
   - These integers should both be positive values. If either is negative, it should be treated as "0" when used.
   - The "target" integer should be < the "threshold" integer. If not, the "theshold" integer should be treated as if
     it contained the same value as the "target" integer when used.
   - If "spokeTargetBalances", a particular chainId is missing from the mapping, or "target" or "threshold" is missing,
     the missing "target" and "threshold" should be defaulted to 0.
 - "transferThreshold"
+  - This field should be a single integer value represented as a string (to allow unlimited precision).
 
 For example, querying `tokenConfig("0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2")` might return:
-> "{"rateModel":{"UBar":"750000000000000000","R0":"50000000000000000","R1":"0","R2":"600000000000000000"},"transferThreshold":"100000000000000000"}"
+> "{"rateModel":{"UBar":"750000000000000000","R0":"50000000000000000","R1":"0","R2":"600000000000000000"},"transferThreshold":"100000000000000000","spokeTargetBalances":{"1":{"threshold":"200000000000000000000","target":"100000000000000000000"},"42161":{"threshold":"400000000000000000000","target":"200000000000000000000"}}}"
 
 _This UMIP will explain later how global and token-specific configuration settings are used._
 
--- a/UMIPs/umip-157.md
+++ b/UMIPs/umip-157.md
@@ -272,20 +272,23 @@ For each `repaymentChainId` and `l1Token` combination, older
 [RootBundleExecuted](https://github.com/across-protocol/contracts-v2/blob/a8ab11fef3d15604c46bba6439291432db17e745/contracts/HubPool.sol#L157-L166) events
 should be queried to find the preceding `RootBundleExecuted` event. This means identifying the most recent `RootBundleExecuted` event with a `chainId` matching the `repaymentChainId` and [identifying the `runningBalanceValue` at the index of the `l1Token`](#matching-l1-tokens-to-running-balances-or-net-send-amounts).
 
-For each tuple of `l1Token` and `repaymentChainId`, we should have computed a total running balance value. There are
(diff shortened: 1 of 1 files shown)


UMIPs/umip-157.md
```diff
@@ -95,6 +95,15 @@ The following constants are currently specified in this UMIP directly, but shoul
 ## Token Constants
 The following constants are also stored in the `AcrossConfigStore` contract but are specific to an Ethereum token address. Therefore, they are fetched by querying the config store's `tokenConfig(address)` function.
 - "rateModel"
+- "spokeTargetBalances"
+  - This is contains a JSON mapping from chainId to JSON structs.
```
nicholaspai, line 99: nit: the description of the constant name should probably follow the same style as for `transferThreshold` which doesn't seem to use the "-" bullet points. Can you make these styles consistent in this PR? Whether that means changing `transferThreshold` description to match this new style or vice versa I don't mind

UMIPs/umip-157.md
```diff
@@ -95,6 +95,15 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+  - Each struct contains two sub-fields, "target" and "threshold".
+  - Each is a string containing an integer.
+  - These integers should both be positive values. If either is negative, it should be treated as "0" when used.
+  - The "target" integer should be < the "threshold" integer. If not, the "theshold" integer should be treated as if
+    it contained the same value as the "target" integer when used.
+  - If "spokeTargetBalances", a particular chainId is missing from the mapping, or "target" or "threshold" is missing,
+    the missing "target" and "threshold" should be defaulted to 0.
```
nicholaspai, line 106: Can you add an example return value like we do in `transferThreshold`?

UMIPs/umip-157.md
nicholaspai, line 99: This point could also apply to `rateModel`

UMIPs/umip-157.md
```diff
@@ -257,11 +266,21 @@ For each `repaymentChainId` and `l1Token` combination, older
...
-If this passes the threshold, then set net send amount to the running balance value and set the running balance value
-to 0.
+For each tuple of `l1Token` and `repaymentChainId`, we should have computed a total running balance value. There are
+two cases to consider.
+
+The first is when the running balance value for this token is positive or 0. In that case, we follow the following
+logic: if the running balance is > the [`transferThreshold`](#token-constants) for this token, set net send amount to
+the running balance and set the running balance to 0.
+
+The other case is when the running balance is negative. In that case, the logic is a bit more complex:
```
nicholaspai, line 276:
```suggestion
The other case is when the running balance is negative. In that case:
```

UMIPs/umip-157.md
```diff
@@ -257,11 +266,21 @@ For each `repaymentChainId` and `l1Token` combination, older
...
+- If the running balance value for this token is < the "threshold" value in [`spokeTargetBalances`](#token-constants)
```
nicholaspai, line 277:
```suggestion
- If the absolute value of the running balance value for this token is < the "threshold" value in [`spokeTargetBalances`](#token-constants)
```

UMIPs/umip-157.md
```diff
@@ -257,11 +266,21 @@ For each `repaymentChainId` and `l1Token` combination, older
...
+  for this token is < the absolute value of this running balance, leave the running balance set to its current value
```
nicholaspai, line 278:
```suggestion
  for this token, leave the running balance set to its current value
```

nicholaspai (commented):
Agree with this structure, I think we can tighten up the diff though!

UMIPs/umip-157.md
```diff
@@ -257,11 +266,21 @@ For each `repaymentChainId` and `l1Token` combination, older
...
+  and set the net send amount for this token to 0.
+- Otherwise, subtract the absolute value of the running balance from the "target" value in
+  [`spokeTargetBalances`](#token-constants). If the result is positive (this should not happen under correct
+  configurations), set it to 0.
+- Set net send amount to that value and subtract that value from the running balance.
```
nicholaspai, line 283: This line is confusing. It kind of implies that runningBalance and netSendAmount can both be non zero

mrice32: As discussed offline, they can both be nonzero with this change.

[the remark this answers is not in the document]
mrice32: This is the most concise math/algo I could come up with:
```
transfer_threshold = the transfer threshold for this token
spoke_balance_threshold = the "threshold" value in `spokeTargetBalances` for this token
spoke_balance_target = the "threshold" value in `spokeTargetBalances` for this token

net_send_amount = 0
if running_balance >= 0 && running_balance >= transfer_threshold:
  net_send_amount = running_balance
  running_balance = 0
else if abs(running_balance) >= spoke_balance_threshold:
  desired_transfer_amount = min(running_balance + spoke_balance_target, 0)
  net_send_amount = abs(desired_transfer_amount) >= transfer_threshold ? desired_transfer_amount : 0
  running_balance = running_balance - net_send_amount
```

Thoughts? Do you have a simpler explanation?

UMIPs/umip-157.md
mrice32: Done. PTAL.

UMIPs/umip-157.md
mrice32: Added some comments for the other fields. PTAL

mrice32:
@nicholaspai thanks for the detailed comments. I've updated the diff based on your comments and I provided a mathematical algorithm that was as simple as I could come up with [here](https://github.com/UMAprotocol/UMIPs/pull/556#discussion_r976967672). If you think this is simpler, I'm happy to replace the existing explanation with the math/pseudocode.

UMIPs/umip-157.md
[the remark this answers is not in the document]
nicholaspai: Nope that's perfect

nicholaspai: I'm still confused thinking about new edge cases introduced with this change. Does this allow us to carry over runningBalances from one bundle to the following? Does it break other assumptions? 

This might be meaningless ultimately but I'm pretty sure the contracts do have comments to the effect of: either running balances or net send amounts should be 0.

mrice32: We definitely do say this in the comments of the contracts, good point!

https://github.com/across-protocol/contracts-v2/blob/f42e843e351d70692bd8575a44d0f563d1920d8b/contracts/HubPoolInterface.sol#L18-L28

Do you think there's any code that depends on this assumption?

nicholaspai: Nothing on contract side, but dataworker makes some assumptions about how running balance and net send amounts are related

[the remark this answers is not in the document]
mrice32: Done.

nicholaspai (approved):
LGTM, thanks for responding to all rounds of feedback I'm really excited to see how this makes the across system more capital efficient

PR #559, opened 2022-10-28
Add uSPAC10g as price identifier

binomfx:
The difference from the existing UMIP-140 is that the proposed price ID does not require manually changing the basket of 10 SPAC shares on a quarterly basis.
The index basket is formed weekly by requesting top-10 gainers from the spacHero database.
A new API is being used to get stock quotes.


binomfx:
Commits from 03.11.2022 change the API for getting stock quotes. The previous API on rapidapi has been moved to the private category and is only available to previously registered users.
Apparently, this is due to the fact that "Yahoo finance" was used in the name of the previous API.

UMIPs/UMIP-uSPAC10g.md
```diff
@@ -0,0 +1,443 @@
...
+
+## IMPLEMENTATION
+### Price Identifier
+In order to determine the index value, the following steps are required:
+
+![uSPAC10g-IdxCalc](https://user-images.githubusercontent.com/25432493/198379446-af065a62-88a7-4c94-9315-f6b98b1123ca.jpg)
+
+#### 1. Obtain Index Basket
+The index basket is formed **weekly** by requesting top-10 gainers from the spacHero database, available from "spacHero – Rapidapi.com" (API)<br> 
+> In order to index can reliably reflect the market picture, a periodic change of the basket of stocks included in the index is required. Therefore, we request a new price identifier weekly. <br>
```
abg4, line 65: Can you explain what you mean by `we request a new price identifier weekly`? Do you just mean the basket is changing?

Also, is there a specific timestamp the change of the basket stocks happens each week?

binomfx: Indeed. It means that we change the index basket weekly. There is no specific timestamp for changing the index basket, but I agree that it is necessary.
I will make the appropriate corrections to UMIP.

UMIPs/UMIP-uSPAC10g.md
```diff
@@ -0,0 +1,443 @@
...
+|:-------------|:-----------------------------------------------------------------------------|
+|UMIP title|Add uSPAC10g as price identifier|
+|Author|BinomFX (binomfx@gmail.com)|
+|Status|Draft| 
+|Created|27.10.2022|
+|Discourse Link|https://discourse.umaproject.org/t/add-uspac10g-as-price-identifier/1835|
+
+## SUMMARY
+The DVM should support price requests for uSPAC10g price identifier<br>
+The purpose of this price identifier is to create synthetic token, price of which is linked to the value of index of **10** most active SPACs (Special Purpose Acquisition Companies) shares.<br> That synthetic token can be used for creating speculative strategies at IPO market.<br>
```
pemulis, line 12: `The purpose of this price identifier is to create a synthetic token, the price of which is linked to the value of an index of the **10** most active SPACs (Special Purpose Acquisition Companies) during a given time period.<br> That synthetic token can be used for creating speculative strategies in SPAC market.<br>`

UMIPs/UMIP-uSPAC10g.md
pemulis: If the markets are totally closed on a Friday because of a holiday, would the basket be updated on Thursday after markets close, and then go back to normal the next week (closing after markets close on Friday)?

pemulis:
Out of curiosity, why was the API source changed for quotes? 
image

pemulis:
Why is the correction factor removed in the new UMIP? 
image

pemulis:
Have you tested that the price feed implementation is working as expected with the new API endpoint?

UMIPs/UMIP-uSPAC10g.md
binomfx: Yes, of course.  We get the index basket via the spacHero API every time the uSPAC10g price is requested. The index bucket changes automatically.

UMIPs/UMIP-uSPAC10g.md
binomfx: I agree. Changed it in UMIP.

UMIPs/UMIP-uSPAC10g.md
```diff
@@ -0,0 +1,443 @@
...
+The difference from the existing UMIP-140 is that the proposed price ID does not require manually changing the basket of 10 SPAC shares on a quarterly basis.
+
+## MOTIVATION
+A synthetic token that tracks the index of the 10 most active SPACs stocks can be used for speculative purposes and allows the user to earn on price movements in one of the most interesting markets without centralized intermediaries such as exchanges and brokers.<br>
```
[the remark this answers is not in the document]
binomfx: I agree. Changed it in UMIP.

UMIPs/UMIP-uSPAC10g.md
```diff
@@ -0,0 +1,443 @@
...
+|Discourse Link|https://discourse.umaproject.org/t/add-uspac10g-as-price-identifier/1835|
+
+## SUMMARY
+The DVM should support price requests for uSPAC10g price identifier<br>
+The purpose of this price identifier is to create synthetic token, price of which is linked to the value of index of **10** most active SPACs (Special Purpose Acquisition Companies) shares.<br> That synthetic token can be used for creating speculative strategies at IPO market.<br>
+The difference from the existing UMIP-140 is that the proposed price ID does not require manually changing the basket of 10 SPAC shares on a quarterly basis.
+
+## MOTIVATION
+A synthetic token that tracks the index of the 10 most active SPACs stocks can be used for speculative purposes and allows the user to earn on price movements in one of the most interesting markets without centralized intermediaries such as exchanges and brokers.<br> 
+In addition, that token can be used as components associated with classical markets by other DeFi and DApp protocols, which makes it possible to scale.
```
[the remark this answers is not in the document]
binomfx: I agree. Changed it in UMIP.

binomfx:
The correction factor made sense with a fairly rare change in the index basket (as it was earlier - quarterly). If the index basket changes relatively frequently (for example, weekly), then the value of the correction coefficient tends to 1. At the same time, its use will unnecessarily complicate the calculation of the index value.

binomfx:
> Have you tested that the price feed implementation is working as expected with the new API endpoint?

Not yet. Our programmer is doing this.

pemulis:
Thanks for the clarifications @binomfx! Interested to hear what your programmer finds in the price feed implementation testing.

One other concern I had was whether the design of this identifier would create waves of liquidations on a weekly basis, since my understanding is that every week there would be a rebalancing to the top-10 gainers from the spacHero database. Is that the top ten gainers in terms of volume or value? If you are always rebalancing to the highest value basket, especially for a very volatile category like SPACs, wouldn't you frequently create a wave of liquidations as synth holders are suddenly undercollateralized every week?

And that's the best case, the worse case (which also seems plausible?) is that the synth itself becomes undercollateralized, due to sharp price movements and index changes.

As you test the price feed, I would be interested in seeing how the identifier's rules would have behaved historically, and what would have been a "safe" collateral factor, for individual minters and in aggregate for the synthetic asset itself, to keep it from going underwater.

I'm also wondering how the synth would perform if you changed the update cadence to hourly or weekly, or add some other factor, like the increased value of the synth slowly ratchets up over the course of the weekend (when markets are closed) instead of happening all at once, giving people time to increase their collateral positions as necessary.

binomfx:
> pemulis

Thanks for the thoughts @pemulis!
I am sure that in the near future we will all see the results of testing.

As for doubts about the waves of liquidations, we will conduct an additional analysis. Perhaps it really makes sense to use TOP-10 gainers by volume, not by value.

Special thanks for the ideas on how to experiment with the update cadence and other factors.

pemulis:
Hi @binomfx I'm sorry for the delay on our end. We should have reviewed your price feed implementation by now but have been caught up in the Across token launch and some other work. We're going to get this reviewed for you but need to push it into next week. I'm sorry again, this was my fault for not being on top of it.

UMIPs/UMIP-uSPAC10g.md
mhairimcalpine: +1

mhairimcalpine (approved):

PR #560, opened 2022-11-18
feat(umip-157): Allow for admin bundles

nicholaspai:
UMIP should allow for system to be able to fix errors via admin bundles in a way that doesn't give owners too much power. For example, the owner of the ConfigStore could be subject to a timelock when proposing an `ADMIN_BUNDLE` value.

Currently this new feature would give the config store owner power to propose malicious admin bundles but there are some checks on them such as a `proposalTimestamp` that the admin bundle must be proposed by and a `proposer` address restricting who can propose it.

### DIFF
--- a/UMIPs/umip-157.md
+++ b/UMIPs/umip-157.md
@@ -85,6 +85,7 @@ ooRequester:0x69CA24D3084a2eea77E061E2D7aF9b76D107b4f6
 The following constants should reflect what is stored in the [`AcrossConfigStore`](https://etherscan.io/address/0x3b03509645713718b78951126e0a6de6f10043f5#code) contract deployed on Etherscan. This contract is owned by Across governance and acts as the source of truth for the following variables. The global variables currently stored in the above contract for this UMIP are:
 - "MAX_POOL_REBALANCE_LEAF_SIZE"
 - "MAX_RELAYER_REPAYMENT_LEAF_SIZE"
+- "ADMIN_BUNDLE"
 
 To query the value for any of the above constants, the `AcrossConfigStore` contract's `globalConfig(bytes32)` function should be called with the hex value of the variable name. For example, the "MAX_POOL_REBALANCE_LEAF_SIZE" can be queried by calling `globalConfig(toHex("MAX_POOL_REBALANCE_LEAF_SIZE"))` which is equivalent to `globalConfig("0x4d41585f504f4f4c5f524542414c414e43455f4c4541465f53495a45")`. For example, this might return 
 >"25"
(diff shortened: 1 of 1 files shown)


UMIPs/umip-157.md
```diff
@@ -366,4 +367,13 @@ Three conditions must be met for the proposal to be deemed valid:
    at the time of proposal.
 4. No obvious griefing or manipulation of the system is being performed via this proposal.
 
-If the proposal is deemed invalid, return 0. If valid, return 1. Note: these values are scaled by `1e18`.
\ No newline at end of file
+If the proposal is deemed invalid, return 0. If valid, return 1. Note: these values are scaled by `1e18`.
+
+# Admin bundle proposals
+
+The following section describes the normal path for determining a proposed bundle's validity. We have to follow a slightly different path to reconstructing the bundle roots if and only if the `ConfigStore`'s admin bundle setting is set: i.e. `globalConfig(toHex("ADMIN_BUNDLE")) !== undefined`. If not `undefined`, then use the value of this setting to reconstruct roots. The config store's "ADMIN_BUNDLE" value should have three keys: `poolRebalanceLeaves`, `relayerRefundLeaves` and `slowRelayLeaves`.
```
nicholaspai, line 374: One downside to this approach is that the admin needs to remember to delete this ADMIN_BUNDLE key from the ConfigStore.globalConfig otherwise the admin can send a duplicate admin bundle. Some ways we could combat that are to include a `proposer` (the account that will propose the admin bundle) and `nonce` (the nonce that the account will use to propose the admin bundle).

Then, the next time that an admin bundle needs to be proposed, the admin can overwrite the entry in the globalConfig. There will never need to be two ADMIN_BUNDLE entries simultaneously.

Practically, to save the disputers computation, the admin should delete the ADMIN_BUNDLE entry as soon as the admin bundle is executed.

UMIPs/umip-157.md
```diff
@@ -366,4 +367,13 @@ Three conditions must be met for the proposal to be deemed valid:
...
+The following section describes the normal path for determining a proposed bundle's validity. We have to follow a slightly different path to reconstructing the bundle roots if and only if the `ConfigStore`'s admin bundle setting is set: i.e. `globalConfig(toHex("ADMIN_BUNDLE")) !== undefined`. If not `undefined`, then use the value of this setting to reconstruct roots. The config store's "ADMIN_BUNDLE" value should have three keys: `poolRebalanceLeaves`, `relayerRefundLeaves` and `slowRelayLeaves`. 
+- `poolRebalanceLeaves`: `{ chainId, bundleLpFees, netSendAmounts, runningBalances, groupIndex, leafId, l1Tokens }`
+- `relayerRefundLeaves`: `{ groupIndex, amountToReturn, chainId, refundAmounts: [], leafId, l2TokenAddress, refundAddresses }`
+- `slowRelayLeaves`: `{ depositor, recipient, destinationToken, amount, originChainId, destinationChainId, realizedLpFeePct, relayerFeePct, depositId }`
+
+The dataworker should convert these values into merkle roots in order to reconstruct a bundle. As long as "ADMIN_BUNDLE" is defined, the dataworker should mark any root bundle containing these exact roots as valid, otherwise the dataworker should follow the normal path to reconstruct roots.
```
mrice32, line 379: Is the idea that the Across Safe would just reset this ADMIN_BUNDLE key anytime it was desirable to push a new ADMIN_BUNDLE through?

Could a malicious actor repeatedly propose this same bundle to get refunded multiple times and the disputer would always ignore it?

mrice32 (commented):
One question about this structure!

UMIPs/umip-157.md
nicholaspai: > Could a malicious actor repeatedly propose this same bundle to get refunded multiple times and the disputer would always ignore it?

Yes, this is the downside. I haven't been able to think of a different way to allow admin bundles through. WDYT?

nicholaspai: One idea is the config store proposer can add a "proposer" key so only the known proposer can send the admin bundle. It does give the config store owner power to now force a bad bundle through

nicholaspai: We could alternatively add an `expirationTime` key to the admin bundle object so that the proposer has until that time to propose (and execute) the bundle or it can be disputed

PR #565, opened 2022-12-22
feat: Allow for truncation of realized LP fee %

nicholaspai:
Reduces risk for relayers to fill relay with trivially different realized lp fee %

### DIFF
--- a/UMIPs/umip-157.md
+++ b/UMIPs/umip-157.md
@@ -213,7 +213,7 @@ Additionally, matching relays should have their `destinationToken` set such that
 
 ## Validating realizedLpFeePct
 To determine the validity of the `realizedLPFeePct` in the `FilledRelay` event, the exact same process is used as in
-the identifier [`IS_RELAY_VALID`, specified in UMIP 136](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-136.md). However, instead ofing a `RateModelStore` contract to look up a deposit's rate model, we can use the `AcrossConfigStore`'s `tokenConfig` to [look up the rate model](#token-constants) for a deposit. The deposited `originToken` can be mapped to an `l1Token` by following step 2 above which can be used to query a `rateModel`. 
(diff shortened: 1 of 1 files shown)


UMIPs/umip-157.md
```diff
@@ -222,6 +222,8 @@ called on the `HubPool` contract, passing in a single argument, the `l1Token` de
 If the `realizedLPFeePct` that is computed using those means does not match the `realizedLPFeePct` in the
 `FilledRelay` event, then the relay is considered invalid.
 
+We acknowledge that this calculation is complex and poses risk to the relayer who would not receive a refund for an honest attempt at a fill where the computed LP fee is off by a trivial amount. Therefore, all `realizedLpFeePct`'s should be floor rounded to 6 decimals. For example, `11000000000000` and `11830749673498` would be considered equivalent fees because `0.000011` and `0.000011830749673498` are both equal to `0.000011` after floor rounding to 6 decimals.
```
mrice32, line 225: How should this be phased in?

Should we increment the config store version to make this change take effect? Otherwise, it will be unclear to relayers when they should start doing this, right?

mrice32 (commented):
Looks awesome! Just a question about phasing it in and ensuring relayers don't get broken by the change.

UMIPs/umip-157.md
pxrl: > How should this be phased in?
> 
> Should we increment the config store version to make this change take effect? Otherwise, it will be unclear to relayers when they should start doing this, right?

My mental modelling of this change was that it should be backwards-compatible, since it permits relayers to be less-precise than they are currently, but doesn't require them to be. Can you see a potential scenario where it backfires?

mrice32: @pxrl it cannot work that way because the fee must match exactly to produce the same hash in the SpokePool. If the hash is different you can get double filled because there could be two distinct, valid fills for the same deposit.

mrice32 (commented):
https://github.com/UMAprotocol/UMIPs/pull/565#discussion_r1085510258

UMIPs/umip-157.md
pxrl: :facepalm: You're completely right, this would open up for double fills.

nicholaspai:
Will be replaced by V3 launch which addresses the same underlying issue

PR #567, opened 2022-12-27
feat: Account for "VERSION" in globalConfig

nicholaspai:
This is a convenience to relayer and dataworker users who can compare their locally stored version against the latest version stored in the `ConfigStore` to know when to safely halt their programs and not send invalid transactions as a result of not using up to date `ConfigStore` client code.


UMIPs/umip-157.md
```diff
@@ -366,4 +368,4 @@ Three conditions must be met for the proposal to be deemed valid:
    at the time of proposal.
 4. No obvious griefing or manipulation of the system is being performed via this proposal.
 
-If the proposal is deemed invalid, return 0. If valid, return 1. Note: these values are scaled by `1e18`.
\ No newline at end of file
+If the proposal is deemed invalid, return 0. If valid, return 1. Note: these values are scaled by `1e18`.
```
pxrl: Is this just a newline change? It's not obvious in the gh diff interface.

UMIPs/umip-157.md
```diff
@@ -85,6 +85,8 @@ ooRequester:0x69CA24D3084a2eea77E061E2D7aF9b76D107b4f6
 The following constants should reflect what is stored in the [`AcrossConfigStore`](https://etherscan.io/address/0x3b03509645713718b78951126e0a6de6f10043f5#code) contract deployed on Etherscan. This contract is owned by Across governance and acts as the source of truth for the following variables. The global variables currently stored in the above contract for this UMIP are:
 - "MAX_POOL_REBALANCE_LEAF_SIZE"
 - "MAX_RELAYER_REPAYMENT_LEAF_SIZE"
+- "VERSION"
```
pxrl, line 88: Should we be explicit about whether this is a config version, or something else? I could imagine that we can tweak the config without bumping the version, but that might be counter-intuitive for anyone trying to understand the behaviour of this variable.

UMIPs/umip-157.md
```diff
@@ -85,6 +85,8 @@ ooRequester:0x69CA24D3084a2eea77E061E2D7aF9b76D107b4f6
...
+  - This is a convenience to relayer and dataworker users who can compare their locally stored version against the latest version stored in the `ConfigStore` to know when to safely halt their programs and not send invalid transactions as a result of not using up to date `ConfigStore` client code.
```
pxrl, line 89: Just throwing some ideas around - wdyt?

```suggestion
  - Across protocol version number. Supporting implementations should query this version against a locally stored value to determine compatibility with the current protocol implementation. Failure to correctly evaluate the version number may mean that filled relays are not refunded from the HubPool, and may therefore result in loss of relayer funds.
```

pxrl:
General comment: GR also observed that there is a version string embedded in the approval for deposit speed-ups. Should we consider integrating them in some way? If not, we may want to consider some language to indicate that they are not directly related.

UMIPs/umip-157.md
nicholaspai: Why would we ever want to modify the config without tweaking the version?

UMIPs/umip-157.md
```diff
@@ -85,6 +85,8 @@ ooRequester:0x69CA24D3084a2eea77E061E2D7aF9b76D107b4f6
 The following constants should reflect what is stored in the [`AcrossConfigStore`](https://etherscan.io/address/0x3b03509645713718b78951126e0a6de6f10043f5#code) contract deployed on Etherscan. This contract is owned by Across governance and acts as the source of truth for the following variables. The global variables currently stored in the above contract for this UMIP are:
 - "MAX_POOL_REBALANCE_LEAF_SIZE"
 - "MAX_RELAYER_REPAYMENT_LEAF_SIZE"
+- "VERSION"
+  - Across protocol version number. Supporting implementations should query this value against a locally stored value to determine compatibility with the current protocol implementation. Failure to correctly evaluate the version number may mean that filled relays are not refunded from the HubPool, and may therefore result in loss of relayer funds.
```
mrice32, line 89:
```suggestion
  - Across protocol version number. Supporting implementations should query this value against the value defined in their implementation to determine compatibility with the current protocol version. Failure to correctly evaluate the version number may mean that filled relays are not refunded from the HubPool, and may therefore result in loss of relayer funds.
```

mrice32 (commented):
Similar to other config variables, I assume bots should check the version string on a per-relay basis. For instance, consider the following sequence of actions:
- VERSION string is initially set to "1.0.0" at time 10.
- Deposit 1 is sent with a quote timestamp of 15.
- VERSION string is changed to "2.0.0" at time 20.
- Deposit 2 is sent with a quote timestamp of 25.

I assume a relayer with version string "1.0.0" would be willing to relay Deposit 1, but not Deposit 2.

Similarly, a relayer with version string "2.0.0" would be willing to relay Deposit 2, but not Deposit 1.

And then maybe you could also have implementations that support multiple version strings simultaneously. For instance, my relayer could, for instance, support 1.0.0 or 2.0.0 and be willing to relay both deposits.

If you agree with this, I think we should just make sure that this clearly states that this works the same way as the other parameters:
- Proposers/Disputers must be able to support the version string of the entire range of a bundle.
- Relayers must support the version string for the quote timestamp for the deposit they are relaying.

Do you agree?

pxrl:
> Similar to other config variables, I assume bots should check the version string on a per-relay basis. For instance, consider the following sequence of actions:
> 
> * VERSION string is initially set to "1.0.0" at time 10.
> * Deposit 1 is sent with a quote timestamp of 15.
> * VERSION string is changed to "2.0.0" at time 20.
> * Deposit 2 is sent with a quote timestamp of 25.
> 
> I assume a relayer with version string "1.0.0" would be willing to relay Deposit 1, but not Deposit 2.
> 
> Similarly, a relayer with version string "2.0.0" would be willing to relay Deposit 2, but not Deposit 1.
> 
> And then maybe you could also have implementations that support multiple version strings simultaneously. For instance, my relayer could, for instance, support 1.0.0 or 2.0.0 and be willing to relay both deposits.

Thinking about a possible implementation of this, would it imply that there is a version dictionary that maps version numbers onto mainnet block numbers? This would allow it to be linked to LP utilisation. Would that also imply that we need to version other aspects of the configuration, so that we can reconcile (for example) changes in fee models over time?

One possible advantage of this it could make it easier to push new updates, because we could specify that a new version takes effect from a future block number, thereby allowing time for relay operators to upgrade. I like this, because I've been thinking about how we would practically coordinate pushing a new version simultaneously with rolling out bot updates, and it seems like there might be some dragons on that quest.

mrice32:
> > Similar to other config variables, I assume bots should check the version string on a per-relay basis. For instance, consider the following sequence of actions:
> 
> > 
> 
> > * VERSION string is initially set to "1.0.0" at time 10.
> 
> > * Deposit 1 is sent with a quote timestamp of 15.
> 
> > * VERSION string is changed to "2.0.0" at time 20.
> 
> > * Deposit 2 is sent with a quote timestamp of 25.
> 
> > 
> 
> > I assume a relayer with version string "1.0.0" would be willing to relay Deposit 1, but not Deposit 2.
> 
> > 
> 
> > Similarly, a relayer with version string "2.0.0" would be willing to relay Deposit 2, but not Deposit 1.
> 
> > 
> 
> > And then maybe you could also have implementations that support multiple version strings simultaneously. For instance, my relayer could, for instance, support 1.0.0 or 2.0.0 and be willing to relay both deposits.
> 
> 
> 
> Thinking about a possible implementation of this, would it imply that there is a version dictionary that maps version numbers onto mainnet block numbers? This would allow it to be linked to LP utilisation. If so, would that also imply that we need to maintain versioning on all other aspects of the configuration, so that we can reconcile (for example) changes in fee model parameters over time?
> 
> 
> 
> One possible advantage of this it could make it easier to push new updates, because we could specify that a new version takes effect from a future block number, thereby allowing time for relay operators to upgrade. I like this, because I've been thinking about how we would practically coordinate pushing a new version simultaneously with rolling out bot updates, and it seems like there might be some dragons on that quest.

@pxrl today, we have some sort of time-based versioning for rate models. The rate model config that applies at mainnet block X is whatever is in the ConfigStore contract at block X. Similarly for pool rebalance thresholds, etc. Note: this type of versioning does not work for the setup you're suggesting where config changes are preloaded on chain to take effect at some future block.

nicholaspai:
> General comment: GR also observed that there is a version string embedded in the approval for deposit speed-ups. Should we consider integrating them in some way? If not, we may want to consider some language to indicate that they are not directly related.

This is a great reminder, I do not think we should integrate with that as that is a SpokePool version being referenced. I also don't think the need for this versioning is neccessary if we upgrade the spoke pools and increment the deposit ID's correctly so that no deposit ID's repeat. This way speed up signatures can't be reused

nicholaspai:
> Similar to other config variables, I assume bots should check the version string on a per-relay basis. For instance, consider the following sequence of actions:
> 
> * VERSION string is initially set to "1.0.0" at time 10.
> * Deposit 1 is sent with a quote timestamp of 15.
> * VERSION string is changed to "2.0.0" at time 20.
> * Deposit 2 is sent with a quote timestamp of 25.
> 
> I assume a relayer with version string "1.0.0" would be willing to relay Deposit 1, but not Deposit 2.
> 
> Similarly, a relayer with version string "2.0.0" would be willing to relay Deposit 2, but not Deposit 1.
> 
> And then maybe you could also have implementations that support multiple version strings simultaneously. For instance, my relayer could, for instance, support 1.0.0 or 2.0.0 and be willing to relay both deposits.
> 
> If you agree with this, I think we should just make sure that this clearly states that this works the same way as the other parameters:
> 
> * Proposers/Disputers must be able to support the version string of the entire range of a bundle.
> * Relayers must support the version string for the quote timestamp for the deposit they are relaying.
> 
> Do you agree?

Agreed and I've updated the language a bit

UMIPs/umip-157.md
mrice32: I want to make sure we're talking about the same thing:
- There's changing the values in the config (i.e. tweaking the config).
- There's changing the _structure_ of the config or how the config is _interpreted_. I would think of this as being more what the version signifies. Do y'all agree? Or are y'all talking about something else?

UMIPs/umip-157.md
```diff
@@ -85,6 +88,8 @@ ooRequester:0x69CA24D3084a2eea77E061E2D7aF9b76D107b4f6
 The following constants should reflect what is stored in the [`AcrossConfigStore`](https://etherscan.io/address/0x3b03509645713718b78951126e0a6de6f10043f5#code) contract deployed on Etherscan. This contract is owned by Across governance and acts as the source of truth for the following variables. The global variables currently stored in the above contract for this UMIP are:
 - "MAX_POOL_REBALANCE_LEAF_SIZE"
 - "MAX_RELAYER_REPAYMENT_LEAF_SIZE"
+- "VERSION"
+  - Across protocol version number. Supporting implementations should query this value against the value defined in their implementation to determine compatibility with the current protocol version. Failure to correctly evaluate the version number may mean that filled relays are not refunded from the HubPool, and may therefore result in loss of relayer funds. For more information go [here](#versions).
```
mrice32, line 92: I think we should make sure this also applies to disputer/proposer implementations.
```suggestion
  - Across protocol version number. Supporting implementations should query this value against the value defined in their implementation to determine compatibility with the current protocol version. Failure to correctly evaluate the version number may mean that filled relays are not refunded from the HubPool, and may therefore result in loss of funds. For more information go [here](#versions).
```

UMIPs/umip-157.md
```diff
@@ -67,6 +67,9 @@ The [`RootBundleExecuted`](https://github.com/across-protocol/contracts-v2/blob/
 
 For example, if `l1Tokens` is "[0x123,0x456,0x789]" and `netSendAmounts` is "[1,2,3]", then the "net send amount" for token with address "0x456" is equal to "2".
 
+## Versions
+The `ConfigStore` stores a "VERSION" value in the `globalConfig`. This is used to protect relayers and dataworkers from using outdated code when interacting with Across. The "VERSION" should be mapped to an integer string that can only increase over time. The "VERSION" is updated by calling `updateGlobalConfig` so it is emitted as an on-chain event. The block time of the event that included the "VERSION" indicates when that "VERSION" became active. Relayers should support the version at the quote timestamp for a deposit, otherwise they risk sending an invalid fill. Proposers and Disputers should support the latest version for a bundle block range to validate or propose a new bundle. This version is independent from the version included in [speed up deposit signatures](https://github.com/across-protocol/contracts-v2/blob/a8ab11fef3d15604c46bba6439291432db17e745/contracts/SpokePool.sol#L668).
```
mrice32, line 71: 👌  great explanation @nicholaspai!

mrice32 (approved):
LGTM!

nicholaspai:
@mrice32 I agree the version should only be bumped when the structure of the config changes

> the structure of the config or how the config is interpreted.

mrice32 (approved):

PR #568, opened 2022-12-30
feat: Comment on provider quorum

nicholaspai:


UMIPs/umip-157.md
```diff
@@ -46,6 +46,8 @@ sort should be on the `transactionIndex`, and the tertiary sort should be on the
 
 Note 3: wherever unspecified, sorting should be ascending by default, not descending.
 
+Note 4: all event data should be identically returned by at least two RPC providers to give confidence in the integrity of the data.
```
pxrl, line 49: Minor tweaks, but I think the gist of this change is good.

```suggestion
Note 4: all event data should be identically returned by at least two independent, reputable RPC providers to give confidence in the integrity of the data.
```

mrice32: Why not just "multiple" instead of 2?

mrice32 (dismissed):

pxrl (approved):

PR #576, opened 2023-02-21
add rules about disabled chains

nicholaspai:
There need to be separate rules for determining bundle end blocks for chains that are disabled. We can't simply remove a chain from the list because its `index` in CHAIN_ID_LIST needs to be permanently the same while Across is running


UMIPs/umip-157.md
```diff
@@ -98,6 +98,7 @@ To query the value for any of the above constants, the `AcrossConfigStore` contr
 
 The following constants are currently specified in this UMIP directly, but should be moved to the `AcrossConfigStore` in the future. Once that happens, this UMIP can be amended to move the following constants in to the above section.
 - "CHAIN_ID_LIST"=[1,10,137,288,42161] # Mainnet, Optimism, Polygon, Boba, Arbitrum
+- "DISABLED_CHAIN_ID_LIST"=[]
```
mrice32, line 101: Two questions:
- Should these be moved to the config store?
- How does the dataworker know what block number/timestamp a chain goes from being enabled to disabled?

mrice32 (commented):
Looks great! One comment

UMIPs/umip-157.md
nicholaspai: hmm thats a good point. I guess if we move it to the config store then the dataworker would know

nicholaspai: What do you think about this change to the rules? https://github.com/UMAprotocol/UMIPs/pull/576/commits/72cb6ce24d5129291cad42dfdab115d053afd287

UMIPs/umip-157.md
```diff
@@ -167,6 +170,8 @@ Use this mechanism to determine the starting block numbers for each `chainId` re
 
 Note that the above rules require that the `bundleEvaluationBlockNumbers` for each `chainId` are strictly greater than the preceding [valid proposal's](#valid-bundle-proposals) `bundleEvaluationBlockNumbers` for the same `chainId`. The block range for each proposal starts at the preceding proposal's `bundleEvaluationBlockNumbers` plus 1 and go to the next `bundleEvaluationBlockNumbers`.
 
+Note that the above rules for determining an end block don't apply if the chain ID is in the "DISABLED_CHAINS" list. If a chain ID is in the DISABLED_CHAINS list then we need to find the last known bundle end block for the chain ID by finding the most recent validated bundle that was proposed before the chain ID was added to the DISABLED_CHAINS list. The bundle end block for the (now) disabled chain should be used as the current bundle's end block for this chain. This bundle end block should keep being re-used until the chain is removedfrom the DISABLED_CHAINS list.
```
mrice32, line 173: Let's say the block number has been incremented for chain A in the current bundle, but it is _currently_ in the DISABLED_CHAINS list. How should a disputer know whether that bundle _validly_ incremented the block number or whether that bundle should be disputed? Put differently, when does a config change apply to a bundle? Do we use the end block for mainnet? The start block for mainnet?

mrice32 (commented):
Looks great! One question

UMIPs/umip-157.md
nicholaspai: This is a good clarification to make, does this answer your question? https://github.com/UMAprotocol/UMIPs/pull/576/commits/abce5349f7ca1b803cf98ce11c47862c16f06fe9

"for any bundles where the bundle end block for chain ID 1 is >= the block at which the disabled chain was added to the DISABLED_CHAINS list, the disabled chain's end block should be re-used in this way"

UMIPs/umip-157.md
```diff
@@ -167,6 +170,8 @@ Use this mechanism to determine the starting block numbers for each `chainId` re
 
 Note that the above rules require that the `bundleEvaluationBlockNumbers` for each `chainId` are strictly greater than the preceding [valid proposal's](#valid-bundle-proposals) `bundleEvaluationBlockNumbers` for the same `chainId`. The block range for each proposal starts at the preceding proposal's `bundleEvaluationBlockNumbers` plus 1 and go to the next `bundleEvaluationBlockNumbers`.
 
+Note that the above rules for determining an end block don't apply if the chain ID is in the "DISABLED_CHAINS" list. If a chain ID is in the DISABLED_CHAINS list then we need to find the last known bundle end block for the chain ID by finding the most recent validated bundle that was proposed before the chain ID was added to the DISABLED_CHAINS list. The bundle end block for the (now) disabled chain should be used as the current bundle's end block for this chain. This bundle end block should keep being re-used until the chain is removedfrom the DISABLED_CHAINS list. For proposal validators, for any bundles where the bundle end block for chain ID 1 is >= the block at which the disabled chain was added to the DISABLED_CHAINS list, the disabled chain's end block should be re-used in this way.
```
pxrl, line 173: I think it's preferable to separate the `what` from the `how`. In this case, the critical information is that if a chain exists in DISABLED_CHAINS, all future proposals must reuse the bundle end block from the last proposal before it was added. Details on how to identify the last bundle end block are (IMO) implementations decisions that could be left up to the implementer.

I've read this sentence a few times and am not sure that I understand it:

```
For proposal validators, for any bundles where the bundle end block for chain ID 1 is >= the block at which the disabled chain was added to the DISABLED_CHAINS list, the disabled chain's end block should be re-used in this way.
```

pxrl:
Hypothetical question: If a chain is ever re-enabled after having been on the disabled list, does that mean the first bundle would need to span the entire set of blocks from when it was disabled, until now?

pxrl:
Additional consideration: if a chain is disabled because it's no longer available (i.e. it disappears for some reason), would this scheme still be robust to that?

UMIPs/umip-157.md
nicholaspai: "if a chain exists in DISABLED_CHAINS, all future proposals must reuse the bundle end block from the last proposal before it was added" this is well said i'm going to use this

nicholaspai:
> Hypothetical question: If a chain is ever re-enabled after having been on the disabled list, does that mean the first bundle would need to span the entire set of blocks from when it was disabled, until now?

yep

nicholaspai:
> Additional consideration: if a chain is disabled because it's no longer available (i.e. it disappears for some reason), would this scheme still be robust to that?

I think so, if you keep it in the disabled list, right? This was my intended use case

UMIPs/umip-157.md
```diff
@@ -92,13 +92,16 @@ The following constants should reflect what is stored in the [`AcrossConfigStore
 - "MAX_RELAYER_REPAYMENT_LEAF_SIZE"
 - "VERSION"
   - Across protocol version number. Supporting implementations should query this value against the value defined in their implementation to determine compatibility with the current protocol version. Failure to correctly evaluate the version number may mean that filled relays are not refunded from the HubPool, and may therefore result in loss of funds. For more information go [here](#versions).
+- "DISABLED_CHAINS"
+  - This should be a stringified list of chain ID numbers. This cannot contain the chain ID "1".
```
james-a-morris, line 96: Should we include an additional constant called `VITAL_CHAINS` which is a list of stringified chain ID numbers that cannot be disabled and are integral to the operation of the Across protocol? We could initially set this to be `[ 1 ]`.

My reason for suggesting this is that I'm unsure if chain `1` will be the only chain we'll ever consider vital enough to not be able to be disabled

UMIPs/umip-157.md
```diff
@@ -92,13 +92,16 @@ The following constants should reflect what is stored in the [`AcrossConfigStore
...
 
 To query the value for any of the above constants, the `AcrossConfigStore` contract's `globalConfig(bytes32)` function should be called with the hex value of the variable name. For example, the "MAX_POOL_REBALANCE_LEAF_SIZE" can be queried by calling `globalConfig(toHex("MAX_POOL_REBALANCE_LEAF_SIZE"))` which is equivalent to `globalConfig("0x4d41585f504f4f4c5f524542414c414e43455f4c4541465f53495a45")`. For example, this might return 
 >"25"
 
 The following constants are currently specified in this UMIP directly, but should be moved to the `AcrossConfigStore` in the future. Once that happens, this UMIP can be amended to move the following constants in to the above section.
 - "CHAIN_ID_LIST"=[1,10,137,288,42161] # Mainnet, Optimism, Polygon, Boba, Arbitrum
 
+
```
james-a-morris, line 104: Nit: do we need this space?

james-a-morris (commented):
Left a comment regarding the exclusion of `chainId=1` from the `DISABLED_CHAINS` list

UMIPs/umip-157.md
```diff
@@ -170,7 +170,7 @@ Use this mechanism to determine the starting block numbers for each `chainId` re
...
 
-Note that the above rules for determining an end block don't apply if the chain ID is in the "DISABLED_CHAINS" list. If a chain ID is in the DISABLED_CHAINS list then we need to find the last known bundle end block for the chain ID by finding the most recent validated bundle that was proposed before the chain ID was added to the DISABLED_CHAINS list. The bundle end block for the (now) disabled chain should be used as the current bundle's end block for this chain. This bundle end block should keep being re-used until the chain is removedfrom the DISABLED_CHAINS list. For proposal validators, for any bundles where the bundle end block for chain ID 1 is >= the block at which the disabled chain was added to the DISABLED_CHAINS list, the disabled chain's end block should be re-used in this way.
+Note that the above rules for determining an end block don't apply if the chain ID is in the "DISABLED_CHAINS" list. if a chain exists in DISABLED_CHAINS, all future proposals must reuse the bundle end block from the last proposal before it was added.
```
mrice32, line 173: Since DISABLED_CHAINS is a dynamic field and can be changed as often as needed, I think it's a little misleading to say "all future proposals".

Assuming the mainnet end block is what we use to define the configuration, I was thinking something like this:
"Note that the above rules for determining an end block don't apply if the chain ID is in the "DISABLED_CHAINS" list. Specifically, if a chain exists in DISABLED_CHAINS at the mainnet end block (chain id 1) for a particular proposal, the end block for that chain should be identical to its value in the executed bundle immediately before the one in question."

Thoughts? Can you think of a way to simplify that language (it's a bit dense).

mrice32 (commented):
One more comment on timing of config changes!

UMIPs/umip-157.md
nicholaspai: Added this language quit a bit PTAL!

UMIPs/umip-157.md
james-a-morris: Fair enough! As long as it the vital chain will only be "1", this sounds good to me 😄

james-a-morris (dismissed):
I feel confident that my questions have been answered. Approving for now, but with resolution pending on @mrice32 's comment

UMIPs/umip-157.md
```diff
@@ -167,6 +170,8 @@ Use this mechanism to determine the starting block numbers for each `chainId` re
 
 Note that the above rules require that the `bundleEvaluationBlockNumbers` for each `chainId` are strictly greater than the preceding [valid proposal's](#valid-bundle-proposals) `bundleEvaluationBlockNumbers` for the same `chainId`. The block range for each proposal starts at the preceding proposal's `bundleEvaluationBlockNumbers` plus 1 and go to the next `bundleEvaluationBlockNumbers`.
 
+Note also that the above rules for determining an end block don't apply if the chain ID is in the "DISABLED_CHAINS" list. if a chain exists in DISABLED_CHAINS, the proposed bundle must reuse the bundle end block from the last valid proposal before it was added. Specifically, if a chain exists in DISABLED_CHAINS at the "mainnet" end block (chain ID 1) for a particular proposal, the end block for that chain should be identical to its value in the latest executed bundle.
```
nicholaspai, line 173: One potential edge case. How would we handle the case where the disabled chain was added to DISABLED_CHAINS on-chain in the middle of a bundle block range? Currently the rules suggest that the disabled chain would be basically ignored for this bundle. I think that's ok. So its up to the DAO to make sure that the disabled chain is only added to DISABLED_CHAINS once all deposits are disabled to/from the disabled chain

pxrl: I was also wondering about the sequencing, and concluded that we'd need to disable the SpokePool in the bundle _before_ it gets added to the list of DISABLED_CHAINS. That seems like the only way to be sure that it won't have transactions included in the following bundles.

It does however make it a little bit awkward if there are pending deposits in SpokePools on other chains. We can minimise the chance of that by disabling it in the API, so that nobody can obtain quotes for that route. An alternative or complementary strategy could be to jack the LP fee sky high for all routes ending on that destination chain, but that seems like quite a bit of extra work and brings its own set of risks.

Another thing to consider is whether there might be historical deposits in the SpokePool that is about to be disabled. Since it's not possible for withdrawals, those funds become stuck indefinitely.

UMIPs/umip-157.md
```diff
@@ -92,6 +92,8 @@ The following constants should reflect what is stored in the [`AcrossConfigStore
 - "MAX_RELAYER_REPAYMENT_LEAF_SIZE"
 - "VERSION"
   - Across protocol version number. Supporting implementations should query this value against the value defined in their implementation to determine compatibility with the current protocol version. Failure to correctly evaluate the version number may mean that filled relays are not refunded from the HubPool, and may therefore result in loss of funds. For more information go [here](#versions).
+- "DISABLED_CHAINS"
+  - This should be a stringified list of chain ID numbers. This cannot contain the chain ID "1".
```
pxrl, line 96: Do we normally try to distinguish between hard requirements and recommendations in UMIPs?

The use of `should` normally indicates something is optional in requirement text, but this seems like more of a `shall`/`must` to me. [IETF RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) gives some good guidance on when to use should or shall/must in requirement text.

mrice32 (dismissed):
LGTM!

UMIPs/umip-157.md
mrice32: Agreed, I think routes get disabled first and then the chain gets added to the disabled list. +1

UMIPs/umip-157.md
nicholaspai: We use "should" all throughout this document, i think its implied these are the rules. I'm happy to change this to "must" though

mhairimcalpine (approved):

PR #578, opened 2023-03-08
feat: ENABLED_CHAINS

nicholaspai:
Replaces and improves on the disabled chain logic introduced in #576 . Instead of storing a list of disabled chains, store the list of all enabled chains. This allows the dataworker to remove hardcoded list of chain ID indices and infer the disabled chain list (by comparing the CHAIN_ID_INDICES list and the ENABLED_CHAINS list)


mrice32:
@nicholaspai sorry, I haven't reviewed this because it's been in draft -- would you like a review on this or are we not planning to move forward with it?

nicholaspai:
> @nicholaspai sorry, I haven't reviewed this because it's been in draft -- would you like a review on this or are we not planning to move forward with it?

@mrice32 I still think we want this reviewed as it seems more flexible going forward

mrice32 (approved):
Looks good! Do you see any issues with adding and removing chains in this setup?

PR #581, opened 2023-04-27
feat!: Update Across rules to replace existing fee model with UBA fee model

nicholaspai:
This PR is a WIP and i'm publishing the draft so we can align closely with engineering implementation of UBA


UMIPs/umip-157.md
```diff
@@ -253,6 +318,19 @@ For a given slow relay identified [above](#finding-slow-relays), we can compute
...
+Slow fills are implicitly created whenever there is a partial fill sent for a deposit and the fill is not eventually filled at the time of a bundle proposed including the deposit.
+
+The deposit has already set a `realizedLpFeePct`, so when the slow fill is executed, the deposit recipient will receive the deposited amount minus the `realizedLpFeePct`, which is equal to the deposit incentive fee.
+
+However, slow fills are unlike normal fills in that there is no relayer to whom should be charged the refund incentive fee. Slow fills ultimately cause a token outflow from the destination spoke pool, so intuitively the slow fill should be charged the refund incentive fee.
+
+The `payoutAdjustmentPct` is therefore set equal to the `refundFee` at the time of the first partial fill `e` that triggered
+the deposit. Follow [this guide](#computing-uba-fees) to determine the refund fee for `e`.
+
+TODO: Since refund fees [are denominated in tokens](#using-the-incentive-fee), its not straightforward how to set `payoutAdjustmentPct` since the amount to be executed can be less than the amount earmarked for the slow fill originally.
```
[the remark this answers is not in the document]
nicholaspai: Actually this might not be an issue as all, as long as `payoutAdjustmentPct = refundFee / e.amount`, the correct portion of the executed slow fill amount is paid

[Update umip-157.md](https://github.com/UMAprotocol/UMIPs/pull/581/commits/45ea6ab03c21491bc0fd490f021dd0e573443e37)

nicholaspai:
Requesting reviews for the new UBA sections I've added. After this round of reviews I'll go through the rest of the existing UMIP and change language that references older rate model/utilization models

UMIPs/umip-157.md
```diff
@@ -65,10 +65,20 @@ The root bundle is valid once *all* of its `PoolRebalanceLeaves` are executed vi
...
-The [`RootBundleExecuted`](https://github.com/across-protocol/contracts-v2/blob/a8ab11fef3d15604c46bba6439291432db17e745/contracts/HubPool.sol#L157-L166) event and [`PoolRebalanceLeaf`] structure both contain equal length arrays: `l1Tokens`, `netSendAmounts`, `bundleLpFees`, and `runningBalances`. Each `l1Token` value in `l1Tokens` is an address correspondingto an ERC20 token deployed on Ethereum Mainnet. It should be mapped to the value in any of the other three arrays (`netSendAmounts`, `bundleLpFees`, and `runningBalances`) that shares the same index within the array.
+The [`RootBundleExecuted`](https://github.com/across-protocol/contracts-v2/blob/a8ab11fef3d15604c46bba6439291432db17e745/contracts/HubPool.sol#L157-L166) event and [`PoolRebalanceLeaf`] structure both contain arraus: `l1Tokens`, `netSendAmounts`, `bundleLpFees`, and `runningBalances`. Each `l1Token` value in `l1Tokens` is an address corresponding to an ERC20 token deployed on Ethereum Mainnet. It should be mapped to the value in any of the other three arrays (`netSendAmounts`, `bundleLpFees`, and `runningBalances`) that shares the same index within the array.
```
pxrl, line 68:
```suggestion
The [`RootBundleExecuted`](https://github.com/across-protocol/contracts-v2/blob/a8ab11fef3d15604c46bba6439291432db17e745/contracts/HubPool.sol#L157-L166) event and [`PoolRebalanceLeaf`] structure both contain arrays: `l1Tokens`, `netSendAmounts`, `bundleLpFees`, and `runningBalances`. Each element in `l1Tokens` is an address corresponding to an ERC20 token deployed on Ethereum Mainnet. It should be mapped to the value in any of the other three arrays (`netSendAmounts`, `bundleLpFees`, and `runningBalances`) that shares the same index within the array.
```

UMIPs/umip-157.md
```diff
@@ -65,10 +65,20 @@ The root bundle is valid once *all* of its `PoolRebalanceLeaves` are executed vi
...
 
 For example, if `l1Tokens` is "[0x123,0x456,0x789]" and `netSendAmounts` is "[1,2,3]", then the "net send amount" for token with address "0x456" is equal to "2".
 
+## Identifying running balances and incentive pool amounts for {token, chain} combinations
+
+The `l1Tokens`, `netSendAmounts`, `bundleLpFees` arrays emitted in `RootBundleExecuted` events should be equal length so we can match each L1 token to a value in the following arrays. The `runningBalances` should be of length exactly twice the `l1Tokens` length. 
+
+Let's say there are `X` L1 tokens. The values in `runningBalances[0:X)` are equal to the running balances, and the values in `runningBalances[X:2x)` are equal to the incentive pool amounts for each token.
```
pxrl, line 76: I haven't seen [x, y) notation used for indexing arrays before - is that a standard anywhere?

If not, should we maybe use [x, y-1] instead?

UMIPs/umip-157.md
```diff
@@ -103,31 +113,162 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+> TODO
 
 _This UMIP will explain later how global and token-specific configuration settings are used._
 
+## Computing UBA Fees
+
+This section gives an overview for computing "UBA" (Universal Bridge Adapter) incentive fees, which are charged on all "Bridging Events".
+
+### Defining Bridging Events
+A Bridging Event could be a `Deposit`, `Fill` or `Refund` event, which are intuitively the only events that lead to inflows and outflows to and from a SpokePool. Every deposit is counted as a bridging event, but not all fills and refunds are. A fill is counted if its `repaymentChainId` is set to the `destinationChainId`, i.e. the fill will eventually be taking funds out of the SpokePool via a refund. If a fill's `repaymentChainId` is not set equal to the `destinationChainId`, then a corresponding refund event on the `repaymentChainId` should be emitted in order to signal when the refund from the SpokePool on the repayment chain should be counted as an outflow. See the section on [validating Refunds](#TODO) for  more details on matching refunds with fills.
```
pxrl, line 135:
```suggestion
A Bridging Event could be a `Deposit`, `Fill` or `Refund` event, which are intuitively the only events that lead to inflows and outflows to and from a SpokePool. Every deposit is counted as a bridging event, but not all fills and refunds are. A fill is counted if it is the first fill for a deposit on another chain, and its `repaymentChainId` is set to the `destinationChainId`, i.e. the fill will eventually be taking funds out of the SpokePool via a refund. If a fill's `repaymentChainId` is not set equal to the `destinationChainId`, then a corresponding refund event on the `repaymentChainId` should be emitted in order to signal when the refund from the SpokePool on the repayment chain should be counted as an outflow. See the section on [validating Refunds](#TODO) for  more details on matching refunds with fills.
```

pxrl: The above is just to reference the constraint about partial fills and slow fills. The wording may not be the best, though.

UMIPs/umip-157.md
```diff
@@ -103,31 +113,162 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+### Computing Incentive Fee using incentive pool size, running balance, and incentive curve for e
+
+First we'll compute the "uncapped" incentive fee amount. This is equal to the inverse of the integral of the incentive curve between [`runningBalance`, `runningBalance + e.amount`] where `e.amount > 0` for deposits and `e.amount < 0` for refunds and fills.
+
+Intuitively, this means that for deposits that add balance to running balances that are significantly under target, the incentive curve will return a higher % for the post-deposit running balance. Because the running balance is under target, the incentive curve will move from a negative % to a more negative % and the integral between these %'s will be a negative value. This means that the uncapped incentive fee amount will be a penalty (i.e. a positive fee) applied to the deposit.
```
pxrl, line 143: I'm having trouble reconciling this statement; perhaps because I have the wrong understanding of what it means for the runningBalance to be _under target_.

It seems to say that such deposits would be penalised by way of a positive fee / negative incentive. Shouldn't deposits into a SpokePool that is _under_ its target balance receive a negative fee / positive incentive ?

UMIPs/umip-157.md
nicholaspai: Good point, its more of [math notation](https://stackoverflow.com/questions/4396290/what-does-this-square-bracket-and-parenthesis-bracket-notation-mean-first1-last) but I think `y-1` is more clear

UMIPs/umip-157.md
```diff
@@ -103,31 +113,162 @@ The following constants are currently specified in this UMIP directly, but shoul
...
-  - If "spokeTargetBalances", a particular chainId is missing from the mapping, or "target" or "threshold" is missing,
-    the missing "target" and "threshold" should be defaulted to 0.
-- "transferThreshold"
-  - This field should be a single integer value represented as a string (to allow unlimited precision).
+- "UBAFeeModel"
+  - This is a JSON struct of parameters keyed to a specific `chain` and `token` that defines a fee curve for that token given an input `balance`. This model is described in more detail [here](#TODO) along with visual aids. The parameters are used to construct curves that return percentages given an input running balance.
+- "routeUBAFeeModel"
+  - Similar to `UBAFeeModel`, this is a dictionary of `originChain-destinationChain` keys mapped to models that take precedence over the `UBAFeeModel` for a token on that specific deposit route. The route rate models should contain exactly the same parameters as the default `UBAFeeModel`
+- "incentivePoolAdjustment"
+  - Used by DAO to keep track of any donations made to add to `incentivePool` liquidity.
```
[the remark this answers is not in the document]
nicholaspai: Imagine we start at at T=0 either below or above target. There is a reward to either deposit or refund, and these rewards must come out of the incentive pool. So in this case, the DAO has no choice but to donate to the pool.

Alternatively, the target must start exactly at target, which isn't always desirable

UMIPs/umip-157.md
nicholaspai: yea I think i'll just remove this sentence as I think all of the information is already included in the definition:
- `The inverse of the integral of the incentive curve between [`runningBalance`, `runningBalance + e.amount`]`

UMIPs/umip-157.md
```diff
@@ -103,31 +113,162 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+   ## Uncapped incentive fee owed to e before we apply any discounts based on size of penalty pool
+   uncappedIncentiveFee = integral_incentive_curve[startingRunningBalance, startingRunningBalance + e.amount]
+
+   ## Deposits are rewarded when startingRunningBalance is under target
+   if (startingRunningBalance < target):
+      maxRewards = integral_incentive_curve[startingRunningBalance, target]
+
+## Handling refunds
+if (e is "Refund"):
```
pxrl, line 169: Is it intended to exclude fills here? Intuitively I feel it should be `Inflow` and `Outflow` rather than `Deposit` and `Refund`.

UMIPs/umip-157.md
```diff
@@ -103,31 +113,162 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+## Discount fee if max reward exceeds penalty pot
+if (maxRewards > P):
+   
+   ## If P << uncappedIncentiveFee, discountFactor --> 100%
+   discountFactor = (uncappedIncentiveFee - P) / uncappedIncentiveFee
+   appliedIncentiveFee *= 1 - discountFactor
+
+## Apply hardcoded multiplier
+
+appliedIncentiveFee *= UBA_REWARD_MULTIPLIER
```
pxrl, line 192: Are there any scenarios where a UBA_REWARD_MULTIPLIER > 1 are anticipated?

UMIPs/umip-157.md
```diff
@@ -103,31 +113,162 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+The final reward is equal to `reward * ubaRewardMultiplier`, a convenient scaler variable set in the [config store](#token-constants).
+
+The following sections explain how to find the specific inputs needed to [compute the incentive fee](#computing-incentive-fee-using-incentive-pool-size-running-balance-and-incentive-curve-for-e)
+
+### Computing Running Balance for Bridging Event
+
+#### Finding the Starting Running Balance
+
+Identify the last validated `runningBalance` included with a `bundleEndBlock` preceding `e`. If `e` is a deposit, then use the `deposit.quoteTimestamp`, and if `e` is a fill, then use the [fill's matched deposit's](#TODO) `quoteTimstamp`. This `runningBalance` should be for the [L1 token equivalent of `e`'s L2 token](#finding-the-preceding-running-balance-for-an-l1-token) and should be included in the latest `ExecutedRootBundle` event (included in a validated bundle) for `e`'s `chainId`.
```
pxrl, line 203: I notice that a refund example is used on line 205, but refunds are not identified in this paragraph. Should `deposit` and `fill` be `inflow` and `outflow`?

pxrl, line 203:
```suggestion
Identify the last validated `runningBalance` included with a `bundleEndBlock` preceding `e`. If `e` is a deposit, then use the `deposit.quoteTimestamp`, and if `e` is a fill, then use the [fill's matched deposit's](#TODO) `quoteTimestamp`. This `runningBalance` should be for the [L1 token equivalent of `e`'s L2 token](#finding-the-preceding-running-balance-for-an-l1-token) and should be included in the latest `ExecutedRootBundle` event (included in a validated bundle) for `e`'s `chainId`.
```

pxrl, line 203: Is it correct to use the deposit quoteTimestamp to identify the start point for the running balance? I'd understood that we should only use the block number of the event itself, and that the quoteTimestamp would instead be used for identifying the applicable rate model that would be applied.

In general I am wary of relying on the quoteTimestamp too much because we permit them to be up to 3600s in the future. In some respect I'd prefer it if we clamped the quoteTimestamp to the timestamp of the block where it is confirmed, though I concede that this means we're cross-referencing HubPool and SpokePool timestamps, which is not desirable.

UMIPs/umip-157.md
```diff
@@ -103,31 +113,162 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+Let's name this preceding running balance the "Starting Running Balance". Let's also name the proposed bundle associated with this starting running balance as the "Preceding Validated Bundle".
+
+#### Finding the Running Balance for e
+
+Step through all events between the last validated bundle's end block and `e` (e.g. all `e`'s with blocks >= the `bundleEndBlock` for the chain `e.chainId` in the preceding validated bundle and <= `e.block`). For each deposit, add the `e.amount` to the starting running balance, and for each refund or fill, subtract `e.amount` from the starting running balance.
+
+##### Handling running balance bounds
+If at any point in time, the running balance exceeds the [target upper or lower bounds](#selecting-running-balance-bounds-for-bridging-event), then reset the running balance count to the target. The running balance for computing the incentive fee for `e` is computed as if the running balance at the time of the event `e` was added to `e.amount`. The next bridge event `e_{i+1}` just will use the target running balance instead of the starting running balance `+ e.amount`.
```
pxrl, line 214: My preference would be for a "soft" pullback to the target, rather than a hard snap back. To use the bowling alley bumper bar analogy, hitting the bumper doesn't mean that the ball is immediately returned to the middle of the lane.

UMIPs/umip-157.md
nicholaspai: Good point i'll switch to inflow/outflow as that was what I was going for

UMIPs/umip-157.md
nicholaspai: Hmm probably not in practice but do you think its worth blocking at the UMIP level?

UMIPs/umip-157.md
```diff
@@ -103,31 +113,162 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+### Computing Running Balance for Bridging Event
+
+#### Finding the Starting Running Balance
+
+Identify the last validated `runningBalance` included with a `bundleEndBlock` preceding `e`. If `e` is a deposit, then use the `deposit.quoteTimestamp`, and if `e` is a fill, then use the [fill's matched deposit's](#TODO) `quoteTimstamp`. This `runningBalance` should be for the [L1 token equivalent of `e`'s L2 token](#finding-the-preceding-running-balance-for-an-l1-token) and should be included in the latest `ExecutedRootBundle` event (included in a validated bundle) for `e`'s `chainId`.
+
+For example if `e` is a deposit on chain 1, then find the latest validated running balance for chain `1` and if `e` is a refund on chain 5, then find the latest validated running balance for chain `5`. 
+
+Let's name this preceding running balance the "Starting Running Balance". Let's also name the proposed bundle associated with this starting running balance as the "Preceding Validated Bundle".
```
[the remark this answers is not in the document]
nicholaspai: Great point. Have tried to edit this in this commit b22b91ed5003ce25f6a24c7c51763ce9bb458605

UMIPs/umip-157.md
nicholaspai: yea this part is confusingly worded but you're right I'll reword

UMIPs/umip-157.md
nicholaspai: This seems like a question for Chase/Korpi but a "soft" pullback is a bit more complex to implement. What if `e` is a very large inflow such that the new running balance is >>> target. How do you determine how much to pullback? Compare that to the case where `e` is a small inflow that just takes running balance = target + 1

UMIPs/umip-157.md
```diff
@@ -103,31 +113,166 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+In the next section, we'll find the "Closing Balance" for e: the running balance of the SpokePool after `e` is executed.
+
+#### Finding the Closing Balance for e
+
+Step through all events between the last validated bundle's end block and `e` (e.g. all `e`'s with blocks >= the `bundleEndBlock` for the chain `e.chainId` in the preceding validated bundle and <= `e.block`). For each deposit, add the `e.amount` to the opening balance, and for each refund or fill, subtract `e.amount` from the opening balance.
+
+##### Handling running balance bounds
+If at any point in time, the running balance exceeds the [target upper or lower bounds](#selecting-running-balance-bounds-for-bridging-event), then reset the running balance count to the target. The running balance for computing the incentive fee for `e` is computed as if the running balance at the time of the event `e` was added to `e.amount`. The next bridge event `e_{i+1}` just will use the target running balance instead of the opening balance `+ e.amount`.
```
mrice32, line 218: Is the target referenced here the same as the UBA target? I would (naively) assume not, right?

I think these need to be distinct because our UBA target is usually the middle of the range of running balances, but our target in our bound threshold/targets is usually farther out on the running balance curve. But maybe I'm thinking about it incorrectly.

For example, a 100 USDC refund on a running balance of 0 might trigger a $2MM transfer from the HubPool to arbitrum, which wouldn't be desirable, right, since it is expensive to get funds off of arbitrum.

UMIPs/umip-157.md
```diff
@@ -103,31 +113,166 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+Once you get to `e`, stop and you have the running balance for `e`.
+
+### Computing Incentive Pool for Bridging Event
+
+#### Finding the Opening Incentive Pool
+
+Similar to [finding the closing balance](#finding-the-closing-balance-for-e), we need to identify the last validated `incentivePool` included with a `bundleEndBlock` preceding `e`. Follow [these steps](#identifying-running-balances-and-incentive-pool-amounts-for-token-chain-combinations) to find the incentive pool size within a `runningBalance` array for the chain associated with `e`.
+
+We now have the "OPening Incentive Pool" amount associated with a "Preceding Validated Bundle".
```
mrice32, line 228:
```suggestion
We now have the "Opening Incentive Pool" amount associated with a "Preceding Validated Bundle".
```

UMIPs/umip-157.md
```diff
@@ -103,31 +113,166 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+Finally, add the [incentive fee adjustment](#token-constants) to theOpening Incentive Pool amount to account for any donations to the pool. This rule implies that any time an Admin wants to donate to the incentive pool, they should increment the `incentiveFeeAdjustment` configuration variable to make sure that the [cap on the applied incentive fee](#computing-incentive-fee-using-incentive-pool-size-running-balance-and-incentive-curve-for-e) is raised.
```
mrice32, line 230:
```suggestion
Finally, add the [incentive fee adjustment](#token-constants) to the Opening Incentive Pool amount to account for any donations to the pool. This rule implies that any time an Admin wants to donate to the incentive pool, they should increment the `incentiveFeeAdjustment` configuration variable to make sure that the [cap on the applied incentive fee](#computing-incentive-fee-using-incentive-pool-size-running-balance-and-incentive-curve-for-e) is raised.
```

UMIPs/umip-157.md
```diff
@@ -103,31 +113,166 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+#### Selecting Running Balance Bounds for Bridging Event
+
+Identify the [UBA Fee Curve Bounds](#token-constants) for the L1 token equivalent and chain for `e` similarly to finding the incentive curve for `e`.
+
+### Using the incentive fee
+
+The incentive fee is charged to inflows and outflows differently. For outflows the `realizedLpFeePct` should be set to a percentage equal to `incentive fee / deposit.amount`. Deposit recipients specified in Inflows will ultimately receive `(1 - (deposit.relayerFeePct + fill.realizedLpFeePct)) * deposit.amount` from the relayer. In other words, relayers will send to deposit recipients the deposited amount minus the relayer fee and the deposit incentive fee.
+
+Relayers will receive a refund from the `HubPool` equal to `(1 - (fill.realizedLpFeePct + fillIncentiveFeePct)) * deposit.amount`. Therefore, relayers will be refunded the deposited amount minus the deposit and refund incentive fees. This is equal to the amount that the relayer sent to the deposit recipient plus the `relayerFee` and minus the refund incentive fee.
```
mrice32, line 252: Should this be `(1 - fill.realizedLpFeePct) * deposit.amount * (1 - fillIncentiveFeePct)`

I might be thinking about this incorrectly, but this is what I'm thinking:

Basically, the spoke pool liability is `(1 - fill.realizedLpFeePct) * amount`. We need to then adjust that payout by the fee/reward by `(1 - fillIncentiveFeePct)`. Put differently, I don't think these fees add -- I think they would compound.

Thoughts?

UMIPs/umip-157.md
```diff
@@ -103,31 +113,166 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+Dataworkers will add the incentive fees for each event in a bundle that they want to propose plus the opening incentive pool amount (from the last validated bundle) and store these in the `runningBalances` array to snapshot the latest validated incentive pool amounts.
+
+## UBA Fee Curve Parameters
+
+UBA Fee Curves are defined completely by parameters set in the `ConfigStore` at a specific point in time. If there are no parameters set for a timestamp, then the default values should be used.
+### UBA Deposit Curve Parameters
```
mrice32, line 259:
```suggestion
UBA Fee Curves are defined completely by parameters set in the `ConfigStore` at a specific point in time. If there are no parameters set for a timestamp, then the default values should be used.

### UBA Deposit Curve Parameters
```

UMIPs/umip-157.md
```diff
@@ -103,31 +113,166 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+### UBA Refund Curve Parameters 
+
+### UBA Running Balance Bounds
+
+The default bounds are INFINITY for all variables.
+
+- `target_lower_bound_usd`
+- `target_lower_bound_min_usd`
+- The lower bound target running balance for a chain will be equal to `min(target_lower_bound_usd, target_lower_bound_min_usd)`.
```
mrice32, line 269: Does the usd mean that these are set in terms of USD and not in the native currency? Also, I don't quite understand what the difference between target_lower_bound_usd and target_lower_bound_min_usd is?

mrice32, line 269: Is there a threshold and target? Or just a target? In the current system, we have threshold and target to avoid flickering/trembling where small user transfers exceed the bound and necessitate small transfers by the system. Do you think we need that in the new system?

UMIPs/umip-157.md
```diff
@@ -253,6 +394,21 @@ For a given slow relay identified [above](#finding-slow-relays), we can compute
...
+## Slow Fill payout adjustments
+
+Slow fills are implicitly created whenever there is a partial fill sent for a deposit and the fill is not eventually filled at the time of a bundle proposed including the deposit.
+
+The deposit has already set a `realizedLpFeePct`, so when the slow fill is executed, the deposit recipient will receive the deposited amount minus the `realizedLpFeePct`, which is equal to the deposit incentive fee.
+
+However, slow fills are unlike normal fills in that there is no relayer to whom should be charged the refund incentive fee. Slow fills ultimately cause a token outflow from the destination spoke pool, so intuitively the slow fill should be charged the refund incentive fee.
+
+The `payoutAdjustmentPct` is therefore set equal to the `refundFee` proportion of the outflow amount at the time of the first partial fill `e` that triggered
+the deposit. `payoutAdjustMentPct = refundFee / e.amount`.
```
mrice32, line 406:
```suggestion
the deposit. `payoutAdjustmentPct = refundFee / e.amount`.
```

UMIPs/umip-157.md
```diff
@@ -253,6 +394,21 @@ For a given slow relay identified [above](#finding-slow-relays), we can compute
 
 Note: Since  we eliminated all fills where `totalFilledAmount == deposit.amount`, the remaining "last fill" should have `totalFilledAmount < deposit.amount` AND have `totalFilledAmount > [all other fills for deposit].totaFilledAmount`.
 
+## Slow Fill payout adjustments
+
+Slow fills are implicitly created whenever there is a partial fill sent for a deposit and the fill is not eventually filled at the time of a bundle proposed including the deposit.
+
+The deposit has already set a `realizedLpFeePct`, so when the slow fill is executed, the deposit recipient will receive the deposited amount minus the `realizedLpFeePct`, which is equal to the deposit incentive fee.
+
+However, slow fills are unlike normal fills in that there is no relayer to whom should be charged the refund incentive fee. Slow fills ultimately cause a token outflow from the destination spoke pool, so intuitively the slow fill should be charged the refund incentive fee.
```
mrice32, line 403:
```suggestion
However, slow fills are unlike normal fills in that there is no relayer who should be charged the refund incentive fee. Slow fills ultimately cause a token outflow from the destination spoke pool, so intuitively the slow fill should be charged the refund incentive fee.
```

UMIPs/umip-157.md
```diff
@@ -253,6 +394,21 @@ For a given slow relay identified [above](#finding-slow-relays), we can compute
...
+
+The `payoutAdjustmentPct` is therefore set equal to the `refundFee` proportion of the outflow amount at the time of the first partial fill `e` that triggered
+the deposit. `payoutAdjustMentPct = refundFee / e.amount`.
+
+Follow [this guide](#computing-uba-fees) to determine the refund fee for `e`.
+
+Slow fills always create outflows from the destination SpokePool, but its possible that follow-on partial fills can cause the slow fill execution to create a smaller outflow than originally earmarked for the fill. This is why the `payoutAdjustmentPct` is set as a percentage of the remaining fill. amount.
```
mrice32, line 410:
```suggestion
Slow fills always create outflows from the destination SpokePool, but its possible that follow-on partial fills can cause the slow fill execution to create a smaller outflow than originally earmarked for the fill. This is why the `payoutAdjustmentPct` is set as a percentage of the remaining fill amount.
```

mrice32 (commented):
This is really really awesome! Very detailed!

We might need a bit more explanation/details on the curves (for instance, how is the inverse of the curve defined), but I think that will come once we have more well defined parameters.

UMIPs/umip-157.md
pxrl: I will concede that I can also see arguments for a hard snap-back.

These kinds of one-sided flows usually seem to occur when there are system-wide events, like an airdrop on a particular chain, or a flight to perceived safety. In such extreme cases (of which we've seen somewhere between 3 - 6 instances of in the past 12 months), it seems like a soft snap-back would limit the ability of Across to support the necessary volumes.

Would we consider making the top-up amount configurable per chain as a percentage of the gap between current and target balance? i.e. we typically see these deficits on mainnet, and it seems much safer to throw additional capital into the mainnet SpokePool, so we'd set it to 100% there. We might have different preferences for other SpokePools.

UMIPs/umip-157.md
```diff
@@ -103,31 +113,166 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+Relayers will receive a refund from the `HubPool` equal to `(1 - (fill.realizedLpFeePct + fillIncentiveFeePct)) * deposit.amount`. Therefore, relayers will be refunded the deposited amount minus the deposit and refund incentive fees. This is equal to the amount that the relayer sent to the deposit recipient plus the `relayerFee` and minus the refund incentive fee.
+
+Dataworkers will add the incentive fees for each event in a bundle that they want to propose plus the opening incentive pool amount (from the last validated bundle) and store these in the `runningBalances` array to snapshot the latest validated incentive pool amounts.
+
+## UBA Fee Curve Parameters
+
+UBA Fee Curves are defined completely by parameters set in the `ConfigStore` at a specific point in time. If there are no parameters set for a timestamp, then the default values should be used.
```
pxrl, line 258: Where are the default values defined?

UMIPs/umip-157.md
```diff
@@ -225,17 +371,12 @@ Additionally, matching relays should have their `destinationToken` set such that
...
-`liquidityUtilizationPostRelay` on the `BridgePool` contract (passing no arguments) to compute the rate model, identically-named methods would be
-called on the `HubPool` contract, passing in a single argument, the `l1Token` derived in the 3 step process above.
+To determine the validity of the `realizedLPFeePct` in the `FilledRelay` event, we must compute the [deposit incentive fee](#computing-uba-fees) for the matched deposit of the fill.
 
-If the `realizedLPFeePct` that is computed using those means does not match the `realizedLPFeePct` in the
-`FilledRelay` event, then the relay is considered invalid.
+## Setting the realizedLpFeePct for a Fill
 
-All valid `FilledRelay` events should then be stored for the construction of the bundle.
+The `realizedLpFeePct` should be equal to `depositorFee` for the matched deposit.  The `refundFee` will ultimately be returned to the Relayer and be included in the refund amount by the Dataworker.
```
pxrl, line 379: `depositorFee` doesn't seem to be a defined term in this document.

Side note: wdyt about carving out a definition for each term in the `Definitions` section?

UMIPs/umip-157.md
```diff
@@ -253,6 +394,21 @@ For a given slow relay identified [above](#finding-slow-relays), we can compute
 
 Note: Since  we eliminated all fills where `totalFilledAmount == deposit.amount`, the remaining "last fill" should have `totalFilledAmount < deposit.amount` AND have `totalFilledAmount > [all other fills for deposit].totaFilledAmount`.
 
+## Slow Fill payout adjustments
+
+Slow fills are implicitly created whenever there is a partial fill sent for a deposit and the fill is not eventually filled at the time of a bundle proposed including the deposit.
```
pxrl, line 399:
```suggestion
Slow fills are implicitly created whenever there is at least one valid partial fill for a deposit, but where the deposit has not been filled completely at the time of the bundle being proposed.
```

UMIPs/umip-157.md
nicholaspai: This is a good discussion and I just want to make sure Chase/Korpi are involved

UMIPs/umip-157.md
nicholaspai: This was suggested by Korpi, I'm going to revert to just quoting it without the USD and not using the `min`

UMIPs/umip-157.md
nicholaspai: good call out. We should distinguish between "target" and "upper bound/lower bound" running balances and also create thresholds.

ac60aa2ff32122d8f5a139d58d5ae6cbc9d3de2a

UMIPs/umip-157.md
nicholaspai: Added more details here ac60aa2ff32122d8f5a139d58d5ae6cbc9d3de2a

UMIPs/umip-157.md
nicholaspai: Ah I mean the incentive fee for the inflow. I'll clarify

UMIPs/umip-157.md
```diff
@@ -65,10 +65,20 @@ The root bundle is valid once *all* of its `PoolRebalanceLeaves` are executed vi
...
-The [`RootBundleExecuted`](https://github.com/across-protocol/contracts-v2/blob/a8ab11fef3d15604c46bba6439291432db17e745/contracts/HubPool.sol#L157-L166) event and [`PoolRebalanceLeaf`] structure both contain equal length arrays: `l1Tokens`, `netSendAmounts`, `bundleLpFees`, and `runningBalances`. Each `l1Token` value in `l1Tokens` is an address correspondingto an ERC20 token deployed on Ethereum Mainnet. It should be mapped to the value in any of the other three arrays (`netSendAmounts`, `bundleLpFees`, and `runningBalances`) that shares the same index within the array.
+The [`RootBundleExecuted`](https://github.com/across-protocol/contracts-v2/blob/a8ab11fef3d15604c46bba6439291432db17e745/contracts/HubPool.sol#L157-L166) event and [`PoolRebalanceLeaf`] structure both contain arrays: `l1Tokens`, `netSendAmounts`, `bundleLpFees`, and `runningBalances`. Each element in `l1Tokens` is an address corresponding to an ERC20 token deployed on Ethereum Mainnet. It should be mapped to the value in any of the other three arrays (`netSendAmounts`, `bundleLpFees`, and `runningBalances`) that shares the same index within the array.
```
cc7768, line 68:
```suggestion
The [`RootBundleExecuted`](https://github.com/across-protocol/contracts-v2/blob/a8ab11fef3d15604c46bba6439291432db17e745/contracts/HubPool.sol#L157-L166) event and [`PoolRebalanceLeaf`] structure both contain arrays: `l1Tokens`, `netSendAmounts`, `bundleLpFees`, and `runningBalances`. Each element in `l1Tokens` is an address corresponding to an ERC20 token deployed on Ethereum Mainnet.
```

UMIPs/umip-157.md
cc7768: User amount and relayer reimbursement come from

`user_receives = amount - lp_fee - deposit_incentive_fee - fill_incentive_fee - relayer_fee`

`relayer_reimbursement = user_receives + relayer_fee`

We also know that

`realizedLpFeePct = (lp_fee + deposit_incentive_fee) / amount`

and

`realizedRelayerFeePct = (fill_incentive_fee + relayer_fee)/amount` (I think?)

which means the relayer should receive

```
relayer_reimbursement = amount - lp_fee - deposit_incentive_fee - fill_incentive_fee - relayer_fee + relayer_fee
= amount * (1 - (lp_fee + deposit_incentive_fee)/amount - fill_incentive_fee/amount)
= amount * (1 - realizedLpFeePct - fillIncentiveFeePct)
```

I'm not super confident which is why I wrote down my logic - Please double check.

cc7768 (commented):
One kind of general comment - I feel like more explicit examples (maybe even going as far to be annoyingly explicit) could potentially go a long ways. They'd serve 2 purposes:

1. Put the pieces of logic into a single place where someone could walk through and sort out what's going on
2. Act as "unit tests" (maybe they could even be used as unit tests for the relayer/sdk code) to make sure everything behaves as expected

This is already very detailed (and in pretty good condition for the draft!) but there's a lot of logic that needs to be followed here and creating a playground for that logic seems worthwhile

UMIPs/umip-157.md
```diff
@@ -103,31 +113,188 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+For each SpokePool on `chainId`, there will be an ordered series of Bridging Events (sorted in [ascending](#comparing-events-chronologically) chronological order). To compute the incentive fee for any event `e` within `E` (the ordered series of Bridging Events), we need to determine the "Running Balance" and the "Incentive Pool" at the time that `e` was emitted. We also will need to select which Incentive Curve to use to find the Incentive fee given the running balance and available incentive pool.
+
+In the following sections we'll use "Deposit" interchangeably with "Inflow", to represent running balance added to the SpokePool. We'll also use "Outflow" interchangeably with "Fills" and "Refunds", which represent balance subtracted from the SpokePool.
+
+### Incentive Curves
+
+Incentive curves are virtual mathematical functions that will be deterministically defined by a set of [global configuration variables stored in the ConfigStore](#global-constants). There will be exactly two curves for each chain:
```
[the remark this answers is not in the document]
nicholaspai: This is interesting and new to me. How do you handle the inflow/outflow case then? Is it one function used twice: once normally, and once as the inverse, and one of them will have some adjustment +/-y applied to it?

This area is where I was sort of guessing and would like to see some example functions

UMIPs/umip-157.md
```diff
@@ -103,31 +113,188 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+- Inflow Curve
+- Outflow Curve
+
+The curves are functions `f_i_t(x)` where the input `x` is denominated as a "running balance", which represents the amount of balance held on a SpokePool. `f_i_t(x)` is unique for chain ID `i` and token `t`. `f_i_t(x)` should return a percentage, representing a "marginal incentive fee" for the input `x` running balance.
+
+These curves can be piecewise functions and should be integratable. These curves will be used to figure out how to charge fees for an inflow or outflow event `e` which adds or subtracts, respectively, to a SpokePool's running balance. Typically, the incentive fee will be equal to the negative of the integral between `e.amount + x` for inflows (or `x - e.amount` for outflows) and `x`. This is intuitively the integral between the opening marginal incentive fee and the closing incentive fee after `e` inflows or outflows some tokens from the SpokePool's running balance.
```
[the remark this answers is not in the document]
nicholaspai: Yeah I think it depends in practice where you are on the curve, but I want to generally just describe the integral. Is `x` to `x+e.amount` more standard?

UMIPs/umip-157.md
```diff
@@ -103,31 +113,188 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+Inflow curves will be used to pay or penalize the senders of `e` (i.e. the "depositor") depending on the sign of the incentive fee.
+
+Outflow curves will also be used to pay or penalize the senders of `e`, but in this case these will be "relayers" (e.g. Relayers take refunds on SpokePools, creating outflows).
+
+### Computing Incentive Fee using incentive pool size, running balance, and incentive curve for e
+
+First we'll compute the "uncapped" incentive fee amount. This is equal to the inverse of the integral of the incentive curve between [`runningBalance`, `runningBalance + e.amount`] where `e.amount > 0` for deposits and `e.amount < 0` for refunds and fills.
```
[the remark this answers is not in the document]
nicholaspai: yes definitely I just lacked the knowledge to write such math

UMIPs/umip-157.md
```diff
@@ -103,31 +113,188 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+> TODO
 
 _This UMIP will explain later how global and token-specific configuration settings are used._
 
+## Computing UBA Fees
+
+This section gives an overview for computing "UBA" (Universal Bridge Adapter) incentive fees, which are charged on all "Bridging Events".
+
+### Defining Bridging Events
+A Bridging Event could be a `Deposit`, `Fill` or `Refund` event, which are intuitively the only events that lead to inflows and outflows to and from a SpokePool. Every deposit is counted as a bridging event, but not all fills and refunds are. A fill is counted if it is the first valid fill for a deposit on another chain, and its `repaymentChainId` is set to the `destinationChainId`, i.e. the fill will eventually be taking funds out of the SpokePool via a refund. If a fill's `repaymentChainId` is not set equal to the `destinationChainId`, then a corresponding refund event on the `repaymentChainId` should be emitted in order to signal when the refund from the SpokePool on the repayment chain should be counted as an outflow. See the section on [validating Refunds](#TODO) for  more details on matching refunds with fills.
```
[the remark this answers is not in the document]
nicholaspai: I don't think an invalid fill should modify running balances

UMIPs/umip-157.md
```diff
@@ -103,31 +113,188 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+In the next section, we'll find the "Closing Balance" for e: the running balance of the SpokePool after `e` is executed.
+
+#### Finding the Closing Balance for e
+
+Step through all events between the last validated bundle's end block and `e` (e.g. all `e`'s with blocks >= the `bundleEndBlock` for the chain `e.chainId` in the preceding validated bundle and <= `e.block`). For each deposit, add the `e.amount` to the opening balance, and for each refund or fill, subtract `e.amount` from the opening balance.
+
+##### Handling running balance bounds
+If at any point in time, the running balance exceeds the [threshold upper or lower running balance](#selecting-running-balance-bounds-for-bridging-event), then reset the running balance count to the "target lower" or "target upper". The running balance for computing the incentive fee for `e` is computed as if the running balance at the time of the event `e` was added to `e.amount`. The next bridge event `e_{i+1}` just will use the target running balance instead of the opening balance `+ e.amount`.
```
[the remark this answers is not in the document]
nicholaspai: I think there is utility in having a target range with a lower and upper target. Imagine in some cases where we only want to send funds to a spoke pool (like Arbitrum) to get it back to 0, since holding funds on Arbitrum is expensive and subject to a 7 day window. If we are over threshold, then we can return some over that bridge, but we shouldn't be sending funds to Arbitrum that takes it above 0 balance but below the upper threshold

So in this case, lower target = 0, upper target = 1_000 for example. Whereas for some chains like Polygon where holding extra funds isn't as expensive, we can make the lower and upper closer

UMIPs/umip-157.md
```diff
@@ -103,31 +113,188 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+To match `e`'s `L2Token` with an `L1Token`, follow the steps [here](#matching-l2-tokens-and-l1-tokens).
+
+Let's name this preceding running balance the "Opening Balance". Let's also name the proposed bundle associated with this starting running balance as the "Preceding Validated Bundle".
+
+In the next section, we'll find the "Closing Balance" for e: the running balance of the SpokePool after `e` is executed.
+
+#### Finding the Closing Balance for e
+
+Step through all events between the last validated bundle's end block and `e` (e.g. all `e`'s with blocks >= the `bundleEndBlock` for the chain `e.chainId` in the preceding validated bundle and <= `e.block`). For each deposit, add the `e.amount` to the opening balance, and for each refund or fill, subtract `e.amount` from the opening balance.
```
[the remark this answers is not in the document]
nicholaspai: Good point. I'll fix this language

UMIPs/umip-157.md
```diff
@@ -103,31 +113,188 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+If `e` is a deposit, then find the Inflow Curve parameters, otherwise find the Outflow Curve parameters.
+
+#### Selecting Running Balance Bounds for Bridging Event
+
+Identify the [UBA Fee Curve Bounds](#token-constants) for the L1 token equivalent and chain for `e` similarly to finding the incentive curve for `e`.
+
+### Using the incentive fee
+
+The incentive fee is charged to inflows and outflows differently. For outflows the `realizedLpFeePct` should be set to a percentage equal to `incentive fee / deposit.amount`. Deposit recipients specified in Inflows will ultimately receive `(1 - (deposit.relayerFeePct + fill.realizedLpFeePct)) * deposit.amount` from the relayer. In other words, relayers will send to deposit recipients the deposited amount minus the relayer fee and the deposit incentive fee.
```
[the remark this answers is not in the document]
nicholaspai: It does account for positive and negative fees. The LP now no longer earns fees from this value. I thought the LP was implicitly earning the difference between the rewards and penalties?

UMIPs/umip-157.md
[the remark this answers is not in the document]
nicholaspai: Not really i'll remove since its kind of a key word in other contexts

UMIPs/umip-157.md
```diff
@@ -103,31 +113,188 @@ The following constants are currently specified in this UMIP directly, but shoul
...
-  - Each struct contains two sub-fields, "target" and "threshold".
-  - Each is an integer represented as a string (to allow unlimited precision).
-  - These integers should both be positive values. If either is negative, it should be treated as "0" when used.
-  - The "target" integer should be < the "threshold" integer. If not, the "theshold" integer should be treated as if
-    it contained the same value as the "target" integer when used.
-  - If "spokeTargetBalances", a particular chainId is missing from the mapping, or "target" or "threshold" is missing,
-    the missing "target" and "threshold" should be defaulted to 0.
-- "transferThreshold"
-  - This field should be a single integer value represented as a string (to allow unlimited precision).
+- "UBAFeeModel"
```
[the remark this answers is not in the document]
nicholaspai: Good point so it seems like each parameter will be set differently, and customizable with different rules (e.g. alpha and gamma can be set per route, whereas omega can be set per chain and per token). I'm going to leave this as a placeholder and not write anything until we figure out the exact parameters.

UMIPs/umip-157.md
```diff
@@ -103,31 +113,188 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+  - This is a JSON struct of parameters keyed to a specific `chain` and `token` that defines a fee curve for that token given an input `balance`. This model is described in more detail [here](#TODO) along with visual aids. The parameters are used to construct curves that return percentages given an input running balance.
+- "routeUBAFeeModel"
+  - Similar to `UBAFeeModel`, this is a dictionary of `originChain-destinationChain` keys mapped to models that take precedence over the `UBAFeeModel` for a token on that specific deposit route. The route rate models should contain exactly the same parameters as the default `UBAFeeModel`
+- "incentivePoolAdjustment"
+  - Used by DAO to keep track of any donations made to add to `incentivePool` liquidity.
+- "ubaRewardMultiplier"
```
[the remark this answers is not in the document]
nicholaspai: yeah exactly or more generally: multiply whatever reward amount you'd normally get by 0.9

UMIPs/umip-157.md
nicholaspai: Chase, this was my thinking as well and I agree with you (I think I have written this down). I don't fully understand the logic to compound fees. Aren't the incentive fees computed on the same "principal" amount?

UMIPs/umip-157.md
james-a-morris, line 1: Nit: We'll need to change the following line regarding the git hash

```
# Data Specifications and Implementation
Note 1: the following details will often refer to the [Across V2](https://github.com/across-protocol/contracts-v2) repo
at commit hash: a8ab11fef3d15604c46bba6439291432db17e745. This allows the UMIP to have a constant reference rather than...
```

UMIPs/umip-157.md
```diff
@@ -103,31 +113,216 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+There will be a different curve for each chain and token combination. Therefore the curve is a function `f_i_t(x)` where the input `x` is denominated as a "running balance", which represents the amount of balance held on a SpokePool on chain `i`. `f_i_t(x)` is unique for chain ID `i` and token `t`. `f_i_t(x)` should return a percentage, representing a "marginal incentive fee" for the input `x` running balance.
+
+These curves can be piecewise functions and should be integratable. These curves will be used to figure out how to charge fees for an inflow or outflow event `e` which adds or subtracts, respectively, to a SpokePool's running balance. Typically, the incentive fee will be equal to the negative of the integral between `e.amount + x` for inflows (or `x - e.amount` for outflows) and `x`. This is intuitively the integral between the opening marginal incentive fee and the closing incentive fee after `e.amount` is added to or subtracted from SpokePool's running balance:
+
+$\text{Incentive Fee} = \int_{\text{Running Balance}_i}^{\text{Running Balance}_i + \text{e.amount}} f_i(x) dx$
```
james-a-morris, line 148: AFAICT we should type this `f()` as `f_i_t(x)`
```suggestion
$\text{Incentive Fee} = \int_{\text{Running Balance}_i}^{\text{Running Balance}_i + \text{e.amount}} f_it(x) dx$
```

UMIPs/umip-157.md
```diff
@@ -103,31 +113,216 @@ The following constants are currently specified in this UMIP directly, but shoul
...
+
+At this point we'll present an example to demonstrate this idea of breaking down the incentive fee into the LP fee and balancing fee:
+
+#### Example computation of incentive fee for e
+
+Let's say that `e` is a deposit into SpokePool on chain `i` for 10 tokens. The SpokePool has a running balance at the time of `e` of 100. If we integrate `f(x)` between `100` and `110` (100 + 10 = 110, the closing running balance after `e` is executed), then we are returned `1`, meaning that `e` should be penalized an incentive fee of `1` token. If we integrate `g(x)` (the "outflow curve") between `100` and `110`, then we are returned `0.8`, meaning that a refund for the same amount as `e` is depositing would be rewarded an incentive fee of `0.8` tokens. This implies that the LP fee component of the `1` incentive fee amount is `1 - 0.8 = 0.2` tokens. 
+
+We now know how much to add to LP fees for `e`: `0.2` tokens. 
+
+We also know how much the running balance should be adjusted following `e`: `10 - 1 + 0.2`. Intuitively, we should be removing the incentive pool addition of `0.8` tokens (out of which we'll pay incentive rewards) from the running balance, but including the LP fees.
```
james-a-morris, line 175: I may be wrong with this, but would the running. balance be adjusted by `10 + 1 + 0.2` because the deposit in this example is additive to the running balance?

PR #590, opened 2023-08-01
feat: Add CHAIN_ID_INDICES to config store

nicholaspai:
Long standing change to move CHAIN_ID list to config store rather than enshrined only in UMIP

### DIFF
--- a/UMIPs/umip-157.md
+++ b/UMIPs/umip-157.md
@@ -93,9 +93,9 @@ The following constants should reflect what is stored in the [`AcrossConfigStore
 - "VERSION"
   - Across protocol version number. Supporting implementations should query this value against the value defined in their implementation to determine compatibility with the current protocol version. Failure to correctly evaluate the version number may mean that filled relays are not refunded from the HubPool, and may therefore result in loss of funds. For more information go [here](#versions).
 - "DISABLED_CHAINS"
-  - This must be a stringified list of chain ID numbers. This cannot contain the chain ID "1". Chains in here must be contained in `CHAIN_ID_INDICES`.
+  - This must be a stringified list of chain ID numbers. This cannot contain the chain ID "1", or the HubPool chain ID. Chains in here must be contained in `CHAIN_ID_INDICES`.
 - "CHAIN_ID_INDICES"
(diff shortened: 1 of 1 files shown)


UMIPs/umip-157.md
```diff
@@ -93,14 +93,13 @@ The following constants should reflect what is stored in the [`AcrossConfigStore
 - "VERSION"
   - Across protocol version number. Supporting implementations should query this value against the value defined in their implementation to determine compatibility with the current protocol version. Failure to correctly evaluate the version number may mean that filled relays are not refunded from the HubPool, and may therefore result in loss of funds. For more information go [here](#versions).
 - "DISABLED_CHAINS"
-  - This must be a stringified list of chain ID numbers. This cannot contain the chain ID "1".
+  - This must be a stringified list of chain ID numbers. This cannot contain the chain ID "1". Chains in here must be contained in `CHAIN_ID_INDICES`.
```
pxrl, line 96: `This cannot contain the chain ID "1".`

1, or the HubPool chain ID?

UMIPs/umip-157.md
```diff
@@ -93,14 +93,13 @@ The following constants should reflect what is stored in the [`AcrossConfigStore
...
+- "CHAIN_ID_INDICES"
+  - This should default to the value [1,10,137,288,42161]. Chains can only be added to this list to be valid.
```
pxrl, line 98: Do we want to hardcode the default chain IDs in the UMIP? It'd mean that we need to update the UMIP to append a new chain ID each time. Given that we're referencing ConfigStore config, it seems redundant to specify a default array here.

nicholaspai: i am just hard coding these to make up for the fact that all blocks prior to now won't have set this variable. Going forward we can always use this variable set in the config store

UMIPs/umip-157.md
nicholaspai: good clarification will add

UMIPs/umip-157.md
nicholaspai: 7a3d2856f5d2ebd314e2168a9ecf3116285f225d

mrice32 (approved):

james-a-morris (approved):
This is a good change.

PR #595, opened 2024-02-08
feat: Add native USDC for Optimism, Arbitrum, and Polygon

abg4:
This UPP is proposing to update Optimism, Polygon, and Arbitrum to include native USDC.


Reinis-FRP (approved):
looking good

PR #599, opened 2024-04-05
Propose MULTIPLE_CHOICE_QUERY to improve integrations using multiple …

daywiss:
The DVM should support requests for the `MULTIPLE_CHOICE_QUERY` price identifier. `MULTIPLE_CHOICE_QUERY` is intended to allow any requester to pose a question with specific answers. This UMIP does not attempt to put any other restrictions on the content of the query, and instead leaves construction of the query up to the requester within ancillary data.

### DIFF
--- a/UMIPs/umip-181.md
+++ b/UMIPs/umip-181.md
@@ -83,7 +83,7 @@ Other examples of why no answer possible can be considered as a response:
 - The description is unintelligable or nonsensical.
 - The proposed options are unintelligable or cannot possibly answer the question.
 
-Answer is too early - This is a possible response if an answer was optimistically proposed before the game ended, the proposer could not have known the answer, but the question is valid.
+Answer is too early - When a request has event based expiry, this is a possible response if an answer was optimistically proposed before the game ended, the proposer could not have known the answer, but the question is valid.
 
 # Rationale
 This specification is not as compact as YES_OR_NO in terms of bytes used, but allows a more expressive format for proposers and improved user experience when interacting with the Oracle UI.
@@ -92,9 +92,8 @@ With this specification requesters can easily validate their ancillary data usin
 # Implementation
 - Voters should decode the ancillary data and attempt to interpret the question. Interpreting the ancillary data may require ignoring any ancillary data past the last closing bracket, "}", in order to properly parse the data as JSON.
 - If there is no valid json, or the JSON data does not match the typescript type specified above, voters should return "no answer possible" (57896044618658097711785492504343953926634992332820282019728792003956564819967).
-- If the request is "event based expiry", voters should determine the time at which an answer was proposed (proposal timestamp) and make a judgement if the proposed answer could have been known at that time, according to the question.
-- If the request is not "event based expiry", voters should attempt to determine the answer at the time the request was proposed, in the case where the answer could change, based on the time the question is asked.
-- If the answer could not have been known at the time the answer was proposed, answer with "early answer" (-57896044618658097711785492504343953926634992332820282019728792003956564819968).
+- Voters should take into consideration the timestamp of the request in order to determine as best they can the correct answer at that time.
+- If the request is "event based expiry", consider the "early answer" (-57896044618658097711785492504343953926634992332820282019728792003956564819968) response if the timestamp is before the time an answer could have been known.
 - If none of those error conditions apply, voters should determine if any of the options specified are valid answers to the question based on the description, or if no options are specified, if yes or no are valid answers.
 - Vote with the value of the option provided. Each value is specified in wei and should be used as is ( no conversions ). If no options are provided assume yes is 1 and no is 0.
 - If a voter believes no answer can be determined based on the question and the options provided, return the value representing "no answer possible".
--- a/UMIPs/umip-181.md
+++ b/UMIPs/umip-181.md
@@ -14,8 +14,8 @@ The DVM should support requests for the `MULTIPLE_CHOICE_QUERY` price identifier
 Price settlement should happen based on inspecting the ancillary data choices and can be settled in the following ways:
 - Options can be specified, and if a valid answer is among them, then return the value associated with that answer.
 - No options are specified, and the answer should be intepreted as yes or no, with a value of 1 and 0 respectively.
-- If the question cannot ever be answered; For instance, it cant be parsed, is unitelligble, the options do not represent a valid answer, it should be answered with a value representing no answer possible: 57896044618658097711785492504343953926634992332820282019728792003956564819967 ( max int 256).
-- If the question can be answered eventually, just not at the current time, or at the time an answer was proposed, submit an error representing "too early", ie "magic number": -57896044618658097711785492504343953926634992332820282019728792003956564819968.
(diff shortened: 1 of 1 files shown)


UMIPs/umip-181.md
```diff
@@ -0,0 +1,93 @@
...
+
+# Rationale
+This specification is not as compact as YES_OR_NO in terms of bytes used, but allows a safer and more expressive format for proposers and improved user experience when interacting with the Oracle UI.
+With this specification requesters can easily validate their ancillary data using typescript, or a javascript runtime schema validator, giving them confidence that their data will be rendered correctly to end users.
+
+# Implementation
+- Voters should decode the ancillary data and attempt to interpret the question.
+- Voters should determine if any of the options specified are valid answers to the question based on the description, or if no options are specified, if yes or no are valid answers.
+- Vote with the value of the option provided. Each value is specified in wei. If yes, submit 1, if no submit 0.
+- If a voter cannot make a determination about what the correct answer to the question is or if there is no question present, UMA voters should return the magic number (-57896044618658097711785492504343953926634992332820282019728792003956564819968).
```
Reinis-FRP, line 86: This magic number is hardcoded in OptimisticOracleV2 and it would apply to all price identifiers (not only `YES_OR_NO_QUERY`). If requester used `MULTIPLE_CHOICE_QUERY` identifier and called `setEventBased` then nobody would be able to propose this magic number (OOv2 would be reverting).

Either explicitly warn requesters on this behavior in the security section or choose alternative magic number since here it has different interpretation than magic number in `YES_OR_NO_QUERY`.

UMIPs/umip-181.md
```diff
@@ -0,0 +1,93 @@
...
+The description instructs the voter on how to interpret the options, and submit the numerical value (0,1,5000000000000000000) according to which option is correct. Magic number is also a valid response if for some reason the question cannot be answered according to the description.
+
+# Rationale
+This specification is not as compact as YES_OR_NO in terms of bytes used, but allows a safer and more expressive format for proposers and improved user experience when interacting with the Oracle UI.
+With this specification requesters can easily validate their ancillary data using typescript, or a javascript runtime schema validator, giving them confidence that their data will be rendered correctly to end users.
+
+# Implementation
+- Voters should decode the ancillary data and attempt to interpret the question.
+- Voters should determine if any of the options specified are valid answers to the question based on the description, or if no options are specified, if yes or no are valid answers.
+- Vote with the value of the option provided. Each value is specified in wei. If yes, submit 1, if no submit 0.
```
Reinis-FRP, line 85: `Each value is specified in wei`: We might need to stress this more throughout this UMIP that this is different than in all other price identifiers where Oracle and Voter dApps scale this up to wei.

UMIPs/umip-181.md
```diff
@@ -0,0 +1,93 @@
...
+})
+```
+The description instructs the voter on how to interpret the options, and submit the numerical value (0,1,5000000000000000000) according to which option is correct. Magic number is also a valid response if for some reason the question cannot be answered according to the description.
+
+# Rationale
+This specification is not as compact as YES_OR_NO in terms of bytes used, but allows a safer and more expressive format for proposers and improved user experience when interacting with the Oracle UI.
+With this specification requesters can easily validate their ancillary data using typescript, or a javascript runtime schema validator, giving them confidence that their data will be rendered correctly to end users.
+
+# Implementation
+- Voters should decode the ancillary data and attempt to interpret the question.
```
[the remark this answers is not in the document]
md0x: +1 agree with @Reinis-FRP

UMIPs/umip-181.md
```diff
@@ -0,0 +1,93 @@
...
+- Quote Currency: NA
+- Rounding: None, there is a predetermined set of results.
+- Estimated current value of price identifier: NA
+
+
+# Ancillary Data Specifications
+When converted from bytes to UTF-8, interpret the string as a stringified JSON object in the following form:
+
+```ts
+{
```
md0x, line 49: The proposal does not specify a mechanism for handling cases where the ancillary data is malformed or cannot be decoded or parsed into JSON.

Should we default to yes or no in those cases?

UMIPs/umip-181.md
```diff
@@ -0,0 +1,93 @@
...
+  // Require a title string
+  title: string; 
+  // The full description of the question, optionally using markdown.
+  description: string; 
+  // Optionally specify labels and values for each option a user can select. 
+  // If not specified the default is ["no", "yes"], which corresponds to prices [0,1] in wei.
+  options?:[label:string,value:string][];
```
md0x, line 56: Should we have a default outcome independently of the ancillary data description for scenarios where the question cannot be answered or data is malformed?

UMIPs/umip-181.md
```diff
@@ -0,0 +1,93 @@
...
+}
+```
+An example vote, decoding ancillary data back into a JS object:
+```ts
```
md0x, line 60: Optionally, we could consider providing a TS script that helps in validating and constructing compliant ancillary data.

UMIPs/umip-181.md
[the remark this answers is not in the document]
daywiss: good point. im going to just say use decimal notation ( base 10  i guess )

UMIPs/umip-181.md
daywiss: im assuming this spec is meant for when you have to vote manualy. but yea i can emphasize this.  in the voterdapp or oracle voters will not ever need to see these numbers hopefully

UMIPs/umip-181.md
daywiss: taking care of this with new constant to represent not answerable

UMIPs/umip-181.md
```diff
@@ -0,0 +1,98 @@
...
+
+# Rationale
+This specification is not as compact as YES_OR_NO in terms of bytes used, but allows a more expressive format for proposers and improved user experience when interacting with the Oracle UI.
+With this specification requesters can easily validate their ancillary data using typescript, or a javascript runtime schema validator, giving them confidence that their data will be rendered correctly to end users.
+
+# Implementation
+- Voters should decode the ancillary data and attempt to interpret the question. Interpreting the ancillary data may require ignoring any ancillary data past the last closing bracket, "}", in order to properly parse the data as JSON.
+- If there is no valid json, or the JSON data does not match the typescript type specified above, voters should return "no answer possible" (57896044618658097711785492504343953926634992332820282019728792003956564819967).
+- Voters should determine the time at which an answer was proposed (proposal timestamp) and make a judgement if the proposed answer could have been known at that time, according to the question.
```
Reinis-FRP, line 88: For event based queries DVM request timestamp will match proposal timestamp, but if requesting contract did not call `setEventBased` then DVM request would hold original request timestamp. I don't think we should change this behavior as there might be legitimate integrations that require outcome data evaluated as of known timestamp (e.g. KPI options) and it would be also additional burden on UMA voters to figure out the original proposal timestamp if `setEventBased` was not called.

mrice32 (approved):
LGTM!

PR #606, opened 2024-12-19
improve(UMIP179): Omit 0-value refunds and 0-value empty-message slow fills

nicholaspai:
These two payments out of the spoke pool will do nothing on-chain and the dataworker should not produce merkle leafs containing them.

A slow fill could technically have an input amount of 0 but have a non-empty message, for example if the message pays the filler then it still might be deemed "fillable" by the filler.

### DIFF
--- a/UMIPs/umip-179.md
+++ b/UMIPs/umip-179.md
@@ -109,7 +109,7 @@ Across V3 defines the following new events:
 - FilledV3Relay
 
 ### V3FundsDeposited
-The `V3FundsDeposited` event emits the unique `V3RelayData` for an individual deposit. No additional fields are defined. Consumers of this event should append the `originChainId` in order to avoid unintentioanlly mixing events from different chains.
+The `V3FundsDeposited` event emits the unique `V3RelayData` for an individual deposit. No additional fields are defined. Consumers of this event should append the `originChainId` in order to avoid unintentionally mixing events from different chains.
 
 Note:
 - The `V3FundsDeposited` `outputToken` is not required to be a known HubPool `l1Token`. In-protocol arbitrary token swaps are technically supported by Across v3.
@@ -374,6 +374,7 @@ Each Relayer Refund Leaf shall be constructed as follows:
     - One entry shall exist per unique address, containing the sum of any outstanding:
         - Relayer repayments, and/or
         - Expired deposits.
+    - If the sum of the relayer repayments for a refund address is 0, then remove this entry in both arrays.
 - The `refundAddresses` and `refundAmounts` arrays shall be ordered according to the following criteria:
     1. `refundAmount` descending order, then
     2. `relayerAddress` ascending order (in case of duplicate `refundAmount` values).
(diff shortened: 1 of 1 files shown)


UMIPs/umip-179.md
```diff
@@ -374,6 +374,7 @@ Each Relayer Refund Leaf shall be constructed as follows:
     - One entry shall exist per unique address, containing the sum of any outstanding:
         - Relayer repayments, and/or
         - Expired deposits.
+    - If the sum of the relayer repayments for a refund address is 0, then remove this entry in both arrays.
```
pxrl, line 377: +1, but I think it's better placed under the subsequent bullet point, which is focused on sorting the fully populated arrays. Because the arrays are sorted first by `refundAmount`, this would just mean popping off any entries at the end of each array.

UMIPs/umip-179.md
```diff
@@ -394,7 +395,9 @@ Note:
 - Once these leaves are constructed, they can be used to form a merkle root as described in the previous section.
 
 ### Constructing the Slow Relay Root
-One Slow Relay Leaf shall be produced per valid `RequestedV3SlowRelay` event emitted within the target block range for a destination SpokePool.
+One Slow Relay Leaf shall be produced per valid `RequestedV3SlowFill` event emitted within the target block range for a destination SpokePool.
+
+A Slow Relay Leaf should not be produced if the `RequestedV3SlowFill` event's `inputAmount` is equal to 0 and the message is a zero bytes string.
```
pxrl, line 400: I think this spec that `should` is best avoided. In standard RFC terminology, `should` is a recommendation but not a requirement, but this is obviously something we want to be strict on.

```suggestion
A Slow Relay Leaf shall not be produced if the `RequestedV3SlowFill` event's `inputAmount` is equal to 0 and the message is a zero bytes string.
```

shall

https://datatracker.ietf.org/doc/html/rfc2119

pxrl, line 400: One additional thought - should we also suppress deposit refunds where `inputAmount` is 0? That's another waste of gas for the executor and we currently don't suppress it in the executor.

UMIPs/umip-179.md
```diff
@@ -394,7 +395,9 @@ Note:
 - Once these leaves are constructed, they can be used to form a merkle root as described in the previous section.
 
 ### Constructing the Slow Relay Root
-One Slow Relay Leaf shall be produced per valid `RequestedV3SlowRelay` event emitted within the target block range for a destination SpokePool.
+One Slow Relay Leaf shall be produced per valid `RequestedV3SlowFill` event emitted within the target block range for a destination SpokePool.
+
+A Slow Relay Leaf shall not be produced if the `RequestedV3SlowFill` event's `inputAmount` is equal to 0 and the message is a zero bytes string.
```
nicholaspai, line 400:
```suggestion
A Slow Relay Leaf shall not be produced if the `RequestedV3SlowFill` event's `inputAmount` is equal to 0 and the `message` is a zero bytes string.
```

pxrl (approved):
Nice additions + fixes here!

PR #607, opened 2025-01-09
feat: add umip 183

md0x:
Changes proposed in this PR:
- Add UMIP 183: Add `MULTIPLE_VALUES` as a supported price identifier

### DIFF
--- a/UMIPs/umip-183.md
+++ b/UMIPs/umip-183.md
@@ -69,7 +69,7 @@ Voters should decode the ancillary data to decipher the values requested and the
   ```ts
     Int256  
     | 32 bits  | 32 bits | 32 bits |...  
-    | unused   | value0  | value1  |... | value7  |  
+    | unused   | value0  | value1  |... | value6  |  
     | 224-255  | 192-223 | 160-191 |... |  0-31   |  
   ```
 - Example Encoding:
@@ -77,15 +77,15 @@ Voters should decode the ancillary data to decipher the values requested and the
     uint32 value0 = 107;  
     uint32 value1 = 133;  
     ...  
-    uint32 value7 = 157;  
-    int256 price = int256(uint256(value0)) << 192  |  int256(uint256(value1)) << 160 | … |  int256(uint256(value7))
+    uint32 value6 = 157;  
+    int256 price = int256(uint256(value0)) << 192  |  int256(uint256(value1)) << 160 | … |  int256(uint256(value6))
   ```
 - Example Decoding:  
   ```ts
   uint32 value0_decoded = uint32(uint256(price >> 192));  
   uint32 value1_decoded = uint32(uint256(price >> 160));  
   …  
-  uint32 value7_decoded = uint32(uint256(price));
+  uint32 value6_decoded = uint32(uint256(price));
   ```
 
 Other valid responses are:  
--- a/UMIPs/umip-183.md
+++ b/UMIPs/umip-183.md
@@ -51,12 +51,24 @@ Example ancillary data for a sports game:
 ```ts
 {
     "title": "Los Angeles Lakers vs Boston Celtics",
-    "description": "Final scores for the Los Angeles Lakers vs Boston Celtics NBA game scheduled for Jan 7, 2025.",
+    "description": `Final scores for the "Los Angeles Lakers" vs "Boston Celtics" NBA game scheduled for Jan 7, 2025.`,
     "labels": [
             "Lakers",
             "Celtics"
      ]
 }
+```
+
+When converting UTF-8 stringified JSON to hex, all double quotes (") within the ancillary data must be escaped with a backslash (\) to ensure proper ancillary data parsing and support for automated verification.
+
+For example, with the above JSON string the string to be converted to hex is:
+```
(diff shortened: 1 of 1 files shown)


UMIPs/umip-183.md
```diff
@@ -0,0 +1,98 @@
...
+-----------------------------------------
+- Price identifier name: MULTIPLE_VALUES
+- Base Currency: N/A
+- Quote Currency: N/A
+- Rounding: values encoded into the price must be unsigned integers
+- Estimated current value of price identifier: N/A
+
+
+# Ancillary Data Specifications
+When converted from bytes to UTF-8, interpret the string as a stringified JSON object in the following form:
```
Reinis-FRP, line 36: If this is to be interpreted as regular JSON object, caution should be made to escape any special characters within field values. E.g., it is common to have some words enclosed in double quotes in the description, so these should be properly escaped in the ancillary data, so that the Oracle UI can properly decode this.

UMIPs/umip-183.md
```diff
@@ -0,0 +1,98 @@
...
+This specification is not as compact as `YES_OR_NO` in terms of bytes used, but allows a more expressive format for proposers and improved user experience when interacting with the Oracle UI. With this specification requesters can easily validate their ancillary data using typescript, or a javascript runtime schema validator, giving them confidence that their data will be rendered correctly to end users.
+
+# Implementation
+Voters should decode the ancillary data to decipher the values requested and the order they should be encoded in. The values should be encoded into a `int256` response as per the following:
+
+- The values should ordered as per the label array included in the decoded ancillary data and encoded as such:  
+  ```ts
+    Int256  
+    | 32 bits  | 32 bits | 32 bits |...  
+    | unused   | value0  | value1  |... | value7  |
```
Reinis-FRP, line 72: If we start at value0 then max we can have value6 since the highest 32 bits are reserved

Reinis-FRP, line 72: also better to specify how the values should be aligned if there are less than 7 lablels: are least significant bits ignored or the highest significant bits ignored?

UMIPs/umip-183.md
```diff
@@ -0,0 +1,98 @@
...
+- Example Decoding:  
+  ```ts
+  uint32 value0_decoded = uint32(uint256(price >> 192));  
+  uint32 value1_decoded = uint32(uint256(price >> 160));  
+  …  
+  uint32 value7_decoded = uint32(uint256(price));
+  ```
+
+Other valid responses are:  
+- **No answer possible**: if values can not be resolved (e.g. the specified event was cancelled) or encoded into a price as per the above (e.g. the ancillary data of the request specified more than 7 labels, any single value exceeds 32 bits) the request should resolve to `57896044618658097711785492504343953926634992332820282019728792003956564819967` or `type(int256).max`.
```
Reinis-FRP, line 92: What happens if there are less than 7 labels, but there are more non-zero values in decoded slots? Is that a valid price or any unused value slots must have all their bits set to 0?

UMIPs/umip-183.md
md0x: Nice catch

md0x: Good point. I would say that the lest significant bits should be ignored and should be zero bits if there are less than 7 labels. For example if only 2 labels:
```
 | 32 bits  | 32 bits | 32 bits | 32 bits | 32 bits | 32 bits | 32 bits | 32 bits |
 | unused   | value0  | value1  | ignored | ignored | ignored | ignored | ignored |
 ```
 
 Wdyt @Reinis-FRP ?

File: UMIPs/umip-183.md (hunk 3 above)
Comment (md0x) [reply UMIPs/umip-183.md:92]: Yeah, I think that should be invalid. I agree that we should specify that if non-zero values are present in ignored slots, the price is invalid.

File: UMIPs/umip-183.md (hunk 1 above)
Comment (md0x) [reply UMIPs/umip-183.md:36]: Thanks @Reinis-FRP, added a comment to clarify this.

File: UMIPs/umip-183.md (hunk 2 above)
Comment (md0x) [reply UMIPs/umip-183.md]: Fixed [here](https://github.com/UMAprotocol/UMIPs/pull/607/commits/364fc907fefbfb54f67e97c36178d49687e143c9)

File: UMIPs/umip-183.md [hunk 4]
```diff
@@ -0,0 +1,115 @@
...
+  uint32 value6_decoded = uint32(uint256(price));
+  ```
+
+ If there are fewer than 7 labels provided, the values should be encoded starting with the most significant slots, leaving the least significant slots unused. The unused slots must be set to zero. Any non-zero values present in these unused slots will render the price invalid.
+
+Other valid responses are:  
+- **No answer possible**: Voters should resolve the request to `57896044618658097711785492504343953926634992332820282019728792003956564819967` or `type(int256).max` if any of the following conditions are met:
+  - The event cannot be resolved (e.g., it was canceled).
+  - The price cannot be encoded as specified (e.g., more than 7 labels, any value exceeds 32 bits).
+  - The ancillary data is invalid JSON or does not match the specified TypeScript schema. However, voters should attempt to decode and interpret the ancillary data, and ignoring any data beyond the last closing bracket (}) is acceptable to properly parse the JSON.
```
md0x, line 109: Note that disputes comming from child chains prepend extra information to the ancillaryData so it's ok to ignore it when parsing the JSON

Ref: https://github.com/UMAprotocol/UMIPs/blob/364fc907fefbfb54f67e97c36178d49687e143c9/UMIPs/umip-181.md#L93

pumpedlunch (approved):

PR #609, opened 2025-01-16
Update umip-183.md

pumpedlunch:
-revised value0 to start at the least significant bits to be more intuitive 
-added solidity and js code snippets to demonstrate encoding and decoding 
-added an example of a values array and the resulting price

### DIFF
--- a/UMIPs/umip-183.md
+++ b/UMIPs/umip-183.md
@@ -85,6 +85,11 @@ Voters should decode the ancillary data to decipher the values requested and the
     | 224-255  | 192-223 | 160-191 |... |  0-31   |  
   ```
 
+For each label provided in the `labels` array, its corresponding value (if present) is encoded into the following bit ranges:
+- The **first label** (`labels[0]`) corresponds to the first 32 bits (bits 0–31).
+- The **second label** (`labels[1]`) corresponds to the next 32 bits (bits 32–63).
+- This pattern continues up to the **seventh label** (`labels[6]`), which corresponds to bits 192–223.
+
 If there are fewer than 7 labels provided, the values should be encoded starting with the least significant positions, leaving the most significant positions unused. The unused positions must be set to zero. Any non-zero values present in these unused positions will render the price invalid. The most significant 32 bits are always unused to prevent collisions with the other valid responses below.
 
 Other valid price responses are:  
--- a/UMIPs/umip-183.md
+++ b/UMIPs/umip-183.md
@@ -41,7 +41,7 @@ When converted from bytes to UTF-8, interpret the string as a stringified JSON o
   title: string; 
   // description of the request
   description: string; 
-  // Values will be encoded into the settled price in the same order as the provided Labels. The oracle UI will display each Label along with an input field. 7 Labels maximum.
+  // Values will be encoded into the settled price in the same order as the provided Labels. The oracle UI will display each label along with an input field. 7 labels maximum.
   labels: string[] 
 }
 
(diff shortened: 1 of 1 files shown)


UMIPs/umip-183.md
```diff
@@ -77,38 +77,97 @@ This specification is not as compact as `YES_OR_NO` in terms of bytes used, but
...
-- Example Decoding:  
-  ```ts
-  uint32 value0_decoded = uint32(uint256(price >> 192));  
-  uint32 value1_decoded = uint32(uint256(price >> 160));  
-  …  
-  uint32 value6_decoded = uint32(uint256(price));
-  ```
 
-If there are fewer than 7 labels provided, the values should be encoded starting with the most significant positions, leaving the least significant positions unused. The unused positions must be set to zero. Any non-zero values present in these unused positions will render the price invalid.
+If there are fewer than 7 labels provided, the values should be encoded starting with the least significant positions, leaving the most significant positions unused. The unused positions must be set to zero. Any non-zero values present in these unused positions will render the price invalid. The most significant 32 bits are always unused to prevent collisions with the other valid responses below.
```
md0x, line 88: It would be helpful to elaborate on this for clarity:

```suggestion
For each label provided in the `labels` array, its corresponding value (if present) is encoded into the following bit ranges:
- The **first label** (`labels[0]`) corresponds to the first 32 bits (bits 0–31).
- The **second label** (`labels[1]`) corresponds to the next 32 bits (bits 32–63).
- This pattern continues up to the **seventh label** (`labels[6]`), which corresponds to bits 192–223.

If there are fewer than 7 labels provided, the values should be encoded starting with the least significant positions, leaving the most significant positions unused. The unused positions must be set to zero. Any non-zero values present in these unused positions will render the price invalid. The most significant 32 bits are always unused to prevent collisions with the other valid responses below.
```

md0x (approved):
LGTM. I added a suggestion to clarify the label ordering.

UMIPs/umip-183.md
```diff
@@ -77,38 +77,97 @@ This specification is not as compact as `YES_OR_NO` in terms of bytes used, but
...
+            throw new Error("Index out of range");
+        }
+        const decodedValue = Number((encodedPrice >> BigInt(32 * index)) & BigInt(0xFFFFFFFF));
+        return decodedValue;
+  }
+  ```
+- For example, encoding the following values:
+  [0, 1, 2, 3, 4, 5, 6],
+  results in the following price:
+  43939712156475775172537355306597378876443456848946572296193
```
daywiss, line 169: im not seeing this decode correctly using the js logic. working on a fix

UMIPs/umip-183.md
```diff
@@ -77,38 +77,97 @@ This specification is not as compact as `YES_OR_NO` in terms of bytes used, but
...
+        return price === MIN_INT256;
+  }
+
+  isUnresolvable(price) {
+        const MAX_INT256 = (BigInt(2) ** BigInt(255)) - BigInt(1);
+        return price === MAX_INT256;
+  }
+
+  decodePriceAtIndex(encodedPrice, index) {
+        if (index < 0 || index >= 7) {
```
daywiss, line 159: this is a bug, index needs to be > 7 not >=7

UMIPs/umip-183.md
daywiss: need to double check this, but when i encode i get `37662610419627592771030380598185861001628732695723288559616 `

daywiss (changes requested):

UMIPs/umip-183.md
```diff
@@ -77,38 +77,97 @@ This specification is not as compact as `YES_OR_NO` in terms of bytes used, but
...
+  ```solidity
+    function encodeValues(uint32[] memory values)
+        external
+        pure
+        returns (int256 encodedPrice)
+    {
+        require(values.length <= 7, "Maximum of 7 values allowed");
+
+        for (uint256 i = 0; i < values.length; i++) {
+            encodedPrice |= int256(uint256(values[i])) << (32 * i);
```
daywiss, line 109: values[1] is wrapped in uint256 twice, this seems like a mistake?

pumpedlunch: it's uint256 first, then int256

UMIPs/umip-183.md
```diff
@@ -77,38 +77,102 @@ This specification is not as compact as `YES_OR_NO` in terms of bytes used, but
...
+        return price === MIN_INT256;
+  }
+
+  isUnresolvable(price) {
+        const MAX_INT256 = (BigInt(2) ** BigInt(255)) - BigInt(1);
+        return price === MAX_INT256;
+  }
+
+  decodePriceAtIndex(encodedPrice, index) {
+        if (index < 0 || index >= 7) {
```
md0x, line 164: Should it be 6 in this case?
```suggestion
        if (index < 0 || index >6) {
```

pumpedlunch: agreed.

UMIPs/umip-183.md
pumpedlunch: agreed

md0x (approved):

daywiss (approved):

PR #602, opened 2024-04-22
[WIP] [fix]: Address UMIP-179 feedback

pxrl:
- Specify proposer action on invalid FilledV3Relay repaymentChainId.
- Clarify proposer action on un-executable RelayerRefundLeaf.
- Add link to OZ audit for Across V3.

PR #611, opened 2025-01-17
[WIP] feat: Support pending SpokePool upgrades

pxrl:
This is a first pass at establishing concurrent support for the new & legacy events. There's stilll work to be done to (at least):
- Specify how to handle new relayer repayment capabilities.
- Specify how to handle pre-fills where the relayer orchestrates the deposit after the fill. There are edge cases to consider.

### DIFF
--- a/UMIPs/umip-179.md
+++ b/UMIPs/umip-179.md
@@ -39,24 +39,49 @@ All global constants from UMIP-157 will be retained for use in Across v3. The gl
     * This shall be a stringified list of chain ID numbers. Each chain in this list shall also appear in the `CHAIN_ID_INDICES` list. There shall be no duplicates in this list; any `LITE_CHAIN_ID_INDICES` update containing duplicates may be ignored. Chains may be removed from the `LITE_CHAIN_ID_INDICES` list in order to remove their "Lite" chain designation.
 
 ## Data Types
-Across v3 introduces the following new data types:
+Across v3 defines the following data types:
 
 ### V3RelayData
-The V3RelayData type underpins the transfer of funds in or out of a SpokePool instance. V3RelayData is defined as follows:
+The `V3RelayData` type underpins the transfer of funds in or out of a SpokePool instance. `V3RelayData` is defined as follows:
 | Name | Type | Description |
 | :--- |:---- | :---------- |
-| depositor | address | The address that made the deposit on the origin chain. |
-| recipient | address | The recipient address on the destination chain. |
-| exclusiveRelayer | address | The optional exclusive relayer who can fill the deposit before the exclusivity deadline. |
-| inputToken | address | The token that is deposited on the origin chain by the depositor. |
-| outputToken | address | The token that is received on the destination chain by the recipient. |
+| depositor | bytes32 | The address that made the deposit on the origin chain. |
+| recipient | bytes32 | The recipient address on the destination chain. |
+| exclusiveRelayer | bytes32 | The optional exclusive relayer who can fill the deposit before the exclusivity deadline. |
+| inputToken | bytes32 | The token that is deposited on the origin chain by the depositor. |
+| outputToken | bytes32 | The token that is received on the destination chain by the recipient. |
 | inputAmount | uint256 | The amount of inputToken that is deposited by the depositor. |
 | outputAmount | uint256 | The amount of outputToken that is received by the recipient. |
 | originChainId | uint256 | The chain ID of the origin SpokePool. |
-| depositId | uint32 | The ID uniquely identifying a deposit on the origin chain. |
+| depositId | uint256 | The ID identifying a deposit on the origin chain. |
 | fillDeadline | uint32 | The Unix timestamp on the destination chain after which the deposit can no longer be filled. |
 | exclusivityDeadline | uint32 | The optional Unix timestamp on the destination chain after which any relayer can fill the deposit. |
 | message | bytes | Optional data that is forwarded to the recipient as part of a relay. |
+
+#### Note
+- `V3RelayData` specifies `bytes32` representation for addresses (`depositor`, `recipient`, ...) in order to interface with non-EVM chains. EVM addressed supplied shall be promoted to type `bytes32` with the upper bytes zeroed. 
+
+### V3RelayDataLegacy
+The `V3RelayDataLegacy` type is supported for backwards compatibility, but is slated for deprecation. `V3RelayDataLegacy` has the following delta to the `V3RelayData` type:
+| Name | Type |
+| :--- |:---- |
+| depositor | address |
+| recipient | address |
+| exclusiveRelayer | address |
+| inputToken | address |
+| outputToken | address |
+| depositId | uint32 |
+
+## V3RelayExecutionParams
+The `V3RelayExecutionParams` type is supplied by a relayer or executor when completing a fill. `V3RelayExecutionParams` is defined as follows:
+| Name | Type | Description |
+| :--- |:---- | :---------- |
+| relay | V3RelayData | The V3RelayData object corresponding to the origin chain deposit to be filled. |
+| relayHash | bytes32 | The keccak256 hash of the V3RelayData object. See also [Computing RelayData hashes](#computing-relaydata-hashes). |
+| updatedOutputAmount | uint256 | The effective amount to be received by the recipient. This may be different to the deposit outputAmount. |
+| updatedRecipient | bytes32 | The effective recipient address. This may be different to the deposit recipient. |
+| updatedMessage | bytes | The effective message (if any) to be executed by the destination SpokePool. |
+| repaymentChainId | uint256 | The repayment chain ID requested by the relayer completing the fill. This field is not relevant for slow fills. |
 
 ### FillStatus
 A mapping of `RelayData` -> `FillStatus` is stored within each SpokePool instance. This mapping can be queried with the hashed `V3RelayData` for a deposit, allowing the status for the corresponding fill to be queried.
@@ -81,7 +106,7 @@ A V3RelayExecutionEventInfo instance is emitted with each `FilledV3Relay` event 
 
 | Name | Type | Description |
 | :--- |:---- | :---------- |
-| updatedRecipient | address | The recipient of the funds being transferred. This may be the `recipient` identified in the original deposit, or an updated `recipient` following a `RequestedSpeedUpV3Deposit` event. |
+| updatedRecipient | bytes32 | The recipient of the funds being transferred. This may be the `recipient` identified in the original deposit, or an updated `recipient` following a `RequestedSpeedUpV3Deposit` event. |
 | updatedOutputAmount | uint256 | The amount sent to `updatedRecipient` by the relayer completing the fill. |
 | updatedMessage | bytes | Data that is forwarded to the recipient as part of a relay. |
 | repaymentChainId | uint256 | The chain specified by the depositor for fill repayment. |
@@ -102,20 +127,50 @@ A V3SlowFill instance is emitted with each `FilledV3Relay` event (see below).
 The `updatedRecipient` field is normally set to the `recipient` from the corresponding `V3FundsDeposited` event. In the event that the relayer completes the fill with an accompanying `RequestedSpeedUpV3Deposit` event, `updatedRecipient` will be set to the address approved by the update.
 
 ## Events
-Across V3 defines the following new events:
+Across V3 defines the following events:
+- FundsDeposited
+- V3FundsDeposited
+- RequestedSpeedUpDeposit
+- RequestedSpeedUpV3Deposit
+- FilledRelay
+- FilledV3Relay
+- RequestedSlowFill
+- RequestedV3SlowFill
+- ClaimedRelayerRefund
+
+### Event Deprecation
+The following events are marked for future deprecation. Deprecation will be announced in a future update to this specification.
 - V3FundsDeposited
 - RequestedSpeedUpV3Deposit
-- RequestedV3SlowFill
 - FilledV3Relay
+- RequestedV3SlowFill
 
-### V3FundsDeposited
-The `V3FundsDeposited` event emits the unique `V3RelayData` for an individual deposit. No additional fields are defined. Consumers of this event should append the `originChainId` in order to avoid unintentionally mixing events from different chains.
+### FundsDeposited, V3FundsDeposited
+The `FundsDeposited` event emits the unique `V3RelayData` for an individual deposit. No additional fields are defined. 
+The `V3FundsDeposited` event emits the unique `V3RelayDataLegacy` for an individual deposit. No additional fields are defined.
+
+Consumers of these events should append the `originChainId` in order to avoid unintentionally mixing events from different chains.
 
 Note:
-- The `V3FundsDeposited` `outputToken` is not required to be a known HubPool `l1Token`. In-protocol arbitrary token swaps are technically supported by Across v3.
+- The `FundsDeposited` and `V3FundsDeposited` `outputToken` field is not required to be a known HubPool `l1Token`. In-protocol arbitrary token swaps are technically supported by Across v3.
+- The address identified by `exclusiveRelayer` has exclusive right to complete the relay on the destination chain until `exclusivityDeadline` has elapsed.
+- If `exclusivityDeadline` is set to a past timestamp, any address is eligible to fill the relay.
+- Any deposit that remains unfilled after the specified `fillDeadline` shall be refunded to the `depositor` address via the origin SpokePool in a subsequent settlement bundle.
 
-### RequestedSpeedUpV3Deposit
-The `RequestedSpeedUpV3Deposit` emits specifies the following fields:
+### RequestedSpeedUpDeposit, RequestedSpeedUpV3Deposit
+The `RequestedSpeedUpDeposit` event emits the following data.
+
+| Name | Type | Description |
+| :--- | :--- | :---------- |
+| depositId | uint256 | The depositId of the corresponding `FundsDeposited` event to be updated. |
+| depositor | bytes32 | The depositor of the corresponding `FundsDeposited` event to be updated. |
+| updatedOutputAmount | uint256 | The new outputAmount approved by the depositor. This should be _lower_ than the original deposit `outputAmount`. |
+| updatedRecipient | bytes32 | The new recipient to receive the funds. |
+| updatedMessage | bytes | The new message to be supplied to the recipient. |
+| depositorSignature | bytes | A signature by the depositor authorizing the above updated fields. |
+
+The `RequestedSpeedUpV3Deposit` event emits the following data:
+
 | Name | Type | Description |
 | :--- | :--- | :---------- |
 | depositId | uint32 | The depositId of the corresponding `V3FundsDeposited` event to be updated. |
@@ -126,28 +181,38 @@ The `RequestedSpeedUpV3Deposit` emits specifies the following fields:
 | depositorSignature | bytes | A signature by the depositor authorizing the above updated fields. |
 
 Note:
-- Relayers may optionally append the updated request from a `RequestedSpeedUpV3Deposit` when filling a relay, but have no obligation to use the updated request.
-- The address identified by `exclusiveRelayer` has exclusive right to complete the relay on the destination chain until `exclusivityDeadline` has elapsed.
-- If `exclusivityDeadline` is set to a past timestamp, any address is eligible to fill the relay.
-- Any attempted fill that occurs after `fillDeadline` has elapsed shall be rejected by the destination SpokePool. The corresponding deposit shall be refunded via the origin SpokePool in the relevant settlement bundle.
+- Relayers may optionally append the updated request from a `RequestedSpeedUpDeposit` or `RequestedSpeedUpV3Deposit` event when filling a relay, but have no obligation to use the updated request.
 
-### RequestedV3SlowFill
-The `RequestedV3SlowFill` event emits `V3RelayData`. This event is emitted on the destination chain and is intended to signal to proposers that a slow fill has been requested.
+### RequestedSlowFill, RequestedV3SlowFill
+The `RequestedSlowFill` event emits an `V3RelayData` instance.
+The `RequestedV3SlowFill` event emits an `V3RelayDataLegacy` instance.
+
+These events are emitted on the destination chain and signal to proposers that a slow fill has been requested for a specific deposit.
 
 Note:
-- `RequestedV3SlowFill` events cannot occur until the `exclusivityDeadline` timestamp has elapsed on the destination chain.
+- `RequestedV3SlowFill` events cannot be emitted until the `exclusivityDeadline` timestamp has elapsed on the destination chain.
 - `RequestedV3SlowFill` events cannot be emitted once the `fillDeadline` timestamp has elapsed on the destination chain.
 
-### FilledV3Relay
-The `FilledV3Relay` event extends the `V3RelayData` type by adding the following fields:
+### FilledRelay, FilledV3Relay
+
+The `FilledRelay` event extends the `V3RelayData` type by applying the following adjustments:
+| Name | Type | Description |
+| :--- | :--- | :---------- |
+| message | omitted | This field is omitted from the `FilledRelay` event in favour of the `messageHash` field. |
+| messageHash | bytes32 | The keccak256 hash of the `V3RelayData` message field. This field is included in place of the `V3RelayData` message field. See also [Computing RelayData hashes](#computing-relaydata-hashes). |
+| relayer | bytes32 | The address completing relay on the destination SpokePool. |
+| repaymentChainId | uint256 | The depositId of the corresponding `V3FundsDeposited` event to be updated. |
+| relayExecutionInfo | V3RelayExecutionEventInfo | The effective `recipient`, `message` and `outputAmount`, as well as the `FillType` performed (FastFill, ReplacedSlowFill, SlowFill). |
+
+The `FilledV3Relay` event extends the `V3RelayDataLegacy` type by adding the following fields:
 | Name | Type | Description |
 | :--- | :--- | :---------- |
 | relayer | address | The address completing relay on the destination SpokePool. |
 | repaymentChainId | uint256 | The depositId of the corresponding `V3FundsDeposited` event to be updated. |
-| relayExecutionInfo | V3RelayExecutionInfo | The effective `recipient`, `message` and `outputAmount`, as well as the `FillType` performed (FastFill, ReplacedSlowFill, SlowFill). |
+| relayExecutionInfo | V3RelayExecutionEventInfo | The effective `recipient`, `message` and `outputAmount`, as well as the `FillType` performed (FastFill, ReplacedSlowFill, SlowFill). |
 
 Note:
-- Consumers of this event should append the `destinationChainId` attribute in order to avoid unintentioanlly mixing events from different chains.
+- Consumers of these events should append the `destinationChainId` attribute in order to avoid unintentioanlly mixing events from different chains.
 
 # Root Bundle Proposals
 
@@ -197,17 +262,39 @@ Across v3 `SlowRelayLeaf` objects are defined by the `V3SlowFill` [data type](#d
 Note:
 - The format of Slow Relay leaves is updated from Across v2.
 
+## Definitions
+
+### Deposits
+A Deposit event is defined as an instance of either of the following events:
+- `FundsDeposited`.
+- `V3FundsDeposited`.
+
+### Fills
+A Fill event is defined as an instance of either of the following events:
+- `FilledRelay`.
+- `FilledV3Relay`.
+
+### Slow Fill Requests
+A Slow Fill event is defined as an instance of either of the following events:
+- `RequestedSlowFill`.
+- `RequestedV3SlowFill`.
+
+### RelayData
+RelayData is defined as an instance of etiher of the following data types:
+- `V3RelayData`.
+- `V3RelayDataLegacy`.
+
 ## Method
 ### Identifying SpokePool Contracts
 The current SpokePool address for a specific chain is available by querying `HubPool.crossChainContracts()`. The chainId must be specified in the query. In case of SpokePool migations, historical SpokePool addresses can be identified by scraping HubPool `CrossChainContractsSet` events.
 
-### Identifying a chains inclusion as a "Lite" Chain
+### Identifying "Lite" deployments
 We consider a deposit to "originate" or be "destined for" a "Lite chain" if the `LITE_CHAIN_ID_INDICES` value in the AcrossConfigStore includes the deposit's origin chain or destination chain respectively as of the deposit's `quoteTimestamp` field. These chains impose constraints on relayer repayments and slow fills.
 
 ### Resolving SpokePool tokens to their HubPool equivalent
 For the purpose of identifying the equivalent HubPool token given a SpokePool token, the following shall be followed:
 1. Find the latest `SetRebalanceRoute` event with a block timestamp at or before the relevant HubPool block number, where the relevant SpokePool chain ID and token address  match the `SetRebalanceRoute` `destinationChainId` and `destinationToken` fields.
-    - In the case of a `V3FundsDeposited` event, the relevant HubPool block number is identified by resolving the `quoteTimestamp` to a HubPool block number.
+    - In the case of a `Deposit` event, the relevant HubPool block number is identified by resolving the `quoteTimestamp` to a HubPool block number.
 3. From the resulting `SetRebalanceRoute` event, select the associated `l1Token` field.
 4. Search the `SetPoolRebalanceRoute` events for the same `l1Token` and `destinationChainId` at or before the applicable HubPool block number.
 5. Using the `l1Token` value found in step 2, search for the latest `SetRebalanceRoute` event at or before the applicable HubPool block number where `l1Token` and `destinationChainId` that match the extracted `l1Token` and SpokePool chain ID. If a match is found, the addresses match and are considered cross-chain equivalents.
@@ -218,15 +305,26 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
 - A "soft pause" of a chain is permitted in the event that the proposer cannot safely increment the bundle block range, or has no events to propose beyond the previous bundle block range. In this case, the proposer may repeat the procedure for
   DISABLED_CHAINS by proposing from and to the previous bundle end block.
 
-### Finding Valid Relays
-For the purpose of computing relayer repayments, each `FilledV3Relay` event emitted within the target block range on a destination SpokePool shall be considered valid by verifying that:
-1. The `FilledV3Relay` `FillType` field is not set to `SlowFill`,
-2. The component `V3RelayData` maps exactly to a corresponding `V3FundsDeposited` event emitted on the relevant `originChainId`. This may be compared by comparing the hashes of the two objects.
+### Computing RelayData Hashes
+#### V3RelayData
+The `V3RelayData` hash is computed as the `keccak256` hash over the ABI-encoded representation of the arguments `relayData`, `destinationChainId`, where:
+- `relayData` is of type `V3RelayData`.
+- `destinationChainId` is of type `uint256`.
 
-If the `V3FundsDeposited` event specifies `outputToken` 0x0 (i.e. the Zero Address), the equivalent SpokePool token on the destination chain shall be substituted in. For the purpose of determining `V3RelayData` equivalency, the updated/substituted `outputToken` shall be used in place of 0x0.
+#### V3RelayDataLegacy
+The `V3RelayDataLegacy` hash is computed as the `keccak256` hash over the ABI-encoded representation of the arguments `relayData`, `destinationChainId`, where:
+- `relayData` is of type `V3RelayDataLegacy`
+- `destinationChainId` is of type `uint256`.
+
+### Finding Valid Relays
+For the purpose of computing relayer repayments, each of the `Fill` events emitted within the target block range on a destination SpokePool shall be considered valid by verifying that:
+1. The `Fill` event `FillType` field is not set to `SlowFill`,
+2. The component `RelayData` maps exactly to a corresponding `Deposit` event emitted on the relevant `originChainId`. This may be compared by comparing the hashes of the two objects.
+
+If the `Deposit` event specifies `outputToken` 0x0 (i.e. the Zero Address), the equivalent SpokePool token on the destination chain shall be substituted in. For the purpose of determining `RelayData` equivalency, the updated/substituted `outputToken` shall be used in place of 0x0.
 
 ### Finding Expired Deposits
-For the purpose of computing depositor refunds, each `V3FundsDeposited` event shall be considered expired by verifying that:
+For the purpose of computing depositor refunds, each `Deposit` event shall be considered expired by verifying that:
 1. The `fillDeadline` timestamp elapsed within the target block range on the destination SpokePool (i.e. the `fillDeadline` expired between the `block.timestamp` of the destination chain's bundle start and end block),
 2. The `FillStatus` on the destination SpokePool is set to `Unfilled` or `SlowFillRequested`.
 
@@ -236,49 +334,50 @@ Note:
 - The `fillDeadline` timestamp shall be resolved to a block number on the destination chain in order to determine inclusion within the target block range.
 
 ### Finding Slow Fill Requests
-For the purpose of computing slow fills to be issued to recipients, each `RequestedV3SlowFill` event emitted within the target block range on a destination SpokePool shall be considered valid by verifying that:
+For the purpose of computing slow fills to be issued to recipients, each `Slow Fill Request` emitted within the target block range on a destination SpokePool shall be considered valid by verifying that:
 1. The `inputToken` and `outputToken` addresses are equivalent at the deposit `quoteTimestamp`,
 2. The `fillDeadline` has not already elapsed relative to the `destinationChainId` bundle end block number,
-3. The destination SpokePool `FillStatus` mapping for the relevant `V3RelayData` hash is `SlowFillRequested`,
-4. The `RequestedV3SlowFill` `V3RelayData` is matched by a corresponding `V3FundsDeposited` event on the origin SpokePool,
+3. The destination SpokePool `FillStatus` mapping for the relevant `RelayData` hash is `SlowFillRequested`,
+4. The `Slow Fill Request` `RelayData` is matched by a corresponding `Deposit` event on the origin SpokePool,
 5. The `originChainId` and `destinationChainId` are not Lite chains.
 
 Note:
-- A slow fill request is made by supplying a complete copy of the relevant `V3RelayData` emitted by a `V3FundsDeposited` event.
-- The resulting set of validated `RequestedV3SlowFill` events shall be included as SlowFills in the subsequent root bundle proposal.
+- A slow fill request is made by supplying a complete copy of the relevant `RelayData` emitted by a `Deposit` event.
+- The resulting set of validated `Slow Fill Requests` shall be included as SlowFills in the subsequent root bundle proposal.
 
 ### Computing LP Fees
-Each valid `FilledV3Relay` event is subject to an LP fee. The procedure for computing LP fees is as defined in [UMIP-136 Add IS_RELAY_VALID as a supported price identifier](https://github.com/UMAprotocol/UMIPs/blob/7b1a046098d3e2583abd0372c5e9c6003b46ad92/UMIPs/umip-136.md), with the following amendments:
+Each valid `Fill` event is subject to an LP fee. The procedure for computing LP fees is as defined in [UMIP-136 Add IS_RELAY_VALID as a supported price identifier](https://github.com/UMAprotocol/UMIPs/blob/7b1a046098d3e2583abd0372c5e9c6003b46ad92/UMIPs/umip-136.md), with the following amendments:
 - The AcrossConfigStore contract shall be used to identify the correct rate model, instead of a `RateModelStore` contract.
 - The `HubPool` `liquidityUtilizationCurrent()` and `liquidityUtilizationPostRelay()` functions shall be used instead of the `BridgePool` variant.
 - The event `inputToken` shall be mapped from the SpokePool address to a HubPool `l1Token` address by following the matching procedure outlined above.
-- The LP fee is computed between the `originChainId` and `FilledV3Relay.repaymentChainId` where the `relayExecutionInfo.FillType != SlowFill` and `FilledV3Relay.destinationChainId` otherwise.
+- The LP fee is computed between the `originChainId` specified by the `Deposit` and `repaymentChainId` specified by the relayer, where the `relayExecutionInfo.FillType != SlowFill` and the Fill `destinationChainId` otherwise.
 
 Note:
-- The LP fee is typically referenced as a multiplier of the `V3FundsDeposited` `inputAmount`, named `realizedLpFeePct` elsewhere in this document.
+- The LP fee is typically referenced as a multiplier of the `Deposit` `inputAmount`, named `realizedLpFeePct` elsewhere in this document.
 
 ### Computing Bundle LP Fees
 The bundle LP fee for a target block range on a SpokePool and token pair shall be determined by summing the applicable LP fees for each of the following validated events:
-- `FilledV3Relay`
+- `FilledRelay`.
+- `FilledV3Relay`.
 
 ### Computing Relayer Repayments
-For a validated `FilledV3Relay` event, the relayer repayment amount shall be computed as follows:
-- `(inputAmount * (1 - realizedLpFeePct)) / 1e18`, where `realizedLpFeePct` is computed over the set of HubPool `l1Token`, `originChainId` and `repaymentChainId` at the HubPool block number corresponding to the relevant `V3FundsDeposited` `quoteTimestamp`.
+For a validated `Deposit` event, the relayer repayment amount shall be computed as follows:
+- `(inputAmount * (1 - realizedLpFeePct)) / 1e18`, where `realizedLpFeePct` is computed over the set of HubPool `l1Token`, `originChainId` and `repaymentChainId` at the HubPool block number corresponding to the relevant `Deposit` `quoteTimestamp`.
 - The applicable rate model shall be sourced from the AcrossConfigStore contract for the relevant `l1Token`.
 - Deposits where the `originChainId` is a "Lite" chain in the AcrossConfigStore as of the `quoteTimestamp` shall always be repaid on the deposit's origin chain. This means the protocol overrides the relayer's requested `repaymentChainId` with the `originChainId` instead.
 
 ### Computing Deposit Refunds
-For an expired `V3FundsDeposited` event, the depositor refund amount shall be computed as `inputAmount` units of `inputToken`.
+For an expired `Deposit` event, the depositor refund amount shall be computed as `inputAmount` units of `inputToken`.
 
 ### Computing Slow Fill updated output amounts
 For the purpose of computing the amount to issue to a recipient for a SlowFill, the relayer fee shall be nulled by applying the following procedure:
 - `updatedOutputAmount = (inputAmount * (1 - realizedLpFeePct)) / 1e18`, where `realizedLpFeePct` is computed at the deposit `quoteTimestamp` between `originChainId` and `destinationChainId`.
 
 Constraint:
-- The `V3FundsDeposited` `outputAmount` shall _not_ be considered when determining SlowFill amounts.
+- The `Deposit` `outputAmount` shall _not_ be considered when determining SlowFill amounts.
 
 Note:
-- The `V3FundsDeposited` `outputAmount` specifies the exact amount to be received by the `recipient` for a standard fill, and is therefore exclusive of any relayer or LP fees paid.
+- The `Deposit` `outputAmount` specifies the exact amount to be received by the `recipient` for a standard fill, and is therefore exclusive of any relayer or LP fees paid.
 
 ### Finding the Opening Running Balance
 The Opening Running Balance is defined as the cumulative running balance as at the previous successful (undisputed) Root Bundle Proposal.
@@ -295,17 +394,17 @@ The procedure for computing running balances for an `l1Token` and `chainId` pair
 1. Initialize the running balance to 0.
 
 2. Add relayer refunds:
-    - For each group of validated `FilledV3Relay` events, initialize a running balance at 0 and add the add the relayer repayment.
+    - For each group of validated `Fill` events, initialize a running balance at 0 and add the add the relayer repayment.
 
 3. Add deposit refunds:
-    - For each group of `V3FundsDeposited` events that expired within the target block range, sum the total deposit refunds on the origin chain. Add the amount to the exsting relayer refunds for that chain.
+    - For each group of `Deposit` events that expired within the target block range, sum the total deposit refunds on the origin chain. Add the amount to the exsting relayer refunds for that chain.
 
 4. Add slow fills:
-    - For each group of validated `RequestedV3SlowFill` events, add each slow relay's `updatedOutputAmount` to the group's running balance.
+    - For each group of validated `Slow Fill Requests`, add each slow relay's `updatedOutputAmount` to the group's running balance.
 
 5. Subtract excesses from unexecuted slow fills:
-    - For each group of validated `FilledV3Relay` events where the `FillType` is `ReplacedSlowFill` and where there is no valid `RequestedV3SlowFill` event with an identical relay data hash in the current bundle data, subtract the SlowFill `updatedOutputAmount` from the running balance in recognition that the SlowFill will never be executed because the fill amount has already been transferred.
-    - For each expired deposit refund identified above where the `FillStatus` on the deposit destination chain for the deposit's relay data is `RequestedSlowFill` and the matching slow fill request is not in the current bundle range, subtract the associated SlowFill `updatedOutputAmount` from the running balance in recognition that the SlowFill cannot be executed past the `fillDeadline`.
+    - For each group of validated `Fills` where the `FillType` is `ReplacedSlowFill` and where there is no valid `Slow Fill Request` with an identical relay data hash in the current bundle data, subtract the SlowFill `updatedOutputAmount` from the running balance in recognition that the SlowFill will never be executed because the fill amount has already been transferred.
+    - For each expired deposit refund identified above where the `FillStatus` on the deposit destination chain is `RequestedSlowFill` and the matching slow fill request is not in the current bundle range, subtract the associated SlowFill `updatedOutputAmount` from the running balance in recognition that the SlowFill cannot be executed past the `fillDeadline`.
 
 6. Add the Opening Running Balance for the selected `l1Token` and `chainId` pair.
 
@@ -331,7 +430,7 @@ The referenced `SpokeTargetBalances` is as specified by [UMIP-157 Token Constant
 ## Constructing Root Bundles
 
 ### Constructing the Pool Rebalance Root
-One Pool Rebalance Leaf shall be produced per unique `chainId` & `l1Token` pair, where the corresponding SpokePool emitted `V3FundsDeposited`, `FilledV3Relay` or`RequestedV3SlowFill` events within the target block range.
+One Pool Rebalance Leaf shall be produced per unique `chainId` & `l1Token` pair, where the corresponding `Deposit`, `Fill` or`Slow Fill Request` events were emitted by the relevant SpokePool within the target block range.
 
 Each Pool Rebalance Leaf shall be constructed as follows:
 1. For each unique `chainId` and `l1Token` pair:
@@ -361,8 +460,8 @@ Note:
 
 ### Constructing the Relayer Refund Root
 At least one Relayer Refund Leaf shall be produced for each unique combination of SpokePool and `l1Token` for any of the following conditions:
-- Valid `FilledV3Relay` events, OR
-- Expired `V3Fundsdeposited` events, OR
+- Valid `Fill` events, OR
+- Expired `Deposit` events, OR
 - A negative running balance net send amount.
 
 Each Relayer Refund Leaf shall be constructed as follows:
@@ -395,13 +494,13 @@ Note:
 - Once these leaves are constructed, they can be used to form a merkle root as described in the previous section.
 
 ### Constructing the Slow Relay Root
-One Slow Relay Leaf shall be produced per valid `RequestedV3SlowFill` event emitted within the target block range for a destination SpokePool.
+One Slow Relay Leaf shall be produced per valid `Slow Fill Request` emitted within the target block range for a destination SpokePool.
 
-A Slow Relay Leaf shall not be produced if the `RequestedV3SlowFill` event's `inputAmount` is equal to 0 and the `message` is a zero bytes string.
+A Slow Relay Leaf shall not be produced if the relevant `Slow Fill Request` `inputAmount` is equal to 0 and the `message` is a zero bytes string.
 
 Each Slow Relay Leaf shall be constructed as follows:
-1. Set `relayData` to the `V3RelayData` data emitted by the validated `RequestedV3SlowFill` event.
-2. Set `chainId` to `destinationChainId` from the corresponding validated `RequestedV3SlowFill` event.
+1. Set `relayData` to the `RelayData` emitted by the validated `Slow Fill Request`.
+2. Set `chainId` to `destinationChainId` from the corresponding validated `Slow Fill Request`.
 3. Set `updatedOutputAmount` to the updated amount computed for the SlowFill.
 
 The array of Slow Relay Leaf instances shall be sorted according to;
@@ -410,7 +509,7 @@ The array of Slow Relay Leaf instances shall be sorted according to;
 
 Note:
 - Once these leaves are constructed, they can be used to form a merkle root as described in the previous section.
-- Deposits with disparate output tokens (i.e. where the outputToken is not the equivalent of inputToken on the destination chain) are explicitly not eligible for slow fills. Any instances of `RequestedV3SlowFill` events for non-equivalent tokens shall be ignored.
+- Deposits with disparate output tokens (i.e. where the outputToken is not the equivalent of inputToken on the destination chain) are explicitly not eligible for slow fills. Any instances of `Slow Fill Requests` for non-equivalent tokens shall be ignored.
 
 # Recommendations
 - Proposers are responsible for detecting and mitigating incorrect or inconsistent RPC data. Proposers should take steps to validate the correctness of their RPC data before proposing.
@@ -421,6 +520,8 @@ Note:
 Across v3 is supplementary to Across v2, adding extra logic support new types of deposits, fills and slow fill requests. At the point of upgrading to Across v3, it will no longer be possible for Across v2 `FundsDeposited` events to be emitted. In order to support continuity of service and to minimise disruption for third-party integrators, it will be possible for pre-existing `FundsDeposited` events to be filled via `FilledRelay` events. During this period, RootBundleProposals will contain relayer repayment and slow fill leaves for both Across v2 and v3.
 
 The V3 rules defined in this UMIP will apply beginning when the VERSION field in the ConfigStore is updated to 3 or higher. The ability for Across bundles to support V2 events may cease in a future VERSION increment.
+
+XXX Identify a transition point for the pending upgrade. This is likely to be handled via a `VERSION` bump.
 
 # Implementation
 The Across v3 implementation is available in the Across [contracts-v2](https://github.com/across-protocol/contracts-v2) repository.


UMIPs/umip-179.md
```diff
@@ -422,6 +521,8 @@ Across v3 is supplementary to Across v2, adding extra logic support new types of
 
 The V3 rules defined in this UMIP will apply beginning when the VERSION field in the ConfigStore is updated to 3 or higher. The ability for Across bundles to support V2 events may cease in a future VERSION increment.
```
pxrl, line 522: Question for the brains trust - do we want to retain the description of the `Across v2 -> v3 Transition`? The migration occurred almost 12 months ago so it's only here for historical interest. We don't really have any formal versioning for `UMIP-179` though, so it's unclear for me whether it's moreso a living document reflecting the current protocol spec, or a record of the protocol evolution.

UMIPs/umip-179.md
```diff
@@ -218,15 +305,48 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
 
-### Finding Expired Deposits
-For the purpose of computing depositor refunds, each `V3FundsDeposited` event shall be considered expired by verifying that:
+#### Validating Pre-fills
+For each of the `Deposits` emitted within the target block range where no corresponding `Fill` is identified on the destination chain, identify the valid `Fill` according to the following criteria:
+1. Verify that the destination chain `FillStatus` for the `Deposit` `RelayData` is `Filled` as at the destination chain end block number for the proposal.
+2. Resolve the corresponding `Fill` on the destination chain.
+
+#### Note
+- No specific method is prescribed for resolving the fill on the destination chain. An `eth_getLogs` request can facilitate this, and if required, the target block range could be narrowed by a binary search over the `FillStatus` field. This is left as an implementation decision.
```
pxrl, line 341: I added this section to describe how to identify pre-fills. This is a fairly abstract description; it arises out of the realisation that an arbitrarily-old fill with a non-deterministic depositId can still be identified via a binary search, which should remove the need to artificially limit the fill->deposit time delta. This should be naively fine OK as long as SpokePool addresses and token mappings don't change. We might consider adding a note to explicitly warn relayers of these risks.

fwiw, the SDK already has an implementation of the necessary logic for a binary search here: https://github.com/across-protocol/sdk/blob/db244fbaec15dc049452cd104ca7bec5856ad3da/src/utils/SpokeUtils.ts#L328 This implementation came out of the resilience work that was done, because RPC provider jank can also cause a deposit without a corresponding fill event.

There is a consideration about the load imposed on proposers and disputers from a binary search, which can only be done per-deposit. I think this is the optimal strategy for now, where only a small number of deposits are expected to fall outside of the proposer/disputer lookback window. The strategy could be improved if the ratio of pre-fills started to become significant. One candidate for optimising this could be to just increase the block range when querying for fills. Since _most_ missing fills are probably only hours old, this strategy could be used quite effectively.

If the load from increasing the getLogs range did ultimately produce a scalability problem then I think we'd ultimately look to bound the fill -> deposit delta by some multiple of the `fillDeadline` specified in the deposit. But I think that should be reserved for a future UMIP change (and may not even be necessary).

UMIPs/umip-179.md
```diff
@@ -39,25 +39,50 @@ All global constants from UMIP-157 will be retained for use in Across v3. The gl
...
+| depositor | bytes32 | The address that made the deposit on the origin chain. |
+| recipient | bytes32 | The recipient address on the destination chain. |
+| exclusiveRelayer | bytes32 | The optional exclusive relayer who can fill the deposit before the exclusivity deadline. |
+| inputToken | bytes32 | The token that is deposited on the origin chain by the depositor. |
+| outputToken | bytes32 | The token that is received on the destination chain by the recipient. |
 | inputAmount | uint256 | The amount of inputToken that is deposited by the depositor. |
 | outputAmount | uint256 | The amount of outputToken that is received by the recipient. |
 | originChainId | uint256 | The chain ID of the origin SpokePool. |
-| depositId | uint32 | The ID uniquely identifying a deposit on the origin chain. |
+| depositId | uint256 | The ID identifying a deposit on the origin chain. |
```
nicholaspai, line 56: Should this umip state that all ids are unique below uint32.max - 1? Or is that an implementation detail you think?

UMIPs/umip-179.md
```diff
@@ -102,20 +127,50 @@ A V3SlowFill instance is emitted with each `FilledV3Relay` event (see below).
...
+The `FundsDeposited` event emits the unique `V3RelayData` for an individual deposit. No additional fields are defined. 
+The `V3FundsDeposited` event emits the unique `V3RelayDataLegacy` for an individual deposit. No additional fields are defined.
 
-### V3FundsDeposited
-The `V3FundsDeposited` event emits the unique `V3RelayData` for an individual deposit. No additional fields are defined. Consumers of this event should append the `originChainId` in order to avoid unintentionally mixing events from different chains.
+Consumers of these events should append the `originChainId` in order to avoid unintentionally mixing events from different chains.
 
 Note:
-- The `V3FundsDeposited` `outputToken` is not required to be a known HubPool `l1Token`. In-protocol arbitrary token swaps are technically supported by Across v3.
+- The `FundsDeposited` and `V3FundsDeposited` `outputToken` field is not required to be a known HubPool `l1Token`. In-protocol arbitrary token swaps are technically supported by Across v3.
```
nicholaspai, line 155:
```suggestion
- The `FundsDeposited` and `V3FundsDeposited` `outputToken` field is not required to be a known HubPool `l1Token`. In-protocol arbitrary token swaps are technically supported by Across v3.
```
```suggestion
- The `FundsDeposited` and `V3FundsDeposited` `outputToken` field is not required to be a known HubPool `l1Token`. In-protocol arbitrary token swaps are supported by Across v3.
```

Think we can be even more explicit here

UMIPs/umip-179.md
```diff
@@ -102,20 +127,50 @@ A V3SlowFill instance is emitted with each `FilledV3Relay` event (see below).
...
+- The address identified by `exclusiveRelayer` has exclusive right to complete the relay on the destination chain until `exclusivityDeadline` has elapsed.
+- If `exclusivityDeadline` is set to a past timestamp, any address is eligible to fill the relay.
+- Any deposit that remains unfilled after the specified `fillDeadline` shall be refunded to the `depositor` address via the origin SpokePool in a subsequent settlement bundle.
```
nicholaspai, line 158: What about if the deposit is a duplicate relay hash? Which is now possible with non uniqueness of deposit ids

UMIPs/umip-179.md
```diff
@@ -218,15 +305,44 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+- Finding Expired Deposits
 
-### Finding Expired Deposits
-For the purpose of computing depositor refunds, each `V3FundsDeposited` event shall be considered expired by verifying that:
+#### Note
+- Depositor refunds are issued via the Relayer Repayment workflow.
+
+#### Validating Fills
+Each of the `Fills` emitted within the target block range on a destination SpokePool shall be considered valid by verifying that:
+1. The `Fill` event `FillType` field is not set to `SlowFill`, AND
```
nicholaspai, line 330: All SlowFill fills are valid by the way, not sure if that's stated elsewhere but this bullet point would lead me to believe otherwise

UMIPs/umip-179.md
```diff
@@ -218,15 +305,44 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+2. The component `RelayData` maps exactly to a corresponding `Deposit` event emitted on the relevant `originChainId`. This may be determined by comparing the hashes of the two objects, AND
+3. The corresponding `Deposit` event occurred within or before the target block range on the origin chain SpokePool.
```
nicholaspai, line 332: Does this have to be true? What is the "target" block range? Could we remove this bullet point and the. Use this section to also validate pre fills?

I see it as the responsibility of the validator/proposer to use a long enough lookback to validate all pre fills and normal fills. 

This lookback is finite for non legacy deposits with non infinite fill deadlines

UMIPs/umip-179.md
```diff
@@ -236,49 +352,58 @@ Note:
...
+For a validated `Deposit` event, the relayer repayment amount shall be computed as follows:
+- `(inputAmount * (1 - realizedLpFeePct)) / 1e18`, where `realizedLpFeePct` is computed over the set of HubPool `l1Token`, `originChainId` and `repaymentChainId` at the HubPool block number corresponding to the relevant `Deposit` `quoteTimestamp`.
 - The applicable rate model shall be sourced from the AcrossConfigStore contract for the relevant `l1Token`.
 - Deposits where the `originChainId` is a "Lite" chain in the AcrossConfigStore as of the `quoteTimestamp` shall always be repaid on the deposit's origin chain. This means the protocol overrides the relayer's requested `repaymentChainId` with the `originChainId` instead.
 
+The recipient of relayer repayments shall be the `Fill` `relayer` address. If the `relayer` address is invalid for the respective `repaymentChainId`, the proposer shall:
+- Issue the refund on the `destinationChainId` instead of the `repaymentChainId`, AND
```
nicholaspai, line 388: What if the destination chain is a lite chain?

UMIPs/umip-179.md
nicholaspai: I think we can remove it since the git history will have the migration details if we want to refer back to it

UMIPs/umip-179.md
```diff
@@ -422,11 +547,10 @@ Across v3 is supplementary to Across v2, adding extra logic support new types of
 
 The V3 rules defined in this UMIP will apply beginning when the VERSION field in the ConfigStore is updated to 3 or higher. The ability for Across bundles to support V2 events may cease in a future VERSION increment.
 
+XXX Identify a transition point for the pending upgrade. This is likely to be handled via a `VERSION` bump.
```
nicholaspai, line 550: Yep agreed

nicholaspai (commented):
This is a fantastic start. My biggest questions at the moment are in the new validating fills section

UMIPs/umip-179.md
pxrl: I went back and forth on this a bit, and I also wondered whether we should specify that pre-fills must have a depositId > 2³²-1. So far I haven't found any concrete need mention it. I'm probably inclined to leave it unspecified unless we find a real need for it. wdyt?

UMIPs/umip-179.md
pxrl: Yep - fair call :+1:

UMIPs/umip-179.md
pxrl: Just to tease this one out into a scenario: `unsafeDeposit()` is called, producing a `relayDataHash` collision on the destination SpokePool. The proposer concludes that the deposit was filled by some previous, unrelated fill. The proposer then identifies that fill and credits the relayer with the repayment, meaning the relayer gets paid twice and the depositor can never receive their refund. Is this the same scenario you're thinking of?

In this specific case, I think there is some risk that the depositor (or relayer, or more broadly the "facilitator" of the order flow) takes when they call `unsafeDeposit()` (this is basically why it's called `unsafe`). It is however possible for them to check the `FillStatus` on the destination chain in advance of selecting the salt for their `depositId`, so they can avoid the scenario where they collide with a very old fill. I think this then reduces the risk to a collision occurring within a much shorter period of time - i.e. seconds of minutes.

This might not be the only scenario, and my reasoning might be flawed on this. This scenario alone also indicates that we need to add some guidance in our docs for how integrators should select their salt when generating pre-fill deposits for depositors to sign. wdyt?

pxrl: One additional and very simple check that could protect against this - the proposer and disputer could be required to verify both the relayDataHash and a subset of the fields - i.e. to verify that some or all of the following fields are an exact match. That would protect against relayDataHash collisions, and would allow the proposer to refund the deposit on the origin chain:
- `depositor`
- `recipient`
- `inputToken`
- `inputAmount`
- `outputAmount`
- `fillDeadline`

UMIPs/umip-179.md
pxrl: Yeah; correct that all slow fills are valid. The intention of this text is to identify that the proposer/disputer should not bother validating fills where the `FillType` is `SlowFill`, because the aim is really to identify fills for repayment. I'll add some context to indicate that it's for repayment only.

UMIPs/umip-179.md
pxrl: > What is the "target" block range?

It's the block range for the current proposal. I'll rename to `Bundle Block Range` and will link to the section on `Identifying Bundle Block Ranges`.

> Does this have to be true? Could we remove this bullet point and the. Use this section to also validate pre fills?

It's a new constraint imposed by the infinite lookback on pre-fills. The scenario that we must protect against is:
In the current proposal:
- Fill is made within the current bundle block range.
- Deposit somehow lands slightly outside of the current bundle block range.
- Proposer matches the fill with the deposit and issues a repayment to the relayer.

In the subsequent proposal:
- Deposit appears within the current bundle block range.
- No fill is identified because it fell into the previous bundle block range.
- Proposer thinks this was a legitimate pre-fill and issues another repayment for it.

The proposed solution is to only permit the initial repayment for deposits that appear within the current bundle block range. So in the scenario above, the initial repayment would be suppressed, and it would only occur in the subsequent bundle. This avoids double repayment to the relayer, but imposes a marginal repayment delay on that single repayment.

UMIPs/umip-179.md
pxrl: I think it's OK when the destination chain is a lite chain (we permit this currently), but it's not OK when the origin chain is a lite chain (good catch). So the logic needs to be tightened up, such that this has a lower precedence than the lite chain constraints.

UMIPs/umip-179.md
```diff
@@ -218,67 +305,105 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
-2. The component `V3RelayData` maps exactly to a corresponding `V3FundsDeposited` event emitted on the relevant `originChainId`. This may be compared by comparing the hashes of the two objects.
-
-If the `V3FundsDeposited` event specifies `outputToken` 0x0 (i.e. the Zero Address), the equivalent SpokePool token on the destination chain shall be substituted in. For the purpose of determining `V3RelayData` equivalency, the updated/substituted `outputToken` shall be used in place of 0x0.
-
-### Finding Expired Deposits
-For the purpose of computing depositor refunds, each `V3FundsDeposited` event shall be considered expired by verifying that:
-1. The `fillDeadline` timestamp elapsed within the target block range on the destination SpokePool (i.e. the `fillDeadline` expired between the `block.timestamp` of the destination chain's bundle start and end block),
+### Computing RelayData Hashes
+#### V3RelayData
+The `V3RelayData` hash is computed as the `keccak256` hash over the ABI-encoded representation of the arguments `relayData`, `destinationChainId`, where:
```
Reinis-FRP, line 310: Note that the upcoming spoke deployments will have modified `_getV3RelayHash` so that `relayData.message` is converted to `bytes32` (either `bytes32(0)` for empy message or `keccak256` hash of message bytes).

UMIPs/umip-179.md
pxrl: @nicholaspai Updated here to be more explicit about the precedence of the applied `repaymentChainId` and reverting to a fallback repayment address in case of an invalid relayer-specified address for the applied `repaymentChainId`:
https://github.com/UMAprotocol/UMIPs/pull/611/commits/a95d43744eaff6c39302c4f4c7d674ba63e50443

wdyt?

UMIPs/umip-179.md
```diff
@@ -218,67 +305,108 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+For a validated `Deposit` event, the relayer repayment amount shall be computed as follows:
+- `(inputAmount * (1 - realizedLpFeePct)) / 1e18`, where `realizedLpFeePct` is computed over the set of HubPool `l1Token`, `originChainId` and `repaymentChainId` at the HubPool block number corresponding to the relevant `Deposit` `quoteTimestamp`.
 - The applicable rate model shall be sourced from the AcrossConfigStore contract for the relevant `l1Token`.
-- Deposits where the `originChainId` is a "Lite" chain in the AcrossConfigStore as of the `quoteTimestamp` shall always be repaid on the deposit's origin chain. This means the protocol overrides the relayer's requested `repaymentChainId` with the `originChainId` instead.
+
+The applied `repaymentChainId` shall be determined as follows:
+- When the `originChainId` is a `Lite chain` as at the `Deposit` `quoteTimestamp`: `originChainId`, ELSE:
+- The `repaymentChainId` as specified in the `Fill`.
```
pxrl, line 388: This bumps up against an [existing PR](https://github.com/UMAprotocol/UMIPs/pull/602) that I've had open for a while, which aims to be explicit about how to handle an invalid repaymentChainId for a given `inputToken` (i.e. what do we do when there's no `PoolRebalanceRoute` to `repaymentChainId` for a given `l1Token`).

The current format here should make it easy enough for us to be explicit about what happens when no `PoolRebalanceRoute` is available. In that case I assume we revert to repayment on `destinationChainId`.

UMIPs/umip-179.md
pxrl: @Reinis-FRP Good catch; I'd overlooked that. We did also bump into an issue around the upgrade that could permit double-fills due to the change in the way `RelayData` hashes are produced. We've floated the idea of reverting the change to the way `RelayData` is hashed on the EVM side (see [PR](https://github.com/across-protocol/contracts/pull/867/files) by @bmzig). This is a point we'll need to align on; I'll leave this comment open so we circle back and fix it once we agree on how to proceed.

UMIPs/umip-179.md
pxrl: Initial wording added here: https://github.com/UMAprotocol/UMIPs/pull/611/commits/fde6509690a9932a746a0b9b943b0f98a7cbb5b6

UMIPs/umip-179.md
nicholaspai: I think it could be a useful constraint to force this into the protocol. If we specify that all pre-fills have a deposit Id > some number, then the validator's logic for searching for deposits for pre-fills can be constrained to only look for deposits with finite fill deadlines (because the only way to send an infinite fill deadline deposit is to use the legacy deposit, which doesn't support deterministic deposit ID's)

UMIPs/umip-179.md
nicholaspai: > The proposer then identifies that fill and credits the relayer with the repayment, meaning the relayer gets paid twice and the depositor can never receive their refund. Is this the same scenario you're thinking of?

No, in this case the relayer could only fill once because the relay hash is the same for both deposits. The deposit pays twice and only gets filled once. The relayer gets refunded for one their one fill

But overall I think the function is aptly named "unsafe" and depositors who do this should be aware they might lose their deposit.

UMIPs/umip-179.md
```diff
@@ -382,11 +382,14 @@ The bundle LP fee for a [Bundle Block Range](#identifying-bundle-block-ranges) o
...
 - `(inputAmount * (1 - realizedLpFeePct)) / 1e18`, where `realizedLpFeePct` is computed over the set of HubPool `l1Token`, `originChainId` and `repaymentChainId` at the HubPool block number corresponding to the relevant `Deposit` `quoteTimestamp`.
 - The applicable rate model shall be sourced from the AcrossConfigStore contract for the relevant `l1Token`.
-- Deposits where the `originChainId` is a "Lite" chain in the AcrossConfigStore as of the `quoteTimestamp` shall always be repaid on the deposit's origin chain. This means the protocol overrides the relayer's requested `repaymentChainId` with the `originChainId` instead.
 
-The recipient of relayer repayments shall be the `Fill` `relayer` address. If the `relayer` address is invalid for the respective `repaymentChainId`, the proposer shall:
-- Issue the refund on the `destinationChainId` instead of the `repaymentChainId`, AND
-- Issue the refund to the `msg.sender` address, instead of the `relayer` address.
+The applied `repaymentChainId` shall be determined as follows:
+- When the `originChainId` is a `Lite chain` as at the `Deposit` `quoteTimestamp`: `originChainId`, ELSE:
```
nicholaspai: This creates an issue where we don't know a valid address for the filler on the origin chain, the same way we can always know the msg.sender on the destination chain as the filler

UMIPs/umip-179.md
nicholaspai: left one comment https://github.com/UMAprotocol/UMIPs/pull/611/commits/a95d43744eaff6c39302c4f4c7d674ba63e50443#r1923960561

UMIPs/umip-179.md
nicholaspai: This scenario likely won't exist today:

> Deposit appears within the current bundle block range.
> No fill is identified because it fell into the previous bundle block range.
> Proposer thinks this was a legitimate pre-fill and issues another repayment for it.

Because, on step 2, fills are still evaluated that are older than the current bundle block range, they're just not added to the refund leaf. The only change we might need to make is to ensure that the fill lookback is sufficiently long. 

Today we make sure that the lookback for all chains is longer than bundleStartBlock - MAX_FILL_DEADLINE (converted to blocks on origin chain) to make sure we see all expired deposits for a bundle. This SHOULD cover most pre-fills also but we probably should be more explicit. We could make a rule perhaps like: you cannot send a prefill more than MAX_FILL_DEADLINE time earlier than the deposit otherwise you might not get refunded

UMIPs/umip-179.md
```diff
@@ -218,67 +305,109 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+- When the `Fill` `relayer` address is valid for the applied `repaymentChainId`: `relayer`, ELSE
+- The `Fill` `msg.sender` address.
+
+#### Note
+- Examples of an invalid `relayer` address include:
+    - An SVM address on an EVM chain.
 
 ### Computing Deposit Refunds
-For an expired `V3FundsDeposited` event, the depositor refund amount shall be computed as `inputAmount` units of `inputToken`.
+For an expired `Deposit` event, the depositor refund amount shall be computed as `inputAmount` units of `inputToken`.
```
bmzig, line 400: I think you should also say that if you set the `depositor` field to an invalid EVM address, (i.e. a `bytes32` address), then you won't be refunded if you aren't filled.

UMIPs/umip-179.md
```diff
@@ -102,20 +127,50 @@ A V3SlowFill instance is emitted with each `FilledV3Relay` event (see below).
...
+- V3FundsDeposited
+- RequestedSpeedUpV3Deposit
 - FilledV3Relay
+- RequestedV3SlowFill
 
-### V3FundsDeposited
-The `V3FundsDeposited` event emits the unique `V3RelayData` for an individual deposit. No additional fields are defined. Consumers of this event should append the `originChainId` in order to avoid unintentionally mixing events from different chains.
+### FundsDeposited, V3FundsDeposited
+The `FundsDeposited` event emits the unique `V3RelayData` for an individual deposit. No additional fields are defined. 
+The `V3FundsDeposited` event emits the unique `V3RelayDataLegacy` for an individual deposit. No additional fields are defined.
```
bmzig, line 150: Should you highlight that this will stop being emitted once all spoke pools are upgraded?

UMIPs/umip-179.md
pxrl: True...there's no guarantee that the fill `relayer` or `msg.sender` on the destination chain are correct on the origin chain. The worst case would be a lite chain deposit being filled on a non-evm chain, where the repayment chain was set as the destination chain and the relayer address was set assuming a non-evm chain.

In this case I don't see what can be done other than to drop the repayment entirely. wdyt?

UMIPs/umip-179.md
```diff
@@ -126,28 +181,38 @@ The `RequestedSpeedUpV3Deposit` emits specifies the following fields:
...
 
-### FilledV3Relay
-The `FilledV3Relay` event extends the `V3RelayData` type by adding the following fields:
+### FilledRelay, FilledV3Relay
+
+The `FilledRelay` event extends the `V3RelayData` type by applying the following adjustments:
+| Name | Type | Description |
+| :--- | :--- | :---------- |
+| message | omitted | This field is omitted from the `FilledRelay` event in favour of the `messageHash` field. |
+| messageHash | bytes32 | The keccak256 hash of the `V3RelayData` message field. This field is included in place of the `V3RelayData` message field. See also [Computing RelayData hashes](#computing-relaydata-hashes). |
```
bmzig, line 202: Is the "See also" necessary here? Since we don't use the message hash anymore for RelayData hashes.

pxrl: True - it should instead be a description of how to hash the message field.

UMIPs/umip-179.md
pxrl: The number of edge cases for this to go wrong are increasing with the ability for relayers to specify their own repayment addresses. I kinda wonder if we should ultimately provide something like a contract where relayers can specify a default repayment addresses for each chain; this would be used in case of the proposer being unable to resolve a valid address.

UMIPs/umip-179.md
pxrl: So does this mean that whenever we have a fill with depositId > 2³²-1, the proposer/disputer would just skip it entirely, and would instead wait for the deposit to arrive?

My expectation is that most deposit/fill pairs would still continue to occur within the same bundle, even when the deposit was made via `unsafeDeposit()`, so in general I think we should still be able to match most deposits and fills out of the log buffer that the proposer/disputer populates on startup.

UMIPs/umip-179.md
pxrl: > No, in this case the relayer could only fill once because the relay hash is the same for both deposits. The deposit pays twice and only gets filled once. The relayer gets refunded for one their one fill

If the proposer/disputer resolves solely on RelayData hash then I'd still expect it to match a new deposit (via `unsafeDeposit()`) with a very old fill that was made for whatever other deposit collided. So in that case, it'd conclude that the old fill was valid and would then enqueue a relayer refund, and would not refund the depositor.

So the argument is not that the fill would occur twice, but that the proposer/disputer would issue two relayer repayments for the same fill. Does that make sense?

UMIPs/umip-179.md
pxrl: @mrice32 @nicholaspai This is something that we can enforce at the contract level, and not enforcing it there leads to this risk that the depositor's funds get locked in the protocol. It costs a bit more gas to do this but gives a lot more safety to the depositor. Given that we already need to modify the SpokePool for the RelayData hash, would we consider also enforcing that `depositor` is a valid EVM address?

UMIPs/umip-179.md
pxrl: I'm in two minds about it. On the one hand, it's a contract-level implementation detail that doesn't really concern the protocol. At the same time, if we bake in some transition concept then we could arguably start stripping the legacy support out of the bots once we complete the migration.

We would still need to circle back to this UMIP and make a cleanup though, so I think the question of whether to bake this in or not is subject to how quickly we expect to remove the legacy support from the UMIP.

UMIPs/umip-179.md
pxrl: https://github.com/UMAprotocol/UMIPs/pull/611/commits/e6e4980e69bfc28bca69e90d7a1a1482c382d4d4

UMIPs/umip-179.md
pxrl: @nicholaspai @mrice32 I've added a description of how to handle deposit refunds and relayer repayments when the receiving address is invalid for the target chain: https://github.com/UMAprotocol/UMIPs/commit/cab40f82e28a877634037c473fc9015c60f782bd

UMIPs/umip-179.md
```diff
@@ -39,25 +39,50 @@ All global constants from UMIP-157 will be retained for use in Across v3. The gl
...
 | outputAmount | uint256 | The amount of outputToken that is received by the recipient. |
 | originChainId | uint256 | The chain ID of the origin SpokePool. |
-| depositId | uint32 | The ID uniquely identifying a deposit on the origin chain. |
+| depositId | uint256 | The ID identifying a deposit on the origin chain. |
 | fillDeadline | uint32 | The Unix timestamp on the destination chain after which the deposit can no longer be filled. |
 | exclusivityDeadline | uint32 | The optional Unix timestamp on the destination chain after which any relayer can fill the deposit. |
 | message | bytes | Optional data that is forwarded to the recipient as part of a relay. |
 
+#### Note
+- `V3RelayData` specifies `bytes32` representation for addresses (`depositor`, `recipient`, ...) in order to interface with non-EVM chains. EVM addressed supplied shall be promoted to type `bytes32` with the upper bytes zeroed.
```
mrice32, line 62:
```suggestion
- `V3RelayData` specifies `bytes32` representation for addresses (`depositor`, `recipient`, ...) in order to interface with non-EVM chains. EVM addressed supplied shall be promoted to type `bytes32` with the upper 12 bytes zeroed. 
```

UMIPs/umip-179.md
mrice32: I think we say that they _are_ deprecated. The simplest way to avoid a separate vote once the upgrade happens is to set a specific block timestamp in this UMIP after which it is assumed that they will no longer be emitted, so they can be ignored by implementations.

Thoughts?

UMIPs/umip-179.md
mrice32: > One additional and very simple check that could protect against this - the proposer and disputer could be required to verify both the relayDataHash and a subset of the fields

If the relayHash matches, these would have to match, right? Otherwise, we've found a keccak hash collision, no? One important thing to also note is that it's pretty hard to spoof collisions without controlling the calling user's wallet. The depositor _and_ msg.sender (in addition to nonce) must match to create a collision.

> The proposer then identifies that fill and credits the relayer with the repayment, meaning the relayer gets paid twice and the depositor can never receive their refund. Is this the same scenario you're thinking of?

Do we want this behavior to be universal? Or if the deposits are within some range of one another, we do a refund (since the duplicate is detectable in that case)?

If I'm understanding correctly, the primary motivation for the double relayer payment is that we don't have a good way to go back and find a deposit from long ago, so we can't distinguish between a prefill and double deposit. But I also think the most likely failure mode is a double transaction submission, where the deposits are very close together, so if we could protect users in that case, I think this sharp edge wouldn't be too painful for users.

Thoughts?

UMIPs/umip-179.md
mrice32: I think we've confirmed we will keep the old hash.

UMIPs/umip-179.md
```diff
@@ -218,67 +305,112 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+- No specific method is prescribed for resolving the fill on the destination chain. An `eth_getLogs` request can facilitate this, and if required, the [Bundle Block Range](#identifying-bundle-block-ranges) could be narrowed by a binary search over the `FillStatus` field. This is left as an implementation decision.
+
+#### Finding Expired Deposits
+For the purpose of computing depositor refunds, each `Deposit` shall be considered expired by verifying that:
+1. The `fillDeadline` timestamp elapsed within the [Bundle Block Range](#identifying-bundle-block-ranges) on the destination SpokePool (i.e. the `fillDeadline` expired between the `block.timestamp` of the destination chain's bundle start and end block),
 2. The `FillStatus` on the destination SpokePool is set to `Unfilled` or `SlowFillRequested`.
 
 Note:
 - Expired deposits shall be refunded to the `depositor` address on the origin SpokePool.
+  - If the `depositor` address is not valid on the `originChainId`, the deposit refund shall be discarded.
```
mrice32, line 351: Just marking this as something we might be able to remove if the check PR goes in.

UMIPs/umip-179.md
```diff
@@ -218,67 +305,112 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+#### Validating Fills
+Each of the `Fills` emitted within the [Bundle Block Range](#identifying-bundle-block-ranges) on a destination SpokePool shall be considered valid by verifying that:
+1. The `Fill` event `FillType` field is not set to `SlowFill`, AND
+2. The component `RelayData` maps exactly to a corresponding `Deposit` event emitted on the relevant `originChainId`. This may be determined by comparing the hashes of the two objects, AND
+3. The corresponding `Deposit` event occurred within or before the [Bundle Block Range](#identifying-bundle-block-ranges) on the origin chain SpokePool.
+
+If the `Deposit` event specifies `outputToken` 0x0 (i.e. the Zero Address), the equivalent SpokePool token on the destination chain shall be substituted in. For the purpose of determining `RelayData` equivalency, the updated/substituted `outputToken` shall be used in place of 0x0.
+
+#### Validating Pre-fills
+For each of the `Deposits` emitted within the [Bundle Block Range](#identifying-bundle-block-ranges) where no corresponding `Fill` is identified on the destination chain, identify the valid `Fill` according to the following criteria:
```
nicholaspai, line 337:
```suggestion
For each of the `Deposits` emitted within the [Bundle Block Range](#identifying-bundle-block-ranges) where no corresponding `Fill` is identified on the destination chain within the Bundle Block Range, identify the valid `Fill` according to the following criteria:
```

UMIPs/umip-179.md
```diff
@@ -218,67 +305,112 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+1. Verify that the destination chain `FillStatus` for the `Deposit` `RelayData` is `Filled` as at the destination chain end block number for the proposal.
+2. Resolve the corresponding `Fill` on the destination chain.
```
nicholaspai, line 339:
```suggestion
2. Resolve the corresponding `Fill` on the destination chain.
3. The `Fill` event occurred before the [Bundle Block Range](#identifying-bundle-block-ranges) on the destination chain SpokePool.
```

UMIPs/umip-179.md
```diff
@@ -218,67 +305,112 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
 
 ### Finding Slow Fill Requests
-For the purpose of computing slow fills to be issued to recipients, each `RequestedV3SlowFill` event emitted within the target block range on a destination SpokePool shall be considered valid by verifying that:
+For the purpose of computing slow fills to be issued to recipients, each `Slow Fill Request` emitted within the [Bundle Block Range](#identifying-bundle-block-ranges) on a destination SpokePool shall be considered valid by verifying that:
 1. The `inputToken` and `outputToken` addresses are equivalent at the deposit `quoteTimestamp`,
 2. The `fillDeadline` has not already elapsed relative to the `destinationChainId` bundle end block number,
-3. The destination SpokePool `FillStatus` mapping for the relevant `V3RelayData` hash is `SlowFillRequested`,
-4. The `RequestedV3SlowFill` `V3RelayData` is matched by a corresponding `V3FundsDeposited` event on the origin SpokePool,
+3. The destination SpokePool `FillStatus` mapping for the relevant `RelayData` hash is `SlowFillRequested`,
+4. The `Slow Fill Request` `RelayData` is matched by a corresponding `Deposit` event on the origin SpokePool,
```
nicholaspai, line 360:
```suggestion
4. The `Slow Fill Request` `RelayData` is matched by a corresponding `Deposit` event on the origin SpokePool that occurred within or before the the Bundle Block Range
```

UMIPs/umip-179.md
nicholaspai: > My expectation is that most deposit/fill pairs would still continue to occur within the same bundle, even when the deposit was made via unsafeDeposit()

Agreed with you here. I think my suggestion here is not needed.

UMIPs/umip-179.md
nicholaspai: Currently the proposer will not reward a refund for any duplicate expired deposits -- only the first would get refunded: https://github.com/across-protocol/sdk/pull/838

UMIPs/umip-179.md
pxrl: @nicholaspai Good catch!

UMIPs/umip-179.md
pxrl: Updated: https://github.com/UMAprotocol/UMIPs/commit/9758785973fb737e6546d5adb6bac956fdbb551b

UMIPs/umip-179.md
pxrl: I added some migration information here - wdyt? https://github.com/UMAprotocol/UMIPs/commit/f74a4ab60c8d6099cf09861fe8fa6981ea055cd7

nb. I constrained removal to earliest 7 days after migration just to allow the dust to settle.

UMIPs/umip-179.md
pxrl: > If the relayHash matches, these would have to match, right? Otherwise, we've found a keccak hash collision, no? One important thing to also note is that it's pretty hard to spoof collisions without controlling the calling user's wallet. The depositor _and_ msg.sender (in addition to nonce) must match to create a collision.

Agreed! The risk of a depositor reusing their salt (`depositNonce`) is much more realistic. It's a lot easier to create collisions now that deposit permits `fillDeadline` to be in the past.

> If I'm understanding correctly, the primary motivation for the double relayer payment is that we don't have a good way to go back and find a deposit from long ago, so we can't distinguish between a prefill and double deposit. But I also think the most likely failure mode is a double transaction submission, where the deposits are very close together, so if we could protect users in that case, I think this sharp edge wouldn't be too painful for users.
> 
> Thoughts?

I still feel allergic to the possibility of a double repayment, even if it can be confined to rare cases. It's very counter-intuitive and is likely to be quite tricky to investigate, if it ever happens. I'm much more in favour of finding solutions where we can eliminate the possibility. It's unclear whether we can achieve that given other constraints though.

UMIPs/umip-179.md
pxrl: Removed here: https://github.com/UMAprotocol/UMIPs/pull/611/commits/f74a4ab60c8d6099cf09861fe8fa6981ea055cd7

UMIPs/umip-179.md
```diff
@@ -125,29 +179,38 @@ The `RequestedSpeedUpV3Deposit` emits specifies the following fields:
...
 
-### FilledV3Relay
-The `FilledV3Relay` event extends the `V3RelayData` type by adding the following fields:
+### FilledRelay, FilledV3Relay
+
+The `FilledRelay` event extends the `V3RelayData` type by applying the following adjustments:
+| Name | Type | Description |
+| :--- | :--- | :---------- |
+| message | omitted | This field is omitted from the `FilledRelay` event in favour of the `messageHash` field. |
+| messageHash | bytes32 | The keccak256 hash of the `V3RelayData` message field. This field is included in place of the `V3RelayData` message field. |
```
Reinis-FRP, line 200: this is `bytes32(0)` for empty message bytes.

UMIPs/umip-179.md
```diff
@@ -218,67 +305,112 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
-For the purpose of computing slow fills to be issued to recipients, each `RequestedV3SlowFill` event emitted within the target block range on a destination SpokePool shall be considered valid by verifying that:
+For the purpose of computing slow fills to be issued to recipients, each `Slow Fill Request` emitted within the [Bundle Block Range](#identifying-bundle-block-ranges) on a destination SpokePool shall be considered valid by verifying that:
 1. The `inputToken` and `outputToken` addresses are equivalent at the deposit `quoteTimestamp`,
 2. The `fillDeadline` has not already elapsed relative to the `destinationChainId` bundle end block number,
-3. The destination SpokePool `FillStatus` mapping for the relevant `V3RelayData` hash is `SlowFillRequested`,
-4. The `RequestedV3SlowFill` `V3RelayData` is matched by a corresponding `V3FundsDeposited` event on the origin SpokePool,
+3. The destination SpokePool `FillStatus` mapping for the relevant `RelayData` hash is `SlowFillRequested`,
+4. The `Slow Fill Request` `RelayData` is matched by a corresponding `Deposit` event on the origin SpokePool,
 5. The `originChainId` and `destinationChainId` are not Lite chains.
```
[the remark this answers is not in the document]
nicholaspai: @pxrl we actually don't need to do any historical queries to address 4 because if the `FillStatus` is `SlowFillRequested` then we know a slow fill request was sent. All that's left to do is to ensure that the request wasn't in the current bundle, so we would only need to check our event cache

UMIPs/umip-179.md
```diff
@@ -218,67 +303,110 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+#### Note
+- Depositor refunds are issued via the Relayer Repayment workflow.
 
-### Finding Expired Deposits
-For the purpose of computing depositor refunds, each `V3FundsDeposited` event shall be considered expired by verifying that:
-1. The `fillDeadline` timestamp elapsed within the target block range on the destination SpokePool (i.e. the `fillDeadline` expired between the `block.timestamp` of the destination chain's bundle start and end block),
+#### Validating Fills
+Each of the `Fills` emitted within the [Bundle Block Range](#identifying-bundle-block-ranges) on a destination SpokePool shall be considered valid by verifying that:
+1. The `Fill` event `FillType` field is not set to `SlowFill`, AND
+2. The component `RelayData` maps exactly to a corresponding `Deposit` event emitted on the relevant `originChainId`. This may be determined by comparing the hashes of the two objects, AND
```
Reinis-FRP, line 326: note that fill and slow fill events don't have full message bytes anymore, so one first needs to hash the message (or use `bytes32(0)`) from the deposit event to match hashes. Or is it expected to inspect fill/slow fill calldata where original message bytes would be still present?

UMIPs/umip-179.md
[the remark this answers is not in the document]
pxrl: @nicholaspai I've added some wording here: https://github.com/UMAprotocol/UMIPs/commit/065dc86b7a0b649b7cccfd53c17d269bc968cafb

I made some tweaks to your initial suggestion. lmk if it sounds OK.

UMIPs/umip-179.md
pxrl: @Reinis-FRP I've updated the description to make it more obvious - is this OK or does it need more? https://github.com/UMAprotocol/UMIPs/commit/c4005eec6ed63f6ccce484d1c31ac28cb7f4c0b2

UMIPs/umip-179.md
pxrl: Updated: https://github.com/UMAprotocol/UMIPs/commit/c4005eec6ed63f6ccce484d1c31ac28cb7f4c0b2

nicholaspai (dismissed):
All good from me

Reinis-FRP (dismissed):

UMIPs/umip-179.md
```diff
@@ -125,29 +179,38 @@ The `RequestedSpeedUpV3Deposit` emits specifies the following fields:
...
-The `RequestedV3SlowFill` event emits `V3RelayData`. This event is emitted on the destination chain and is intended to signal to proposers that a slow fill has been requested.
+### RequestedSlowFill, RequestedV3SlowFill
+The `RequestedSlowFill` event emits an `V3RelayData` instance.
+The `RequestedV3SlowFill` event emits an `V3RelayDataLegacy` instance.
 
-Note:
-- `RequestedV3SlowFill` events cannot occur until the `exclusivityDeadline` timestamp has elapsed on the destination chain.
+#### Note
+- These events are emitted on the destination chain and signal to proposers that a slow fill has been requested for a specific deposit.
+- `RequestedV3SlowFill` events cannot be emitted until the `exclusivityDeadline` timestamp has elapsed on the destination chain.
```
mrice32, line 191: Why does this need to be in the UMIP. Isn't this enforced by the contracts?

I guess this is really a broader question about these notes: are we intending for them to be informational in nature or enforced by the UMIP (i.e. a RequestedSlowFill event emitted before `exclusivityDeadline` would be ignored or otherwise considered invalid)? If informational, we should make that clear. "The contracts are expected to enforce ...".

UMIPs/umip-179.md
mrice32: Looks like this isn't resolved AFAICT. The UMIP suggests that deposits are refunded if not filled, but I don't see a clarification about what "filled" really means here (i.e. a duplicate deposits are allowed and can all be considered filled by the same fill and thus result in loss of funds).

UMIPs/umip-179.md
```diff
@@ -218,67 +306,130 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+- Fills of type `SlowFill` are valid, but are not relevant for computing relayer repayments.
+
+### Validating Pre-fills
+For each of the `Deposits` emitted within the `Bundle Block Range` where no corresponding `Fill` is identified on the destination chain within the `Bundle Block Range`, identify the valid `Fill` according to the following criteria:
+1. Verify that the destination chain `FillStatus` for the `Deposit` `RelayData` is `Filled` as at the destination chain end block number for the proposal.
+2. Resolve the corresponding `Fill` on the destination chain.
+3. Verify that the `Fill` occurred prior to the current the `Bundle Block Range` on the destination chain SpokePool.
+
+#### Note
+- No specific method is prescribed for resolving the fill on the destination chain. An `eth_getLogs` request can facilitate this, and if required, the `Bundle Block Range` could be narrowed by a binary search over the `FillStatus` field. This is left as an implementation decision.
```
mrice32, line 351:
```suggestion
- No specific method is prescribed for resolving the fill event on the destination chain. An `eth_getLogs` request can facilitate this, and if required, the `Bundle Block Range` could be narrowed by a binary search over the `FillStatus` field. This is left as an implementation decision.
```

UMIPs/umip-179.md
```diff
@@ -218,67 +306,130 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
 
 ### Finding Expired Deposits
-For the purpose of computing depositor refunds, each `V3FundsDeposited` event shall be considered expired by verifying that:
-1. The `fillDeadline` timestamp elapsed within the target block range on the destination SpokePool (i.e. the `fillDeadline` expired between the `block.timestamp` of the destination chain's bundle start and end block),
+For the purpose of computing depositor refunds, each `Deposit` shall be considered expired by verifying that:
+1. The `fillDeadline` timestamp elapsed within the `Bundle Block Range` on the destination SpokePool (i.e. the `fillDeadline` expired between the `block.timestamp` of the destination chain's bundle start and end block),
 2. The `FillStatus` on the destination SpokePool is set to `Unfilled` or `SlowFillRequested`.
```
mrice32, line 356: Should we clarify what block the FillStatus should be queried at?

UMIPs/umip-179.md
mrice32: I think we can remove this for now. I suspect this check will be included.

UMIPs/umip-179.md
```diff
@@ -218,67 +306,130 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
-For the purpose of computing slow fills to be issued to recipients, each `RequestedV3SlowFill` event emitted within the target block range on a destination SpokePool shall be considered valid by verifying that:
-1. The `inputToken` and `outputToken` addresses are equivalent at the deposit `quoteTimestamp`,
-2. The `fillDeadline` has not already elapsed relative to the `destinationChainId` bundle end block number,
-3. The destination SpokePool `FillStatus` mapping for the relevant `V3RelayData` hash is `SlowFillRequested`,
-4. The `RequestedV3SlowFill` `V3RelayData` is matched by a corresponding `V3FundsDeposited` event on the origin SpokePool,
+For the purpose of computing slow fills to be issued to recipients, each `Slow Fill Request` emitted within the `Bundle Block Range` on a destination SpokePool shall be considered valid by verifying that:
+1. The `fillDeadline` has not already elapsed relative to the `destinationChainId` bundle end block number,
+2. The `inputToken` and `outputToken` addresses are equivalent at the deposit `quoteTimestamp`,
+3. The destination SpokePool `FillStatus` mapping for the relevant `RelayData` hash is `SlowFillRequested`,
```
mrice32, line 368: Same question, should we clarify what block this should be queried at?

UMIPs/umip-179.md
```diff
@@ -218,67 +306,130 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+#### Note
+- A `slow Fill Request` is made by supplying a complete copy of the relevant `RelayData` emitted by a `Deposit` event.
+- The resulting set of validated `Slow Fill Requests` shall be included as SlowFills in the subsequent root bundle proposal.
+- A `Slow Fill Request` may correspond to a `Deposit` from previous bundles.
+
+### Finding Early Slow Fill Requests
+When an early `Slow Fill Request` is implied, the `Slow Fill Request` shall be validated as follows:
+1. The `fillDeadline` has not already elapsed relative to the `destinationChainId` bundle end block number,
+2. The `inputToken` and `outputToken` addresses are equivalent at the deposit `quoteTimestamp`,
+3. The `originChainId` and `destinationChainId` are not Lite chains.
```
mrice32, line 381:
```suggestion
3. Neither the`originChainId` nor the `destinationChainId` is a Lite chain.
```

mrice32 (commented):
Based on my current read of this UMIP, it's unclear how we deal with duplicate deposits.

I think we have 2 options:
1. If a deposit occurs within the bundle block range and one or more deposits with an identical relay hash occurs < X seconds before the deposit in question, a refund is generated for the later deposit.
2. If filled, only the first deposit is filled. All other deposits generate a relayer repayment, but no additional fill. If unfilled, each deposit is refunded.

nicholaspai:
> Based on my current read of this UMIP, it's unclear how we deal with duplicate deposits.
> 
> I think we have 2 options:
> 
> 1. If a deposit occurs within the bundle block range and one or more deposits with an identical relay hash occurs < X seconds before the deposit in question, a refund is generated for the later deposit.
> 2. If filled, only the first deposit is filled. All other deposits generate a relayer repayment, but no additional fill. If unfilled, each deposit is refunded.

I think 2 is a rephrasing of how I understand duplicate deposits should be handled:

2b. All expired, unfilled deposits are refunded

this is what I recommend to keep the logic simple. It's also good UX for duplicate depositors. It does not modify the fill logic (all fills matched with deposits are refunded). And your latter point "All other deposits generate a relayer repayment, but no additional fill. If unfilled, each deposit is refunded." is how I believe we handle "pre-fills" now.

UMIPs/umip-179.md
```diff
@@ -218,67 +306,130 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
-- Deposits where the `originChainId` is a "Lite" chain in the AcrossConfigStore as of the `quoteTimestamp` shall always be repaid on the deposit's origin chain. This means the protocol overrides the relayer's requested `repaymentChainId` with the `originChainId` instead.
+
+The applied `repaymentChainId` shall be determined as follows:
+- When the `originChainId` is a `Lite chain` as at the `Deposit` `quoteTimestamp`: `originChainId`, ELSE
+- When no `PoolRebalanceRoute` exists for the `repaymentToken` on the `Fill` `repaymentChainId`: `destinationChainId`, ELSE
+- The `repaymentChainId` as specified in the `Fill`.
+
+The applied `repayment address` shall be determined as follows:
+- When the `Fill` `relayer` address is valid for the applied `repaymentChainId`: `relayer`, ELSE
+- The `Fill` `msg.sender` address.
```
bmzig, line 413: The `msg.sender` is only defined for the destination chain ID and not necessarily a valid address on the repayment chain, I believe the best we can reasonably do is verify that the `msg.sender` is a bytes20 address and the chain is EVM, but this still leaves room to refund to some dead address if the filler is not an EOA.

UMIPs/umip-179.md
pxrl:
```suggestion
2. The `FillStatus` on the destination SpokePool is set to `Unfilled` or `SlowFillRequested` as at the end of the `Bundle Block Range`.
```

UMIPs/umip-179.md
pxrl:
```suggestion
```

UMIPs/umip-179.md
pxrl:
```suggestion
3. The destination SpokePool `FillStatus` mapping for the relevant `RelayData` hash is `SlowFillRequested` as at the end of the `Bundle Block Range`,
```

UMIPs/umip-179.md
pxrl: Yeah; this is a hangover from the v2 -> v3 upgrade and was intended for informational purposes.  It seemed relevant at the time; it's probably fine to drop it now.

```suggestion
```

UMIPs/umip-179.md
pxrl: Agree. There's still opportunity for a relayer to lose funds if they specify a mismatched `repaymentChainId` and `relayer` repayment address. This is most obviously an issue on chains where it's not possible to deploy to the same contract address (i.e. zkSync family).

Examples of mis-specifying the repayment information include:
- Requesting non-origin repayment when the origin chain is a lite chain.
- Requesting repayment on a chain where `outputToken` has no corresponding mapping.
- Requesting repayment to a non-EVM address on an EVM chain, or an EVM address on a non-EVM chain.
- Any combination of the above.

One distinction that that sticks out to me is that specifying invalid repayment information does not actually constitute an invalid fill - especially since the repayment information is separate from the fill (RelayData) information. I'm in favour of considering the fill valid, but then skipping the relayer repayment in this case.

I think eventually it makes sense for relayers to maintain their own repayment lookup tables on chain, such that the proposer can query the repayment mappings correctly (i.e. the relayer specifies a repayment chain, and the proposer queries the correct repayment address from that chain).

nicholaspai:
> A `Deposit` is considered `Filled` when the `Deposit` `RelayData` hash has state `Filled` or `ReplacedSlowFill` at the destination SpokePool.

I think you're confusing the FillStatus type from the FillType type:

```solidity
export enum FillStatus {
  Unfilled = 0,
  RequestedSlowFill,
  Filled,
}

export enum FillType {
  FastFill = 0,
  ReplacedSlowFill,
  SlowFill,
}
```

We can probably change this to be "A `Deposit` is considered `Filled` when the relayFillStatuses() for the relay data hash returns `Filled` on the destination SpokePool"

nicholaspai:
> > > Based on my current read of this UMIP, it's unclear how we deal with duplicate deposits.
> > > I think we have 2 options:
> > > 
> > > 1. If a deposit occurs within the bundle block range and one or more deposits with an identical relay hash occurs < X seconds before the deposit in question, a refund is generated for the later deposit.
> > > 2. If filled, only the first deposit is filled. All other deposits generate a relayer repayment, but no additional fill. If unfilled, each deposit is refunded.
> > 
> > 
> > I think 2 is a rephrasing of how I understand duplicate deposits should be handled:
> > 2b. All expired, unfilled deposits are refunded
> > this is what I recommend to keep the logic simple. It's also good UX for duplicate depositors. It does not modify the fill logic (all fills matched with deposits are refunded). And your latter point "All other deposits generate a relayer repayment, but no additional fill. If unfilled, each deposit is refunded." is how I believe we handle "pre-fills" now.
> 
> I'm also leaning towards (2) in absence of any more comprehensive solution to it (...I haven't been able to think of one).
> 
> I'm not sure it's the best UX for depositors though - it's painful in the case they make a duplicate deposit and the repayment gets credited to the relayer of the original fill, but I think this is caveat emptor when using a function called `unsafeDeposit()`. The obligation lies with the depositor to avoid submitting a duplicate deposit, so we just need to provide clear guidance about the need to pre-emptively verify the `FillStatus` on the destination chain when selecting their `depositNonce`. Maybe in future we can find a better way to handle this.
> 
> I think (2) also allows us to settle on the simplest definition of what a filled relay is - when a deposit's RelayData hash has state `Filled` or `ReplacedSlowFill` on the destination chain, then it is considered `Filled`. This is deliberately ignorant of RelayData duplicates, collisions and even timing.
> 
> [edit]: There is one sticking point with the above proposed definition - it works poorly when the deposit is generated once, but is incidentally submitted more than once. This can be done by the depositor (standard deposit-first scenario) and can happen within the `quoteTimestampBuffer`, which (per present `quoteTimestampBuffer` configuration of 3600 seconds) guarantees that the duplicate events should be emitted within the proposer's lookback.
> 
> Scenarios to consider are:
> 
> * Duplicate deposits are submitted within the same bundle block range.
> * Duplicate deposits are submitted in adjacent bundle block ranges.
> 
> The `quoteTimestamp` parameter _should_ protect against accidental re-submitted deposits being further apart than that, though we have no prescribed rules for how `quoteTimeBuffer` can be set relative to the proposal challenge window.
> 
> The best I can think of at the moment is this convoluted description (edited after Nick feedback):
> 
> _A `Deposit` is considered `Filled` when the destination SpokePool's `FillStatus` for the `Deposit` `RelayData` has state `Filled`._
> 
> _Note:_
> 
> * _In case multiple identical `Deposits` occur on a given origin SpokePool within the same or adjacent `Bundle Block Ranges`, only the initial `Deposit` shall be considered `Filled`. Subsequent `Deposits` within this range that are inherently unable to be filled on the destination chain shall be refunded._

The rule I've implemented so far in https://github.com/across-protocol/sdk/pull/835 is:
- When a fill is validated and due for a refund in a bundle, look for any duplicate deposits matching the fill (i.e. not including the original deposit) that are in the same or a prior bundle, and refund these duplicate deposits
    - These duplicates can no longer get refunded as expired deposits since the relay data hash is Filled on the destination SpokePool, and these deposits cannot trigger pre-fill refunds anymore
- When a deposit is sent that triggers a pre-fill refund, look for any duplicate deposits in the same bundle as this deposit and refund these duplicate deposits
    - These duplicate deposits would have otherwise triggered duplicate pre-fill refunds, resulting in a transfer of funds from duplicate depositors to pre-fillers, so we might as well repay the depositor in this case
    - If a deposit triggers a pre-fill refund and its the only deposit in the bundle, then the depositor's funds will pay the pre-filler

These rules do not pay duplicate depositors who are:
- unlucky to send duplicate deposits across two different bundles that match with a pre-fill and are the only deposit in that bundle. If two or more duplicate deposits land in the same

I think this scenario is very unlikely if you consider how deposits are designed to work with pre-fills:
- We expect pre-fill depositors to send deposits in batches, in close succession to each other, so the chance of sending these deposits in different bundles is low. The depositor can also remedy the situation by making sure they send batch duplicate deposits in at least batches of 2. This way all duplicate deposits get refeunded. Or...the depositor can just not send duplicate deposits...

mrice32:
Is all this special logic worthwhile if depositors can just get unlucky with bundle timing and lose their funds anyway?

My thinking is that the following logic would work well and require no additional lookback. It is similar to the casework above, but it combines the two cases and decouples it the duplicate refund logic from the fill status.

1. We (already) grab deposits from 6 hours before the start of the bundle block range to the end of the block range.
2. For each deposit in our bundle block range, we check if there are one or more deposits that are earlier (block number or event index if the block number is equivalent), have a matching relay hash, and have a block.timestamp within 6 hours of the deposit in question.
3. If so, enque a refund for this deposit. Remove it from consideration for any other actions (matching fills, refund after expiry, etc).
4. ... continue the bundling process as usual.

Thoughts? Do you think this works?

UMIPs/umip-179.md
```diff
@@ -218,67 +305,129 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+### Validating Fills
+Each of the `Fills` emitted within the `Bundle Block Range` on a destination SpokePool shall be considered valid by verifying that:
+1. The `Fill` event `FillType` field is not set to `SlowFill`, AND
+2. The component `RelayData` maps exactly to a corresponding `Deposit` event emitted on the relevant `originChainId`, AND
+3. The corresponding `Deposit` event occurred within or before the `Bundle Block Range` on the origin chain SpokePool.
+
+#### Note
+- If the `Deposit` event specifies `outputToken` 0x0 (i.e. the Zero Address), the equivalent SpokePool token on the destination chain shall be substituted in. For the purpose of determining `RelayData` equivalency, the updated/substituted `outputToken` shall be used in place of 0x0.
+- `RelayData` equality can be determined by comparing the keccak256 hashes of two `RelayData` objects.
+- Fills of type `SlowFill` are valid, but are not relevant for computing relayer repayments.
```
nicholaspai, line 341: For SlowFill-ed deposits where the fill is a pre-fill, we should ideally refund the depositor. Otherwise there is no way for the depositor to get funds back. We can essentially assume deposits sent after a slow fill are always duplicates, by definition for that slow fill to have been created

nicholaspai:
Hm I think this simple algorithm might have legs but there are some important details:

> Is all this special logic worthwhile if depositors can just get unlucky with bundle timing and lose their funds anyway? I think, in general, we should aim to make as little depend on bundle block ranges as possible, since they can be manipulated by the proposer and (even if non-malicious) are unpredictable for users/relayers.
> 
> My thinking is that the following logic would work well and require no additional lookback. It is similar to the casework above, but it combines the two cases and decouples the duplicate refund logic from the fill status.
> 
> 1. We (already) grab deposits from 6 hours before the start of the bundle block range to the end of the block range.
> 2. For each deposit in our bundle block range, we check if there are one or more deposits that are earlier (block number or event index if the block number is equivalent), have a matching relay hash, and have a block.timestamp within 6 hours of the deposit in question.

If there are three duplicate deposits in the bundle block range, won't this refund the same deposit multiple times?

> 3. If so, enque a refund for this deposit. Remove it from consideration for any other actions (matching fills, refund after expiry, etc).
> 4. ... continue the bundling process as usual.
> 
> Thoughts? Do you think this works?

nicholaspai:
So I think the duplicate deposit rules we settled on are:
- if duplicate deposits are sent in the same bundle as a fill, refund the duplicate deposits
- if duplicate deposits are sent after a fill such that they trigger a pre-fill refund, send a pre-fill refund for each such deposit

UMIPs/umip-179.md
pxrl: Just to round this out, I think we settled on the maximally simple definition of a fill - tldr, a deposit is `filled` when the deposit FillStatus on the destination chain is in state `Filled`.

pxrl:
> So I think the duplicate deposit rules we settled on are:
> 
> * if duplicate deposits are sent in the same bundle as a fill, refund the duplicate deposits
> * if duplicate deposits are sent after a fill such that they trigger a pre-fill refund, send a pre-fill refund for each such deposit
> * a single deposit can only result in a single payout: to the depositor or the filler, depending on the timing of the deposit and the fill
> 
> The last bullet point is unsatisfying but this ruileset is simple to implement, unlike one where we'd want to remove the non-determinism for the depositor

Updated version:

```
So we are dealing with duplicate deposits in two cases:
duplicates are sent in the same bundle or in a prior bundle to the fill
In this case, when we see the fill:
If the fill is not a SlowFill, send duplicate deposits to filler
If the fill is a slow fill, send duplicate deposits back to depositor
duplicates are sent in a bundle after the fill’s bundle
If the fill is not a SlowFill, send duplicate deposits to filler
If the fill is a slow fill, send duplicate deposits back to depositor
So in all cases, duplicate deposits only get refunded once the deposit gets filled (or has already been filled at the time the deposit is mined):
duplicate deposits go to depositor if the fill is a SlowFill, Or
the filler
```

UMIPs/umip-179.md
```diff
@@ -218,67 +313,141 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+#### Note
 - Expired deposits shall be refunded to the `depositor` address on the origin SpokePool.
 - Depositor refunds are to be issued as part of the relayer refund procedure.
-- The `fillDeadline` timestamp shall be resolved to a block number on the destination chain in order to determine inclusion within the target block range.
+- The `fillDeadline` timestamp shall be resolved to a block number on the destination chain in order to determine inclusion within the `Bundle Block Range`.
+
+### Finding Unfillable Deposits
+For the purpose of computing depositor refunds, each duplicate `Deposit` shall be considered unfillable by verifying that:
+1. The `Deposit` is identical with another `Deposit`.
+2. The destination chain `FillStatus` for the `Deposit` is `Filled`.
```
nicholaspai, line 374: Do you mean to say that the FillStatus is `Filled` at the end of the bundle block range that contains the `Deposit`?  Just wondering when you should determine the `FillStatus`

UMIPs/umip-179.md
```diff
@@ -218,67 +313,141 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+3. The destination chain `Fill` `FillType` was `SlowFill`.
+4. The destination chain `Fill` occurred within the current `BundleBlockRange`.
```
nicholaspai, line 376: Do all these conditions have to happen? If so we should probably explicitly state that above the list or adding an AND after each point

UMIPs/umip-179.md
```diff
@@ -218,67 +313,141 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+Each valid `Fill` is subject to an LP fee. The procedure for computing LP fees is as defined in [UMIP-136 Add IS_RELAY_VALID as a supported price identifier](https://github.com/UMAprotocol/UMIPs/blob/7b1a046098d3e2583abd0372c5e9c6003b46ad92/UMIPs/umip-136.md), with the following amendments:
 - The AcrossConfigStore contract shall be used to identify the correct rate model, instead of a `RateModelStore` contract.
 - The `HubPool` `liquidityUtilizationCurrent()` and `liquidityUtilizationPostRelay()` functions shall be used instead of the `BridgePool` variant.
 - The event `inputToken` shall be mapped from the SpokePool address to a HubPool `l1Token` address by following the matching procedure outlined above.
-- The LP fee is computed between the `originChainId` and `FilledV3Relay.repaymentChainId` where the `relayExecutionInfo.FillType != SlowFill` and `FilledV3Relay.destinationChainId` otherwise.
+- The LP fee is computed between the `originChainId` specified by the `Deposit` and `repaymentChainId` specified by the relayer, where the `relayExecutionInfo.FillType != SlowFill` and the Fill `destinationChainId` otherwise.
```
nicholaspai, line 409: We actually collect LP fees on each SlowFill's execution, not sure if that was a mistake from the previous version but we should note that. SlowFill's pay out `inputAmount - lpFee` to the recipient and when and if the the SlowFill gets executed, the LP fee gets accumulated to the `bundleLpFees` entry in the `PoolRebalanceRoot`

UMIPs/umip-179.md
```diff
@@ -218,67 +313,141 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+2. The `FillStatus` on the destination SpokePool is set to `Unfilled` or `SlowFillRequested` as at the end of the `Bundle Block Range`.
 
-Note:
+#### Note
 - Expired deposits shall be refunded to the `depositor` address on the origin SpokePool.
 - Depositor refunds are to be issued as part of the relayer refund procedure.
-- The `fillDeadline` timestamp shall be resolved to a block number on the destination chain in order to determine inclusion within the target block range.
+- The `fillDeadline` timestamp shall be resolved to a block number on the destination chain in order to determine inclusion within the `Bundle Block Range`.
+
+### Finding Unfillable Deposits
```
nicholaspai, line 371: Would a better name be Duplicate Slow Filled Deposits ?

UMIPs/umip-179.md
```diff
@@ -218,67 +313,141 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+- `Deposits` are considered duplicate when they are emitted after their initial identical instance.
 
 ### Finding Slow Fill Requests
-For the purpose of computing slow fills to be issued to recipients, each `RequestedV3SlowFill` event emitted within the target block range on a destination SpokePool shall be considered valid by verifying that:
-1. The `inputToken` and `outputToken` addresses are equivalent at the deposit `quoteTimestamp`,
-2. The `fillDeadline` has not already elapsed relative to the `destinationChainId` bundle end block number,
-3. The destination SpokePool `FillStatus` mapping for the relevant `V3RelayData` hash is `SlowFillRequested`,
-4. The `RequestedV3SlowFill` `V3RelayData` is matched by a corresponding `V3FundsDeposited` event on the origin SpokePool,
+For the purpose of computing slow fills to be issued to recipients, each `Slow Fill Request` emitted within the `Bundle Block Range` on a destination SpokePool shall be considered valid by verifying that:
+1. The `fillDeadline` has not already elapsed relative to the `destinationChainId` bundle end block number,
```
mrice32, line 384: Is the bundle inclusive of the end block? If so, this is probably a bit clearer.

```suggestion
1. The `fillDeadline` is greater than `destinationChainId` bundle end block's `block.timestamp`,
```

UMIPs/umip-179.md
```diff
@@ -218,67 +313,141 @@ In addition to the description [UMIP-157](https://github.com/UMAprotocol/UMIPs/b
...
+2. The `inputToken` and `outputToken` addresses are equivalent at the deposit `quoteTimestamp`,
+3. The destination SpokePool `FillStatus` mapping for the relevant `RelayData` hash is `SlowFillRequested` as at the end of the `Bundle Block Range`,
```
mrice32, line 386:
```suggestion
3. The destination SpokePool `FillStatus` mapping for the relevant `RelayData` hash is `SlowFillRequested` at the end of the `Bundle Block Range`,
```

mrice32 (commented):
I've done a pass. Looks mostly good, but I think we may missing the SlowFill duplicate handling.

UMIPs/umip-179.md
nicholaspai: The bundle is inclusive of the end block

nicholaspai (approved):

bmzig (approved):

PR #619, opened 2025-04-10
feat: Handle deposits that do not map to a PoolRebalanceRoute mapping

nicholaspai:


UMIPs/umip-179.md
```diff
@@ -428,12 +428,13 @@ For each validated matching `Deposit` event, the relayer repayment amount shall
...
 - When the `originChainId` is a `Lite chain` as at the `Deposit` `quoteTimestamp`: `originChainId`, ELSE
-- When no `PoolRebalanceRoute` exists for the `repaymentToken` on the `Fill` `repaymentChainId`: `destinationChainId`, ELSE
+- When no `PoolRebalanceRoute` exists for the `inputToken` on the `Fill` `originChainId`: `originChainId`, ELSE
+- When no `PoolRebalanceRoute` exists for the `repaymentToken` on the `Fill` `repaymentChainId`: `originChainId`, ELSE
 - The `repaymentChainId` as specified in the `Fill`.
 
 The applied `repayment address` shall be determined as follows:
 - When the `Fill` `relayer` address is valid for the applied `repaymentChainId`: `relayer`, ELSE
-- The `Fill` `msg.sender` address.
+- The `Fill` `msg.sender` address. In this case, change the applied `repaymentChainId` to the `destinationChainId` if a `PoolRebalanceRoute` exists for both the `outputToken` on the `destinationChainId` and the `inputToken` on the `originChainId`.
```
pxrl, line 437: I'm a bit uncertain on this change. `repaymentChainId` should already be set according to `PoolRebalanceRoute` considerations, and `repayment address` is intended to be downstream of this. It seems awkward to me to circle back and change the `repaymentChainId` here.

`msg.sender` also doesn't work for `destinationChainId` where origin & destination chain families are not compatible (i.e. evm/svm).

UMIPs/umip-179.md
```diff
@@ -534,15 +537,20 @@ The Pool Rebalance Merkle Tree shall be constructed such that it is verifyable u
...
 
 ### Constructing the Relayer Refund Root
-At least one Relayer Refund Leaf shall be produced for each unique combination of SpokePool and `l1Token` for any of the following conditions:
+At least one Relayer Refund Leaf shall be produced for each unique combination of SpokePool and `repaymentToken` for any of the following conditions:
 - Valid `Fills`, OR
 - Expired `Deposits`, OR
 - Unfillable `Deposits`, OR
 - A negative running balance net send amount.
 
+Where `repayment` token is determined as follows for
```
pxrl, line 546:
```suggestion
Where `repaymentToken` is determined as follows for
```

UMIPs/umip-179.md
```diff
@@ -534,15 +537,20 @@ The Pool Rebalance Merkle Tree shall be constructed such that it is verifyable u
...
+- `Fills`: the [equivalent](#resolving-spokepool-tokens-to-their-hubpool-equivalent) L2 token address for the `repaymentChainId`, as computed [here](#computing-relayer-repayments), of the `Fill.
+- `Deposits`: the `inputToken`
+- negative running balance net send amounts: the L2 token address for the corresponding `l1Token` considered in Pool Rebalance Root production
```
pxrl, line 549: Not super keen on using `L2` terminology here since we're also supporting alternative L1s.

```suggestion
- `Fills`: the [equivalent](#resolving-spokepool-tokens-to-their-hubpool-equivalent) token address for the `repaymentChainId`, as computed [here](#computing-relayer-repayments), of the `Fill.
- `Deposits`: the `inputToken`
- negative running balance net send amounts: the token address for the corresponding `l1Token` considered in Pool Rebalance Root production
```

UMIPs/umip-179.md
```diff
@@ -534,15 +537,20 @@ The Pool Rebalance Merkle Tree shall be constructed such that it is verifyable u
...
+
 Each Relayer Refund Leaf shall be constructed as follows:
 - `amountToReturn` shall be set to `max(-netSendAmount, 0)`.
-- `l2TokenAddress` shall be set to the L2 token address for the corresponding `l1Token` considered in Pool Rebalance Root production.
+- `l2TokenAddress` shall be set to the `repayment` token as computed previously
```
pxrl, line 553:
```suggestion
- `l2TokenAddress` shall be set to the `repaymentToken` as computed previously
```

pxrl (commented):
This looks good. I added a couple of minor comments.

UMIPs/umip-179.md
```diff
@@ -428,12 +428,13 @@ For each validated matching `Deposit` event, the relayer repayment amount shall
 
 The applied `repaymentChainId` shall be determined as follows:
```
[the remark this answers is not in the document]
nicholaspai: I like deposit quote timestamp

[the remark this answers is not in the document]
nicholaspai: 55826194bf0c6e68776cba7c64b3183d751bbc5a

UMIPs/umip-179.md
nicholaspai: The situation I'm trying to capture with this change is:
- repaymentChainId is a valid chain given the Deposit.inputToken and Deposit.originChain
- However, `repaymentAddress` is an invalid EVM address (e.g. its a 32 byte SVM address)
- (I believe) we should then attempt to change `repaymentAddress` to the `msg.sender` and therefore also change the `repaymentChain` to the `destinationChain`

> msg.sender also doesn't work for destinationChainId where origin & destination chain families are not compatible (i.e. evm/svm).

Agreed, and so this is why I added the line: "if a `PoolRebalanceRoute` exists for both the `outputToken` on the `destinationChainId` and the `inputToken` on the `originChainId`"

WDYT?

UMIPs/umip-179.md
```diff
@@ -428,12 +428,13 @@ For each validated matching `Deposit` event, the relayer repayment amount shall
...
 - When the `originChainId` is a `Lite chain` as at the `Deposit` `quoteTimestamp`: `originChainId`, ELSE
-- When no `PoolRebalanceRoute` exists for the `repaymentToken` on the `Fill` `repaymentChainId`: `destinationChainId`, ELSE
+- When no `PoolRebalanceRoute` exists for the `inputToken` on the `Fill` `originChainId` as at the `Deposit` `quoteTimestamp`: `originChainId`, ELSE
+- When no `PoolRebalanceRoute` exists for the `repaymentToken` on the `Fill` `repaymentChainId` as at the `Deposit` `quoteTimestamp`: `originChainId`, ELSE
 - The `repaymentChainId` as specified in the `Fill`.
 
 The applied `repayment address` shall be determined as follows:
 - When the `Fill` `relayer` address is valid for the applied `repaymentChainId`: `relayer`, ELSE
-- The `Fill` `msg.sender` address.
+- The `Fill` `msg.sender` address. In this case, change the applied `repaymentChainId` to the `destinationChainId` if a `PoolRebalanceRoute` exists for both the `outputToken` on the `destinationChainId` and the `inputToken` on the `originChainId` as at the `Deposit` `quoteTimestamp`.
```
bmzig, line 437: This means that we can reach a path where we refund `msg.sender` on `originChainId`, which may not exist if `msg.sender` is a contract on destination. Is this fine? Should we add a third `ELSE` which marks the fill as unrepayable?

bmzig: Never mind. I think it's probably just better to refund here.

bmzig (approved):

UMIPs/umip-179.md
```diff
@@ -345,7 +345,7 @@ Each of the `Fills` emitted within the `Bundle Block Range` on a destination Spo
 3. The corresponding `Deposit` event occurred within or before the `Bundle Block Range` on the origin chain SpokePool.
 
 #### Note
-- If the `Deposit` event specifies `outputToken` `bytes32(0)` (i.e. the Zero Address), the equivalent SpokePool token on the destination chain shall be substituted in. For the purpose of determining `RelayData` equivalency, the updated/substituted `outputToken` shall be used in place of the Zero Address.
+- If the `Deposit` event specifies `outputToken` `bytes32(0)` (i.e. the Zero Address), the equivalent SpokePool token on the destination chain shall be substituted in. For the purpose of determining `RelayData` equivalency, the updated/substituted `outputToken` shall be used in place of the Zero Address. If there is no equivalent SpokePool token on the destination chain at the `Deposit` `quoteTimestamp`, then this deposit cannot be filled successfully. This means that any fills or slow fill requests sent for this deposit cannot be validated and will not be refunded.
```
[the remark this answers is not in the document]
nicholaspai: I think the UMIP should block these fills from being repaid, and then the dataworker should add its own protection as we've done [here](https://github.com/across-protocol/sdk/pull/970#discussion_r2045511985)

UMIPs/umip-179.md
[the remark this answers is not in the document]
nicholaspai: we would use the mainnet bundle end block. This is a good point, we actually can't really use deposit quote timestamp. Otherwise there's a possibility that a deposit gets marked as "can be repaid" but the pool rebalance route was deleted by the end of the bundle and we can't actually pay them back

UMIPs/umip-179.md
```diff
@@ -428,12 +428,13 @@ For each validated matching `Deposit` event, the relayer repayment amount shall
 
 The applied `repaymentChainId` shall be determined as follows:
 - When the `originChainId` is a `Lite chain` as at the `Deposit` `quoteTimestamp`: `originChainId`, ELSE
-- When no `PoolRebalanceRoute` exists for the `repaymentToken` on the `Fill` `repaymentChainId`: `destinationChainId`, ELSE
+- When no `PoolRebalanceRoute` exists for the `inputToken` on the `Fill` `originChainId` as at the `Deposit` `quoteTimestamp`: `originChainId`, ELSE
+- When no `PoolRebalanceRoute` exists for the `repaymentToken` on the `Fill` `repaymentChainId` as at the `Deposit` `quoteTimestamp`: `originChainId`, ELSE
```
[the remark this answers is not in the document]
nicholaspai: good point, clarified it a bit here: ad0ca526af6ac48b90acce1318a6d03bd0575902

UMIPs/umip-179.md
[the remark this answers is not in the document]
mrice32: Agreed. Since the DW blocks this, I think I would tweak the UMIP language to `this deposit will not be repaid regardless of whether it is fillable on the destination.`

It's a nit, but since the DW invalidates it regardless, we should specify that here, too.

mrice32 (approved):
LGTM!

On [this](https://github.com/UMAprotocol/UMIPs/pull/619#discussion_r2054718897) point, I think specifying 0x0 on origin creates a huge footgun no matter how we do it.
- If we resolve these at bundle end block, a change between the deposit and the bundle end block could invalidate the relayer's fill at no fault of their own. Note: this would just mean changes should be done very carefully.
- If it's at quote timestamp, there's no invalidation, but the bundling system may attempt a repayment that's impossible.

This all goes away if we just deprecate the 0x0 behavior (especially since we're now removing the old deposit function, which used this).

Note: even after deprecation, there's still some repayment location changes that might be impacted by a change in PoolRebalanceRoutes mid bundle, but I feel like that's a much smaller negative consequence than an invalidated fill.

mrice32 (approved):

PR #625, opened 2025-05-29
feat: svm support for across

Reinis-FRP:
Adds support for SVM chains, also fixing some interface inconsistencies for EVM

### DIFF
--- a/UMIPs/umip-179.md
+++ b/UMIPs/umip-179.md
@@ -623,14 +623,11 @@ The applied `repayment address` shall be determined as follows:
 - When the `Fill` `relayer` address is valid for the applied `repaymentChainId`: `relayer`, ELSE
 - The `Fill` `msg.sender` address. In this case, change the applied `repaymentChainId` to the `destinationChainId` if a `PoolRebalanceRoute` exists for both the `outputToken` on the `destinationChainId` and the `inputToken` on the `originChainId` as at the bundle's hub chain end block
 
-If the applied `repayment address` is not valid for the applied `repaymentChainId`, the repayment shall be discarded.
+If the applied `repayment address` is not valid for the applied `repaymentChainId`, the repayment shall be discarded from the relayer refund root but still considered when [computing running balances & net send amounts](#computing-running-balances--net-send-amounts). Examples of an invalid `relayer` address include:
+- An SVM address on an EVM chain.
+- An EVM address on an SVM chain. Even though in theory one might hold a corresponding private key, Across would not support refunding a relayer on a SVM chain to an address that has zeroes in its first 12 bytes.
 
 The rules above enforce that a `repaymentToken` for a successful relayer repayment is always equal to a token that is [equivalent](#resolving-spokepool-tokens-to-their-hubpool-equivalent) to the `inputToken` at the time of relevant bundle's hub chain end block.
-
-#### Note
-- Examples of an invalid `relayer` address include:
-    - An SVM address on an EVM chain.
-    - An EVM address on an SVM chain. Even though in theory one might hold a corresponding private key, Across would not support refunding a relayer on a SVM chain to an address that has zeroes in its first 12 bytes.
 
 #### SVM support
 
--- a/UMIPs/umip-179.md
+++ b/UMIPs/umip-179.md
@@ -294,7 +294,7 @@ The `FilledRelay` event extends the `V3RelayData` type by applying the following
 | :--- | :--- | :---------- |
 | message | omitted | This field is omitted from the `FilledRelay` event in favour of the `messageHash` field. |
 | messageHash | bytes32 | The keccak256 hash of the `V3RelayData` message field where the message is non-empty, or `bytes32(0)` for an empty message. This field is included in place of the `V3RelayData` message field. |
-| relayer | bytes32 | The address completing relay on the destination SpokePool. |
+| relayer | bytes32 | The repayment address supplied by the relayer to be refunded on the applied repayment chain. |
 | repaymentChainId | uint256 | The repayment chain ID requested by the relayer completing the fill. |
 | relayExecutionInfo | V3RelayExecutionEventInfo | The effective `recipient`, `message` and `outputAmount`, as well as the `FillType` performed (FastFill, ReplacedSlowFill, SlowFill). |
 
@@ -316,7 +316,7 @@ The `FilledRelay` event extends the `RelayData` type by applying the following a
 | :--- | :--- | :---------- |
 | message | omitted | This field is omitted from the `FilledRelay` event in favour of the `message_hash` field. |
 | message_hash | [u8; 32] | The hash of the message sent to the recipient. This is computed as `solana_program::keccak::hash(message)` where `message` is the `message` bytes in the `RelayData`. If the `message` field is empty, this will be set to `[0u8; 32]`. |
-| relayer | Pubkey | The address completing relay on the destination SpokePool, same as in EVM. |
+| relayer | Pubkey | The repayment address supplied by the relayer to be refunded on the applied repayment chain, same as in EVM. |
 | repayment_chain_id | u64 | The repayment chain ID requested by the relayer completing the fill, same as in EVM. |
 | relay_execution_info | RelayExecutionEventInfo | See details in the SVM support for [RelayExecutionEventInfo](#v3relayexecutioneventinfo) section above. |
 
--- a/UMIPs/umip-179.md
+++ b/UMIPs/umip-179.md
@@ -828,4 +828,3 @@ The SVM SpokePool implementation is available as the `svm_spoke` program in the 
 
 # Security considerations
 Across v3 has been audited by OpenZeppelin.
-
--- a/UMIPs/umip-179.md
+++ b/UMIPs/umip-179.md
@@ -214,7 +214,7 @@ Across V3 defines the following events:
 - RequestedV3SlowFill
 - ClaimedRelayerRefund
 
-`svm_spoke` program uses Anchor's [`emit_cpi`](https://www.anchor-lang.com/docs/features/events#emit_cpi) macro to emit events that are comparable with EVM events. On SVM Across supports only a subset of events as explicitly documented.
+`svm_spoke` program uses Anchor's [`emit_cpi`](https://www.anchor-lang.com/docs/features/events#emit_cpi) macro to emit events that are comparable with EVM events. On SVM Across supports only a subset of events as explicitly documented in the relevant subsections below.
 
 ### Event Deprecation
 The following events are marked for deprecation. See [Migration](#migration) for more information.
@@ -542,6 +542,10 @@ For each of the `Deposits` emitted within the `Bundle Block Range` where no corr
 #### Note
 - No specific method is prescribed for resolving the `Fill` on the destination chain. An `eth_getLogs` request can facilitate this, and if required, the `Bundle Block Range` could be narrowed by a binary search over the `FillStatus` field. This is left as an implementation decision.
 
+#### SVM support
+
+In order to resolve the `Fill` on the destination SVM chain, one can inspect transactions where the `FillStatusAccount` PDA was involved (using `getSignaturesForAddress` RPC method) as described on the SVM supported [Data Types](#data-types) section above and looking for the emitted `FilledRelay` event.
+
 ### Finding Expired Deposits
 For the purpose of computing depositor refunds, each `Deposit` shall be considered expired by verifying that:
 1. The `fillDeadline` timestamp elapsed within the `Bundle Block Range` on the destination SpokePool (i.e. the `fillDeadline` expired between the `block.timestamp` of the destination chain's bundle start and end block),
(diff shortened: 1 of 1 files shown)


UMIPs/umip-179.md
```diff
@@ -617,3 +617,232 @@ The Across v3 implementation is available in the Across [contracts-v2](https://g
 
 # Security considerations
 Across v3 has been audited by OpenZeppelin.
+
+# SVM support
```
pxrl, line 621: One structural comment - I'm not sure how to integrate this cleanly into the existing structure without creating an `SVM` section under each major section...but I'm also not sure whether it's a good fit being appended to the bottom.

pxrl (commented):
@Reinis-FRP This looks really good. I've made a first pass and only have minor comments. I'd like some additional sets of eyes on it + another pass.

UMIPs/umip-179.md
```diff
@@ -617,3 +617,232 @@ The Across v3 implementation is available in the Across [contracts-v2](https://g
...
+- FilledRelay
+- RequestedSlowFill
+
+### FundsDeposited
+
+The `FundsDeposited` event extends the `RelayData` type by applying the following adjustments:
+
+| Name | Type | Description |
+| :--- | :--- | :---------- |
+| origin_chain_id | omitted | This field is omitted from the `FundsDeposited` event. |
```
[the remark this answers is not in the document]
Reinis-FRP: I had spotted some more inconsistencies on the EVM side, but thought to better group them in a separate PR

UMIPs/umip-179.md
```diff
@@ -617,3 +617,232 @@ The Across v3 implementation is available in the Across [contracts-v2](https://g
...
+
+#### Bundle Block Range
+
+On SVM chains the `Bundle Block Range` is the pair of start and end slots for a given proposal. See [Identifying Bundle Block Ranges](#identifying-bundle-block-ranges) for guidance on identifying the `Bundle Block Range`.
+
+#### Fill Status
+
+A `Deposit` is considered to be `Filled` on the destination SVM chain when its `FillStatus` is resolved to `Filled` using `FillStatusAccount` as described in the [SVM Data Types](#svm-data-types) section above.
+
+### Method
```
pxrl, line 801: For the `Method` section, I am wondering if this is better integrated directly into the existing `Method` section above. In particular I think this would make sense for `Computing Relayer Repayments`, because the dance we have to do for validating `repaymentChainId` and `repaymentAddress` is already a bit complex.

UMIPs/umip-179.md
```diff
@@ -617,3 +617,232 @@ The Across v3 implementation is available in the Across [contracts-v2](https://g
...
+
+As SVM does not have a concept of chain IDs it should be derived by using the 48 least significant bits of `keccak256` hashed value of the chain name. As an illustration, the chain ID for Solana mainnet is `34268394551451` that can be derived as:
+
+```bash
+cast to-dec $(cast shr $(cast shl $(cast keccak solana-mainnet) $((256-48))) $((256-48)))
+```
+
+## SVM Data Types
+
+Across supports only selected data types on SVM as listed in the sections below.
```
md0x, line 635: Nit:
```suggestion
Across supports the following data types on SVM:
```

UMIPs/umip-179.md
Reinis-FRP: agree, have restructured SVM bits across relevant UMIP section

UMIPs/umip-179.md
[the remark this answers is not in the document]
Reinis-FRP: as I refactored SVM bits across the UMIP I ended up also fixing any EVM inconsistencies I found

UMIPs/umip-179.md
```diff
@@ -61,6 +79,25 @@ The `V3RelayData` type underpins the transfer of funds in or out of a SpokePool
...
+
+| Name | Type |
+| :--- |:---- |
+| depositor | Pubkey |
+| recipient | Pubkey |
+| exclusive_relayer | Pubkey |
+| input_token | Pubkey |
+| output_token | Pubkey |
+| input_amount | u64 |
+| output_amount | u64 |
```
mrice32, line 94: Maybe we've already discussed this, but if the output amount and input amounts are stored in u64, how are we going to represent amounts with 18 decimals of precision. If I'm doing my math right, we wouldn't be able to represent anything higher than ~18.4 ETH in 64 bits, for example.

UMIPs/umip-179.md
```diff
@@ -101,31 +141,53 @@ A FillType instance is emitted with each `FilledV3Relay` event (see below).
 | ReplacedSlowFill | 1 | The relay was initially requested to be completed via slow fill, but was subsequently fast-filled by a relayer. |
 | SlowFill | 2 | The relay was completed via slow fill. |
 
+#### SVM support
+
+`FillType` enum is emitted with each `FilledRelay` event on SVM with the same interpretation as when emitted with `FilledV3Relay` events on EVM.
```
mrice32, line 146: Side note: the EVM event should be called `FilledRelay` as well. See https://github.com/across-protocol/contracts/blob/919bfd3f90d28a6e0013b21fc4238e8b238547f3/contracts/SpokePool.sol#L1653

Note: I think we may have the old name throughout the UMIP and some note that `FilledRelay` should be considered an equivalent event (since both are technically allowed in the current version of the UMIP).

Would be good cleanup (in a follow-up) to completely drop `FilledV3Relay` now that it's gone from the contracts.

UMIPs/umip-179.md
```diff
@@ -368,6 +547,10 @@ For the purpose of computing depositor refunds, each `Deposit` shall be consider
 - Depositor refunds are to be issued as part of the relayer refund procedure.
 - The `fillDeadline` timestamp shall be resolved to a block number on the destination chain in order to determine inclusion within the `Bundle Block Range`.
 
+#### SVM support
+
+When evaluating if the `fillDeadline` / `fill_deadline` timestamp elapsed within the `Bundle Block Range` on the SVM destination SpokePool, one can use `getBlock` RPC method for the destination chain's bundle start and end slot and use the `blockTime` field to compare against the `fillDeadline` / `fill_deadline` timestamp.
```
mrice32, line 552: What if there is no block at that slot?

mrice32 (commented):
Looks great! Only minor comments.

Maybe I missed it, but I didn't see any explanation of how we deal with unresolvable relayer repayments:
1. Fill on SVM for a deposit whose origin chain is EVM and has no LP routes to Solana. The passed repayment address is also an SVM address. Since the repayment cannot happen on Solana, but we only have SVM addresses, what does the protocol do?
2. Fill on EVM for a deposit whose origin chain is SVM and has no LP routes to the EVM chain. The passed repayment address is also EVM. Since the repayment cannot happen on the EVM chain, but we only have EVM addresses, what does the protocol do?

Reinis-FRP:
> Looks great! Only minor comments.
> 
> Maybe I missed it, but I didn't see any explanation of how we deal with unresolvable relayer repayments:
> 
>     1. Fill on SVM for a deposit whose origin chain is EVM and has no LP routes to Solana. The passed repayment address is also an SVM address. Since the repayment cannot happen on Solana, but we only have SVM addresses, what does the protocol do?
> 
>     2. Fill on EVM for a deposit whose origin chain is SVM and has no LP routes to the EVM chain. The passed repayment address is also EVM. Since the repayment cannot happen on the EVM chain, but we only have EVM addresses, what does the protocol do?

Currently under [Computing Relayer Repayments](https://github.com/UMAprotocol/UMIPs/blob/b440df8c295c3657c9690193d6893ce5dc8d9c49/UMIPs/umip-179.md#computing-relayer-repayments) we state:

> If the applied `repayment address` is not valid for the applied `repaymentChainId`, the repayment shall be discarded.

UMIPs/umip-179.md
Reinis-FRP: good point! I think the end slot must be chosen with a produced block, but if start slot happens to be empty we should use the timestamp from the last produced block before it. Added this in 585e74308e2b2c164aeb74384435929f1c9cf968

UMIPs/umip-179.md
Reinis-FRP: Indeed, this could be an issue, addressed in 770376e98e7bc934c94143304be012bd9bf48f55

UMIPs/umip-179.md
```diff
@@ -247,55 +353,108 @@ A `PoolRebalanceLeaf` shall consist of the following:
...
 | :--- | :--- | :---------- |
-| chainId | uint256 | The SpokePool `chainId` referenced by the `RelayerRebalanceLeaf`. |
-| leafId | uint8 | Index of the `RelayerRefundLeaf` within the ordered array of `RelayerRefundLeaves`. |
-| l2TokenAddress | address[] | The SpokePool token used by this `RelayerRefundLeaf`.
 | amountToReturn | uint256 | Amount of `l2TokenAddress` to return to the HubPool. |
-| refundAddresses | uint256[] | Ordered array of addresses to be refunded by this `RelayerRefundLeaf`. |
+| chainId | uint256 | The SpokePool `chainId` referenced by the `RelayerRebalanceLeaf`. |
 | refundAmounts | uint256[] | Ordered array of amounts of `l2TokenAddress`to be refunded to the corresponding `refundAddress`. |
+| leafId | uint32 | Index of the `RelayerRefundLeaf` within the ordered array of `RelayerRefundLeaves`. |
+| l2TokenAddress | bytes32 | The SpokePool token used by this `RelayerRefundLeaf`.
```
md0x, line 360:
```suggestion
| l2TokenAddress | bytes32 | The SpokePool token used by this `RelayerRefundLeaf`. |
```

md0x (dismissed):

UMIPs/umip-179.md
```diff
@@ -436,13 +631,15 @@ The applied `repayment address` shall be determined as follows:
 - When the `Fill` `relayer` address is valid for the applied `repaymentChainId`: `relayer`, ELSE
 - The `Fill` `msg.sender` address. In this case, change the applied `repaymentChainId` to the `destinationChainId` if a `PoolRebalanceRoute` exists for both the `outputToken` on the `destinationChainId` and the `inputToken` on the `originChainId` as at the bundle's hub chain end block
 
-If the applied `repayment address` is not valid for the applied `repaymentChainId`, the repayment shall be discarded.
+If the applied `repayment address` is not valid for the applied `repaymentChainId`, the repayment shall be discarded from the relayer refund root but still considered when [computing running balances & net send amounts](#computing-running-balances--net-send-amounts). Examples of an invalid `relayer` address include:
```
pxrl, line 634: What's the intended impact of considering the repayment in running balance & net send amount computation? Should the discarded repayment be allocated to the LP instead, or should the tokens remain in the SpokePool for some possible future reconciliation??

UMIPs/umip-179.md
Reinis-FRP: I do have a separate branch for removing `V3` from EVM as they are out of contracts now. Will post that as a separate PR once we settle on this SVM specific PR.

UMIPs/umip-179.md
```diff
@@ -203,19 +295,31 @@ The `FilledRelay` event extends the `V3RelayData` type by applying the following
...
 
+#### SVM support
+
+The `FilledRelay` event extends the `RelayData` type by applying the following adjustments:
+
+| Name | Type | Description |
+| :--- | :--- | :---------- |
+| message | omitted | This field is omitted from the `FilledRelay` event in favour of the `message_hash` field. |
+| message_hash | [u8; 32] | The hash of the message sent to the recipient. This is computed as `solana_program::keccak::hash(message)` where `message` is the `message` bytes in the `RelayData`. If the `message` field is empty, this will be set to `[0u8; 32]`. |
+| relayer | Pubkey | The address completing relay on the destination SpokePool, same as in EVM. |
```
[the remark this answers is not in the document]
Reinis-FRP: Agree, Now that relayer can specify the `repaymentAddress` that is passed to `relayer` field instead of `msg.sender` on EVM we need to fix the description both for EVM and SVM. Addressed in 6819fb72817ca839f63ba2a5061cddc0103d1e89.

UMIPs/umip-179.md
Reinis-FRP: Had discussed this with @mrice32 before that better keep the excess in the SpokePool for easier tracking and potentially refunding out of protocol via spoke admin function. Clarified this in 17736a93172e80f1bfdca44bc68e9835a86d708c

pxrl (approved):
@Reinis-FRP Thanks for all your work on this :pray:

UMIPs/umip-179.md
mrice32: What does it mean if blockTime is not available? Is that just that node doesn't have the info? Or is that a hard property of the block (i.e. guaranteed to be the same on other nodes by consensus)?

Reinis-FRP: From [getBlockTime](https://solana.com/docs/rpc/http/getblocktime) docs:
> Each validator reports their UTC time to the ledger on a regular interval by intermittently adding a timestamp to a Vote for a particular block. A requested block's time is calculated from the stake-weighted mean of the Vote timestamps in a set of recent blocks recorded on the ledger.

So I understand `blockTime` can be null for some period of time, but not clear if that's possible once the block is finalized.

